PR #7, opened 2018-03-09
Completed TPM2 wrappers and wrapper example code

dgarske:
* Added new TPM2 wrapper functions, structures and example code.
* Progress with supporting parameter encryption/decryption using both session auth and object auth.
* Enhanced the `wolfTPM2_GetRCString` to include the enum name and for non-debug builds shows only the enum name.
* Changed `TPM2_GetNonce` to use current ctx and not use arg.
* Cleanup with the response parsing to reduce code / simplify.
* Cleanup of bad arg return code and error message display.
* Cleanup of line lengths / formatting.

### DIFF
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -117,7 +117,8 @@ static TPM_RC TPM2_SendCommandAuth(TPM2_CTX* ctx, TPM2_Packet* packet,
             /* check if encrypting parameters */
             if (auth->sessionAttributes & TPMA_SESSION_decrypt) {
                 /* get new nonce if required */
-                if (ctx->authCmd->sessionHandle != TPM_RS_PW && auth->nonce.size > 0) {
+                if (ctx->authCmd->sessionHandle !=
+                                            TPM_RS_PW && auth->nonce.size > 0) {
                     rc = TPM2_GetNonce(auth->nonce.buffer, auth->nonce.size);
                     if (rc != 0)
                         return rc;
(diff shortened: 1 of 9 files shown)


wolftpm/tpm2_wrap.h
```diff
@@ -25,15 +25,81 @@
...
+} WOLFTPM2_HANDLE;
+
+typedef struct WOLFTPM2_SESSION {
+    WOLFTPM2_HANDLE handle;
+    TPM2B_NONCE     nonceTPM;
+} WOLFTPM2_SESSION;
+
+typedef struct WOLFTPM2_KEY {
+    WOLFTPM2_HANDLE   handle;
+    TPM2B_PRIVATE     private;
```
JacobBarthelmeh: private or public will cause problems if trying to build with g++

JacobBarthelmeh: please also do a check through the for other locations of public/private

src/tpm2.c
```diff
@@ -4686,7 +4442,7 @@ TPM_RC TPM2_NV_Certify(NV_Certify_In* in, NV_Certify_Out* out)
     TPM2_CTX* ctx = TPM2_GetActiveCtx();
 
     if (ctx == NULL || in == NULL || out == NULL || ctx->authCmd == NULL)
-        return TPM_RC_BAD_ARG;
+        return BAD_FUNC_ARG;
```
JacobBarthelmeh: Why the change from separate error codes for wolfTPM to dependent on wolfSSL error codes?

wolftpm/tpm2_wrap.h
dgarske: Good find! Making fix for that.

src/tpm2.c
dgarske: The TPM_RC codes are from the TCG specification and adding non-standard codes for those isn't ideal. Since we have this strong dependency on wolfSSL I thought we'd just use wolf error codes. In fact the wolfTPM2_GetRCString function calls into wc_GetErrorString for negative values.

PR #26, opened 2018-08-15
Fixes for building with different wolfCrypt options

dgarske:
Fix for byte swap build error.

### DIFF
--- a/wolftpm/tpm2_types.h
+++ b/wolftpm/tpm2_types.h
@@ -154,14 +154,19 @@ typedef int64_t  INT64;
 #ifndef LABEL_MAX_BUFFER
 #define LABEL_MAX_BUFFER 128
 #endif
+
 #ifndef MAX_RSA_KEY_BITS
 #define MAX_RSA_KEY_BITS 2048
 #endif
 #ifndef MAX_RSA_KEY_BYTES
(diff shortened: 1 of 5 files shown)


wolftpm/tpm2_types.h
```diff
@@ -154,14 +154,19 @@ typedef int64_t  INT64;
...
 #endif
+
 #ifndef MAX_RSA_KEY_BITS
 #define MAX_RSA_KEY_BITS 2048
 #endif
 #ifndef MAX_RSA_KEY_BYTES
 #define MAX_RSA_KEY_BYTES ((MAX_RSA_KEY_BITS/8)*2)
 #endif
+
+#ifndef MAX_ECC_KEY_BITS
```
JacobBarthelmeh: Why not have this be MAX_ECC_BITS? so that if set by wolfCrypt then it would also be set to same size here?

dgarske:
They are different values. The wolfCrypt could be configured to a different max than the hardware.

wolftpm/tpm2_types.h
dgarske: They are different values. The wolfCrypt could be configured to a different max than the hardware.

PR #29, opened 2018-08-16
Improvements to the Native tests for AES encrypt/decrypt

dgarske:
* Added native examples for AES key creation and Encrypt/Decrypt. On the SLB 9670 both variants of `TPM2_EncryptDecrypt` and `TPM2_EncryptDecrypt2` appear to be unsupported commands.
* Moved the TPM 2.0 NV indexes in native_tests.c to tpm2.h.
* Improved the cleanup handling and code comments.
* Started on support for HMAC example.


examples/native/native_test.c
```diff
@@ -1111,21 +1229,36 @@ int TPM2_Native_Test(void* userCtx)
...
-    cmdIn.flushCtx.flushHandle = storage.handle;
-    TPM2_FlushContext(&cmdIn.flushCtx);
-
+    if (endorse.handle != TPM_RH_NULL) {
+        cmdIn.flushCtx.flushHandle = endorse.handle;
+        TPM2_FlushContext(&cmdIn.flushCtx);
+    }
+    if (storage.handle != TPM_RH_NULL) {
+        cmdIn.flushCtx.flushHandle = storage.handle;
+        TPM2_FlushContext(&cmdIn.flushCtx);
```
JacobBarthelmeh: do we need a check here on the return value and on other calls to TPM2_FlushContext

JacobBarthelmeh:
In the case of using SLB9760 TPM can we get a print out of unsupported feature ... or not have the tests run yet with ./examples/native/native_test.

```
TPM2_Load New AES Key Handle 0x80000002
TPM2_EncryptDecrypt2 failed 0x143: TPM_RC_COMMAND_CODE
TPM2_EncryptDecrypt2 failed 0x143: TPM_RC_COMMAND_CODE
Encrypt/Decrypt test failed, result not as expected!
```

dgarske:
@JacobBarthelmeh : Now says `TPM2_EncryptDecrypt2: Is not a supported feature without enabling due to export controls`.

PR #32, opened 2018-09-04
Added PKCS7 7 sign/verify example demonstrating large data

dgarske:
Demo shows PKCS7 sign/verify of large data using smaller buffer "chunks" and new `_ex` functions.


JacobBarthelmeh:
This breaks compatibility with older versions of wolfSSL. We should conditionally compile or give a warning. And update the readme to inform users.

PR #33, opened 2018-09-06
Added support for ST33TP I2C TPM 2.0 chip

dgarske:
Build using `./configure --enable-st33 --enable-i2c`.

Added key gen to benchmark.
Updated README.md.


JacobBarthelmeh:
This should have a follow up with README on altering the config.txt for use.

PR #40, opened 2018-10-05
Added wrapper API's and tests for performing hash

dgarske:
New API's are: `wolfTPM2_HashStart`, `wolfTPM2_HashUpdate` and `wolfTPM2_HashFinish`.


src/tpm2_wrap.c
```diff
@@ -1376,6 +1376,130 @@ int wolfTPM2_Clear(WOLFTPM2_DEV* dev)
...
+        }
+        pos += hashSz;
+    }
+
+#ifdef DEBUG_WOLFTPM
+    printf("wolfTPM2_HashUpdate: Handle 0x%x, DataSz %d\n",
+        (word32)in.sequenceHandle, dataSz);
+#endif
+
+    return rc;
```
JacobBarthelmeh: rc could be returned as a garbage value

PR #42, opened 2018-10-16
Fix to detect improper packing for TPM2_HEADER

dgarske:
Made the pack macros overridable.
Fix for compiler warning about XSTRLEN type conversion warning in wrapper hash test.


dgarske:
@JacobBarthelmeh : This is ready now. Received confirmation that this now works with the TI compiler.

PR #45, opened 2018-11-06
Refactor to eliminate the packed TPM2_HEADER

dgarske:
Adds better portability for source code.

### DIFF
--- a/wolftpm/tpm2_packet.h
+++ b/wolftpm/tpm2_packet.h
@@ -26,8 +26,8 @@
 
 #define TPM2_HEADER_SIZE 10 /* expected TPM2 header size */
 
-/* Note: The TPM2_HEADER structure must be packed */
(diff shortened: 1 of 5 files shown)


wolftpm/tpm2_packet.h
```diff
@@ -26,16 +26,16 @@
 
 #define TPM2_HEADER_SIZE 10 /* expected TPM2 header size */
```
embhorn: Should this be set to `sizeof(TPM2_HEADER)` just in case it ever gets used?

dgarske: No, because it would not be guaranteed to be 10 due to not being packed. I thought the note below about it not being used was good enough, but it can be totally removed. I like it there for reference purposes.

embhorn: Makes sense.

PR #50, opened 2018-11-14
ECDH enhancements

dgarske:
* Added TLS client/server support for ECDH using TPM.
* Added ECC KeyGen and SharedSecret support for CryptoDev callbacks.
* Added new wrapper API's `wolfTPM2_ECDHGenKey` and `wolfTPM2_ECDHGenZ`.
* Added ECDH tests in native and wrapper.

### DIFF
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -4856,6 +4856,62 @@ int TPM2_GetCurveSize(TPM_ECC_CURVE curveID)
     return 0;
 }
 
+int TPM2_GetTpmCurve(int curve_id)
+{
+    int ret = -1;
+#if !defined(WOLFTPM2_NO_WOLFCRYPT) && defined(HAVE_ECC)
+    switch (curve_id) {
+        case ECC_SECP192R1:
+            ret = TPM_ECC_NIST_P192;
+            break;
+        case ECC_SECP224R1:
+            ret = TPM_ECC_NIST_P224;
+            break;
+        case ECC_SECP256R1:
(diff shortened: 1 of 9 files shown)


src/tpm2.c
```diff
@@ -4856,6 +4856,62 @@ int TPM2_GetCurveSize(TPM_ECC_CURVE curveID)
...
+            ret = ECC_CURVE_OID_E;
+    }
+#endif
+    (void)curve_id;
+    return ret;
+}
+
+int TPM2_GetWolfCurve(int curve_id)
+{
+    int ret = ECC_CURVE_OID_E;
```
embhorn: When configured with `--disable-wolfcrypt` : 
```
  CC       src/src_libwolftpm_la-tpm2_wrap.lo
src/tpm2.c: In function ‘TPM2_GetWolfCurve’:
src/tpm2.c:4889:15: error: ‘ECC_CURVE_OID_E’ undeclared (first use in this function)
     int ret = ECC_CURVE_OID_E;
               ^~~~~~~~~~~~~~~
src/tpm2.c:4889:15: note: each undeclared identifier is reported only once for each function it appears in
```

embhorn:
I'm also seeing a failure from tls_client (with default config): 
```
pi@raspberrypi:~/wolfTPM $ ./examples/tls/tls_client 
TPM2 TLS Client Example
Failure -248 (0xffffff08): Error with hardware crypto use
```

embhorn:
Thanks, David!

PR #72, opened 2019-07-25
Added support for keeping an HMAC key loaded

dgarske:
Added support for keeping an HMAC key loaded on finish to allow using the same key for additional HMAC operations. ZD 5442

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ examples/csr/csr
 examples/tls/tls_client
 examples/pkcs7/pkcs7
 pkcs7tpmsigned.p7s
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -36,6 +36,7 @@ examples/pkcs7/pkcs7
 pkcs7tpmsigned.p7s
```
embhorn: Please add 
`pkcs7tpmsignedex.p7s`

dgarske: That file is created when running `./examples/pkcs7/pkcs7` and should not be tracked in repo.

PR #74, opened 2019-07-29
Fix to restore existing TPM context after calling `wolfTPM2_Test`.

dgarske:
ZD 5489 confirms fix.


src/tpm2_wrap.c
```diff
@@ -100,8 +100,13 @@ static int wolfTPM2_Init_NoDev(TPM2_CTX* ctx, TPM2HalIoCb ioCb, void* userCtx,
 int wolfTPM2_Test(TPM2HalIoCb ioCb, void* userCtx, WOLFTPM2_CAPS* caps)
 {
     int rc;
+    TPM2_CTX* current_ctx;
     TPM2_CTX ctx;
 
+    /* Backup active TPM context */
+    current_ctx = TPM2_GetActiveCtx();
```
embhorn: Does it make sense to validate the return value here?

dgarske: No need to check. NULL is okay or any pointer is okay.

embhorn:
Great! Thanks, David.

PR #79, opened 2019-08-14
Added support for using authentication with NV

dgarske:
Adds new wrapper API's: `wolfTPM2_NVWriteAuth`, `wolfTPM2_NVReadAuth` and `wolfTPM2_NVDeleteAuth`.


dgarske:
@embhorn : Fixes force pushed. Missed a change from that file when coping things over.

PR #81, opened 2019-10-22
Added new wrappers for shutdown and handle cleanup.

dgarske:
This PR adds three new wolfTPM wrapper API's for helping with TPM shutdown and handle cleanup.

The API's are:

```c
WOLFTPM_API int wolfTPM2_Shutdown(WOLFTPM2_DEV* dev, int doStartup);
WOLFTPM_API int wolfTPM2_UnloadHandles(WOLFTPM2_DEV* dev, word32 handleStart,
    word32 handleCount);
WOLFTPM_API int wolfTPM2_UnloadHandles_AllTransient(WOLFTPM2_DEV* dev);
```

ZD 5775

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -2897,6 +2897,70 @@ int wolfTPM2_HmacFinish(WOLFTPM2_DEV* dev, WOLFTPM2_HMAC* hmac,
     return rc;
 }
 
+/* performs a reset sequence */
+int wolfTPM2_Shutdown(WOLFTPM2_DEV* dev, int doStartup)
+{
+    int rc;
+    Shutdown_In shutdownIn;
+    Startup_In startupIn;
+
+    if (dev == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+    /* shutdown */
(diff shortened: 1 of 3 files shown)


src/tpm2_wrap.c
```diff
@@ -2897,6 +2897,70 @@ int wolfTPM2_HmacFinish(WOLFTPM2_DEV* dev, WOLFTPM2_HMAC* hmac,
...
+    shutdownIn.shutdownType = TPM_SU_CLEAR;
+    rc = TPM2_Shutdown(&shutdownIn);
+    if (rc != TPM_RC_SUCCESS) {
+    #ifdef DEBUG_WOLFTPM
+        printf("TPM2_Shutdown failed 0x%x: %s\n", rc, TPM2_GetRCString(rc));
+    #endif
+    }
+
+    /* startup */
+    if (doStartup) {
```
embhorn: Should the startup continue even if the shutdown failed?

dgarske: Yes, I'd like it to try and startup even if there was a shutdown failure... Assuming `doStartup` is set.

PR #87, opened 2020-04-16
Fix for building wolfTPM with older wolfSSL versions

dgarske:
Fix for building wolfTPM with older wolfSSL versions not supporting `wc_HashFree`.

Fixes Error:

```
src/tpm2_wrap.c:738:9: error: implicit declaration of function 'wc_HashFree' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        wc_HashFree(&hash, hashType);
        ^
src/tpm2_wrap.c:738:9: error: this function declaration is not a prototype [-Werror,-Wstrict-prototypes]
src/tpm2_wrap.c:2555:9: error: implicit declaration of function 'wc_HashFree' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        wc_HashFree(&hash, hashType);
        ^
3 errors generated.
```

### DIFF
--- a/wolftpm/tpm2_types.h
+++ b/wolftpm/tpm2_types.h
@@ -82,6 +82,7 @@ typedef int64_t  INT64;
     #ifdef WOLF_CRYPTO_CB
         #include <wolfssl/wolfcrypt/cryptocb.h>
     #elif defined(WOLF_CRYPTO_DEV)
+        /* old name for crypto callback support */
         #include <wolfssl/wolfcrypt/cryptodev.h>
     #endif
     #ifndef WOLFCRYPT_ONLY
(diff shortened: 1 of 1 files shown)


dgarske:
Thanks @tomoveu for the report!

tomoveu:
@dgarske Would it be easier to add minimal wolfSSL version as check during build configuration?

If wolfSSL is under version 3.15.4, the configure script can trigger the option you mentioned to me "--disable-wolfcrypt" and print some info message. 

I only found this small issue, because I made the assumption wolfTPM can be build against the official Rapsbian. And I really jumped to conclusions there, because the README mentions WolfTPM is tested on RPI2. Package provided in Raspbian right now for wolfSSL is version 3.15.3 and this is a really close call, because wc_HashFree is added just after that verison.

wolftpm/tpm2_types.h
```diff
@@ -92,6 +93,13 @@ typedef int64_t  INT64;
 	#ifdef DEBUG_WOLFTPM
 		#include <stdio.h>
 	#endif
+
+    #include <wolfssl/version.h>
+    #if defined(LIBWOLFSSL_VERSION_HEX) && LIBWOLFSSL_VERSION_HEX < 0x03015004
+        /* The wc_HashFree was added in v3.15.4, so use stub to allow building */
+        #define wc_HashFree(h, t)
```
danielinux, line 100: in recent versions, the prototype for this function is
`int wc_HashFree(wc_HashAlg* hash, enum wc_HashType type)`

I'd suggest to define this as its hardcoded return value, because it should be legal to call it as in:
`ret = wc_HashFree(x,y);` when it not compiled in.

We should do:
`#define wc_HashFree(h, t)  (0)` to ignore calls when the symbol is not there

PR #89, opened 2020-04-17
For for ECC verify software fallback if curve not supported

dgarske:
* Fix for ECC verify in crypto callback to try software if the curve is not supported (`TPM_RC_CURVE`) by the TPM hardware.
* Fixes for building without wolfCrypt RSA (when `NO_RSA` is defined).

ZD 9855


embhorn:
Tests are all passing. Thanks, David!

PR #91, opened 2020-04-28
Add linux dev tpm interface

tomoveu:
This PR adds the ability to talk to a TPM using the Linux TIS kernel driver. This means, talking to any TPM that supports the TCG TIS protocol, regardless if on I2C, SPI or even LPC bus. The Linux kernel abstracts the physical interface and provides TIS interface using char device at /dev/tpmX

All native tests pass, but one: the PCR Extend command test.

The code is tested against:
* Infineon TPM2.0 Iridium 9670 on RPI4 (SW: latest Raspbian/ Kernel 4.12)
* Nuvoton TPM2.0 npct650 on x86 test machine (SW: Ubuntu 19.10 / Kernel 5.3)

One thing that I would want to also add are unit tests.

This is my first try to add functionality to wolfTPM, so all feedback is highly welcomed :)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -162,6 +162,19 @@ then
 fi
 
 
+# Linux kernel TPM device Support
+AC_ARG_ENABLE([devtpm],
+    [AS_HELP_STRING([--enable-devtpm],[Enable use of TPM through the Linux kernel driver (default: disabled)])],
+    [ ENABLED_DEVTPM=$enableval ],
+    [ ENABLED_DEVTPM=no ]
+    )
+
+if test "x$ENABLED_DEVTPM" = "xyes"
+then
+    AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_LINUX_DEV"
+fi
+
+
 # STM ST33 Support
 AC_ARG_ENABLE([st33],
     [AS_HELP_STRING([--enable-st33],[Enable ST33 TPM Support (default: disabled)])],
--- a/configure.ac
+++ b/configure.ac
@@ -242,6 +242,7 @@ AM_CONDITIONAL([BUILD_ADVIO], [test "x$ENABLED_ADVIO" = "xyes"])
 AM_CONDITIONAL([BUILD_ST33], [test "x$ENABLED_ST33" = "xyes"])
 AM_CONDITIONAL([BUILD_MCHP], [test "x$ENABLED_MCHP" = "xyes"])
 AM_CONDITIONAL([BUILD_INFINEON], [test "x$ENABLED_INFINEON" = "xyes"])
+AM_CONDITIONAL([BUILD_DEVTPM], [test "x$ENABLED_DEVTPM" = "xyes"])
 
 
 
@@ -359,3 +360,4 @@ echo "   * Infineon SLB9670           $ENABLED_INFINEON"
 echo "   * STM ST33:                  $ENABLED_ST33"
 echo "   * Microchip ATTPM20:         $ENABLED_MCHP"
 echo "   * I2C:                       $ENABLED_I2C"
+echo "   * Linux kernel TPM device:   $ENABLED_DEVTPM"
--- a/src/include.am
+++ b/src/include.am
@@ -8,8 +8,12 @@ src_libwolftpm_la_SOURCES      = \
                                 src/tpm2.c \
                                 src/tpm2_packet.c \
                                 src/tpm2_tis.c \
-                                src/tpm2_wrap.c \
-                                src/tpm2_linux.c
+                                src/tpm2_wrap.c
+
+if BUILD_DEVTPM
+src_libwolftpm_la_SOURCES      += src/tpm2_linux.c
+endif
+
 src_libwolftpm_la_CFLAGS       = -DBUILDING_WOLFTPM $(AM_CFLAGS)
 src_libwolftpm_la_CPPFLAGS     = -DBUILDING_WOLFTPM $(AM_CPPFLAGS)
 src_libwolftpm_la_LDFLAGS      = ${AM_LDFLAGS} -no-undefined -version-info ${WOLFTPM_LIBRARY_VERSION} 
--- a/src/include.am
+++ b/src/include.am
@@ -5,10 +5,10 @@
 
 lib_LTLIBRARIES+=  src/libwolftpm.la
 src_libwolftpm_la_SOURCES      = \
-								src/tpm2.c \
-								src/tpm2_packet.c \
-								src/tpm2_tis.c \
-								src/tpm2_wrap.c \
+                                src/tpm2.c \
+                                src/tpm2_packet.c \
+                                src/tpm2_tis.c \
+                                src/tpm2_wrap.c \
                                 src/tpm2_linux.c
 src_libwolftpm_la_CFLAGS       = -DBUILDING_WOLFTPM $(AM_CFLAGS)
 src_libwolftpm_la_CPPFLAGS     = -DBUILDING_WOLFTPM $(AM_CPPFLAGS)
--- a/src/include.am
+++ b/src/include.am
@@ -8,7 +8,8 @@ src_libwolftpm_la_SOURCES      = \
 								src/tpm2.c \
 								src/tpm2_packet.c \
 								src/tpm2_tis.c \
-								src/tpm2_wrap.c 
+								src/tpm2_wrap.c \
+                                src/tpm2_linux.c
 src_libwolftpm_la_CFLAGS       = -DBUILDING_WOLFTPM $(AM_CFLAGS)
 src_libwolftpm_la_CPPFLAGS     = -DBUILDING_WOLFTPM $(AM_CPPFLAGS)
 src_libwolftpm_la_LDFLAGS      = ${AM_LDFLAGS} -no-undefined -version-info ${WOLFTPM_LIBRARY_VERSION} 
--- a/src/tpm2_linux.c
+++ b/src/tpm2_linux.c
@@ -94,12 +94,14 @@ int TPM2_LINUX_SendCommand(TPM2_CTX* ctx, byte* cmd, word16 cmdSz)
                 printf("Failed to get a response from fd %d, got errno %d ="
                     "%s\n", fd, errno, strerror(errno));
             }
-        }
-        else {
-        printf("Failed to send the TPM command to fd %d, got errno %d ="
-            "%s\n", fd, errno, strerror(errno));
         #endif
         }
+        #ifdef WOLFTPM_DEBUG_VERBOSE
+        else {
+            printf("Failed to send the TPM command to fd %d, got errno %d ="
+                "%s\n", fd, errno, strerror(errno));
+        }
+        #endif
 
         close(fd);
     }
--- a/src/tpm2_linux.c
+++ b/src/tpm2_linux.c
@@ -1,0 +1,116 @@
+/* tpm2_linux.c
+ *
+ * Copyright (C) 2006-2020 wolfSSL Inc.
+ *
+ * This file is part of wolfTPM.
+ *
+ * wolfTPM is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfTPM is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+
+#include <wolftpm/tpm2_linux.h>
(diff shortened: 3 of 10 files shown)


tomoveu:
Build using "-enable-devtpm"

Attaching logs from the native tests:
* native_test.log is output without debug prints 
* and native_test.debug.log is the wolfTPM output with all debug prints enabled

[native_test.log](https://github.com/wolfSSL/wolfTPM/files/4546249/native_test.log)
[native_test.debug.log](https://github.com/wolfSSL/wolfTPM/files/4546250/native_test.debug.log)

wolftpm/tpm2_linux.h
```diff
@@ -0,0 +1,38 @@
+/* tpm2_linux.h
```
dgarske, line 1: Please add `tpm2_linux.h` to `wolftpm/include.am`.

src/include.am
```diff
@@ -8,7 +8,8 @@ src_libwolftpm_la_SOURCES      = \
 								src/tpm2.c \
 								src/tpm2_packet.c \
 								src/tpm2_tis.c \
-								src/tpm2_wrap.c 
+								src/tpm2_wrap.c \
+                                src/tpm2_linux.c
```
dgarske, line 12: Check indent. Feel free to adjust ones above to use spaces (we prefer spaces).

src/tpm2_linux.c
```diff
@@ -0,0 +1,116 @@
...
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <poll.h>
+#include <errno.h>
+#include <string.h>
+
+
+#define TPM2_LINUX_DEV "/dev/tpm0"
```
dgarske, line 35: Let's wrap this in `#ifndef TPM2_LINUX_DEV`, so users can override it at build-time.

src/tpm2_linux.c
```diff
@@ -0,0 +1,116 @@
...
+    int fd, rspSz;
+    int rc_poll, nfds= 1; /* Polling single TPM dev file */
+    struct pollfd fds;
+
+#ifdef DEBUG_WOLFTPM /* TODO: Change to WOLFTPM_DEBUG_VERBOSE */
+    printf("Command size: %d\n", cmdSz);
+    TPM2_PrintBin(cmd, cmdSz);
+#endif
+
+    fd = open(TPM2_LINUX_DEV, O_RDWR | O_NONBLOCK);
```
dgarske, line 61: With `O_NONBLOCK` is it possible to get an `errno` with `EWOULDBLOCK` or `EAGAIN`? Do we need to retry in that case to resolve concurrent access?

src/tpm2_linux.c
```diff
@@ -0,0 +1,116 @@
...
+
+/* Talk to a TPM device exposed by the Linux tpm_tis driver */
+int TPM2_LINUX_SendCommand(TPM2_CTX* ctx, byte* cmd, word16 cmdSz)
+{
+    int rc = TPM_RC_FAILURE;
+    int fd, rspSz;
+    int rc_poll, nfds= 1; /* Polling single TPM dev file */
+    struct pollfd fds;
+
+#ifdef DEBUG_WOLFTPM /* TODO: Change to WOLFTPM_DEBUG_VERBOSE */
```
dgarske, line 56: Please cleanup the TODO items and use the proper `WOLFTPM_DEBUG_VERBOSE`, `WOLFTPM_DEBUG_IO`, or `WOLFTPM_DEBUG_TIMEOUT`.

src/tpm2_linux.c
```diff
@@ -0,0 +1,116 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+
+#include <wolftpm/tpm2_linux.h>
+#include <wolftpm/tpm2_packet.h>
+#include <wolftpm/tpm2_wrap.h> /* Needed only for WOLFTPM2_MAX_BUFFER */
+#include <sys/types.h>
```
dgarske, line 26: Please wrap this entire file in `#ifdef WOLFTPM_LINUX_DEV`. This will allow someone using a compiler that includes *.c to still build okay on a non-linux system.

configure.ac
```diff
@@ -162,6 +162,19 @@ then
...
+    [AS_HELP_STRING([--enable-devtpm],[Enable use of TPM through the Linux kernel driver (default: disabled)])],
+    [ ENABLED_DEVTPM=$enableval ],
+    [ ENABLED_DEVTPM=no ]
+    )
+
+if test "x$ENABLED_DEVTPM" = "xyes"
+then
+    AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_LINUX_DEV"
+fi
+
```
dgarske, line 176: Please add a conditional below and only include `tpm2_linux.c` when `BUILD_DEVTPM` is set by adding this check in `src/include.am`.

```
AM_CONDITIONAL([BUILD_DEVTPM], [test "x$ENABLED_DEVTPM" = "xyes"])
```

Also add a `# output config summary` entry below after line 361.

dgarske:
@tomoveu , One more please that needs `TPM2_LINUX_SendCommand` used instead of `TPM2_TIS_SendCommand` at `src/tpm2.c:170` ...

```
sudo gdb ./examples/wrap/wrap_test
...
run
...
bt
#0  0x00000000 in ?? ()
#1  0x00023400 in TPM2_TIS_Read (ctx=0x7effe2e0, addr=13893656, result=0x7effa426 "", len=1) at src/tpm2_tis.c:201
#2  0x00023920 in TPM2_TIS_Status (ctx=0x7effe2e0, status=0x7effa426 "") at src/tpm2_tis.c:346
#3  0x00023bac in TPM2_TIS_SendCommand (ctx=0x7effe2e0, cmd=0x7effe32c "\200\002", cmdSz=355) at src/tpm2_tis.c:430
#4  0x00014528 in TPM2_SendCommandAuth (ctx=0x7effe2e0, packet=0x7effa64c, authCnt=1, inHandleCnt=1, outHandleCnt=1) at src/tpm2.c:170
#5  0x00015f58 in TPM2_CreatePrimary (in=0x7effaaec, out=0x7effa67c) at src/tpm2.c:875
#6  0x00024f6c in wolfTPM2_CreatePrimaryKey (dev=0x7effe2e0, key=0x7effdfb8, primaryHandle=1073741835, publicTemplate=0x7effb64c, auth=0x0, authSz=0) at src/tpm2_wrap.c:524
#7  0x0001161c in TPM2_Wrapper_Test (userCtx=0x0) at examples/wrap/wrap_test.c:165
#8  0x00013f1c in main (argc=1, argv=0x7efff764) at examples/wrap/wrap_test.c:838
```

Thanks,
David Garske, wolfSSL

src/tpm2_linux.c
```diff
@@ -0,0 +1,116 @@
...
+    fd = open(TPM2_LINUX_DEV, O_RDWR | O_NONBLOCK);
+    if (fd > 0) {
+        /* Send the TPM command */
+        if (write(fd, cmd, cmdSz) == cmdSz) {
+            fds.fd = fd;
+            fds.events = POLLIN;
+            /* Wait for response to be available */
+            rc_poll = poll(&fds, nfds, TPM2_LINUX_DEV_POLL_TIMEOUT);
+            if (rc_poll > 0 && fds.revents == POLLIN) {
+                rspSz = read(fd, cmd, TPM2_LINUX_DEV_RSP_SIZE);
```
dgarske, line 70: Cast warning:

```
src/tpm2_linux.c:70:25: error: implicit conversion loses integer precision: 'ssize_t' (aka 'long') to 'int' [-Werror,-Wshorten-64-to-32]
                rspSz = read(fd, cmd, TPM2_LINUX_DEV_RSP_SIZE);
                      ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

wolftpm/tpm2_linux.h
tomoveu: Thanks, it is added.

src/include.am
tomoveu: The ones above were not spaces, fixed ;-)

src/tpm2_linux.c
tomoveu: Indeed, I was thinking the same thing, what if someone build bare-metal on Linux without using /dev/tpm0 . It is added.

src/tpm2_linux.c
tomoveu: Concurrent access is solved by the kernel driver. I do not think it is possible to get EWOULDBLOCK or EAGAIN, but definitely worth a test.

src/tpm2_linux.c
tomoveu: I intentionally left them as TODO, so I get feedback if I have selected the proper DEBUG flags. Thanks, the TODOs are now applied.

src/tpm2_linux.c
tomoveu: Indeed, I was thinking the same thing, what if someone build bare-metal on Linux without using /dev/tpm0 . It is added.

src/tpm2_linux.c
tomoveu: Thanks for catching this warning. Made rspSz of type size_t.

src/tpm2_linux.c
tomoveu: Just tested - The kernel returns a standard EBUSY if the TPM char device is already opened by another process. I already have added perror debug message to state when this occurs. There can be only one user per physical TPM, unelss the TPM kernel resource manager is used. Not sure we can do more than have a nice print about the TPM device being busy/already in use.

configure.ac
tomoveu: Output config summary and conditional build check added. Both of these changes go to configure.am, I guess the "src/include.am" mentioning was a typo.

tomoveu:
@dgarske indeed, I was taking a longer look at the Auth command and somehow I did forget to add the new TPM2_Linux_SendCommand. Now the PCR Extend naturally succeeds. Also the native tests execute completely. Will check if there are some errors that need investigating further.

tomoveu:
All feedback is taken in. 
1. rspSz is initalized to 0
2. catching when permissions are needed and having a nice print
3. This print is due to the way native_tests is made. For example, wrap_tests generate:
trm@DFDEV2:~/wolfTPM$ sudo examples/wrap/wrap_test 
TPM2 Demo for Wrapper API's
Mfg NTC (0), Vendor rlsNPCT  , Fw 1.3 (65536), FIPS 140-2 0, CC-EAL4 0
We could modify the native tests to follow the wrap test initial print messages.
4. Updating the README.md is the final and only thing remaining. Doing it right now :)

tomoveu:
@dgarske Thanks, David! Good reviews make a good code great

src/tpm2_linux.c
```diff
@@ -0,0 +1,128 @@
...
+                /* The caller parses the TPM_Packet for correctness */
+                if (rspSz >= TPM2_HEADER_SIZE) {
```
toddouska, line 77: Shouldn't the caller have access to the read() size now in cmd to properly parse?

src/tpm2_linux.c
```diff
@@ -0,0 +1,128 @@
...
+#endif
+
+    fd = open(TPM2_LINUX_DEV, O_RDWR | O_NONBLOCK);
+    if (fd > 0) {
+        /* Send the TPM command */
+        if (write(fd, cmd, cmdSz) == cmdSz) {
+            fds.fd = fd;
+            fds.events = POLLIN;
+            /* Wait for response to be available */
+            rc_poll = poll(&fds, nfds, TPM2_LINUX_DEV_POLL_TIMEOUT);
```
toddouska, line 73: So this always blocks for now.  A nice enhancement would be to support non blocking as well.

src/tpm2_linux.c
tomoveu: The caller can check the integrity of the TPM Packet response based purely on the fields in the packet, as defined in the TIS specification. This is also the way TPM_TIS_SendCommand is implemented, it returns TPM_RC_SUCCESS or TPM_RC_FAILURE.

The reasoning behind this is that the size field in the received TPM2_HEADER has the highest significance for the TIS protocol. And here we do guarantee there are at least TPM2_HEADER_SIZE bytes read for a meaningful Packet payload.

Perhaps the comment could be extended to explain details why "the caller parses the TPM_Packet for correctness". At the same time, a quick look how TPM_LINUX_SendCommand is used shows that too.

If it was up to me I would leave the comment as it is. I am open to suggestions.

src/tpm2_linux.c
tomoveu: Good question. The motivation behind this solution comes from two places:

1. There can be only one user to a physical TPM when using /dev/tpmX, unless the Linux (kernel) TPM resource manager is used. This is by design.

2. I checked the other Linux TSS2 stacks implementations of /dev/tpmX access and they do it in the same way, i.e with blocking access for response.

I think others decided to implement blocking access because /dev/tpmX is meant for explicit process ownership, while /dev/tpmrm is designed for multiple users of a single physical TPM. 

Good to mention: Although /dev/tpmrm is not widely adopted and its maturity is questionable, the kernel TPM resource manage is also meant to solve the problem with other physical limitations of the TPM like key slots. In my opinion, /dev/tpmrm is in heavy development stage.

PR #93, opened 2020-05-13
Adding more testing

tomoveu:
The idea of this PR is to add more TPM2 Native commands testing. It also includes small fix for wolfTPM2 unit testing.

I am also looking into making the testing somewhat functional, I changed the PCR Extend test to be on PCR16, so I can then use it in the PCR_Reset test, which is the first new native test. Here is an example of how the output looks for the user running the native tests:

```
TPM2_PCR_Extend success
TPM2_PCR_Read: Index 16, Count 1
TPM2_PCR_Read: Index 16, Digest Sz 32, Update Counter 129

	bb 22 75 c4 9f 28 ad 52 ca e6 d5 5e 34 a9 74 a5 | ."u..(.R...^4.t.
	8c 7a 3b a2 6f 97 6e 8e cb be 7a 53 69 18 dc 73 | .z;.o.n...zSi..s

TPM2_PCR_Reset command success
PCR Reset: PCR16 value check after reset
PCR Reset: PCR16 read successfully after reset

	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................
	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | ................

```

The user can clearly see the PCR was extended and then cleared, besides just testing the wolfTPM2 implementation of the TPM2 commands.

I am adding more commits to this PR, this is why I publish it as a draft for now. In next commit I am adding TPM2_ReadClock. Stay tuned :)

### DIFF
--- a/examples/native/native_test.c
+++ b/examples/native/native_test.c
@@ -68,6 +68,7 @@ int TPM2_Native_Test(void* userCtx)
         IncrementalSelfTest_In incSelfTest;
         PCR_Read_In pcrRead;
         PCR_Extend_In pcrExtend;
+        PCR_Reset_In pcrReset;
         CreatePrimary_In createPri;
         Create_In create;
         EvictControl_In evict;
@@ -370,7 +371,7 @@ int TPM2_Native_Test(void* userCtx)
     }
 
     /* PCR Extend and Verify */
-    pcrIndex = 0;
+    pcrIndex = 16; /* Working with PCR16 because of next PCR Reset test */
     XMEMSET(&cmdIn.pcrExtend, 0, sizeof(cmdIn.pcrExtend));
     cmdIn.pcrExtend.pcrHandle = pcrIndex;
     cmdIn.pcrExtend.digests.count = 1;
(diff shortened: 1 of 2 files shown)


tomoveu:
This commit was supposed to be in already on Thursday, but I am having difficulty creating an auth sesion for using the Endorsement hierarchy. I spent quite some time and now I am pushing the code, so we can discuss.

GetTime requires Auth Session, because it provides signed timestamp, even though in this case we only want to test with NULL signature.

@dgarske we have primary endorsement and owner hierarchy established at the beginning of the native tests, but somehow I can not make the wolfTPM TPM2_StartAuthSession work the way I need it.

tomoveu:
Some notes from native tests I started doing, but then I realized it is not such a good idea to have

1. Modifying the clock could leave the TPM in bad state for the user afterwards. 

```
WOLFTPM_API TPM_RC TPM2_ClockSet(ClockSet_In* in);
WOLFTPM_API TPM_RC TPM2_ClockRateAdjust(ClockRateAdjust_In* in);
```

Using these two commands would very likely lead to modifying the counter from TPM_Clear and the flag for future time. Not sure if we should mark that we will not native test these?

2. This one `WOLFTPM_API int TPM2_GetHashDigestSize(TPMI_ALG_HASH hashAlg);` is used in the PCR_Extend command, so testing it separately is redundant.

3. Similarly, this one `WOLFTPM_API int TPM2_GetHashType(TPMI_ALG_HASH hashAlg);` is used in the wrap test, so I think it is better to not test natively as well.

tomoveu:
After deep investigation of Policy and Authorization Sessions due to the GetTime we are splitting the work in several PRs

#93 remains fix of unit testing and two new TPM2 Native tests

As result of the GetTime work, a bug was found when composing TPM packets for commands with more than one authorization. Reported to @dgarske  and fixed by #95 

#94 also outcome of the GetTime work, still an ongoing discussion.

#96 "Adding GetTime example" will be something I push later today. Creating a signed timestamp.

Additionally, I also have prepared some wrappers for creating EK SRK and most importantly AIK (Attestation Identity Key) used for GetTime and also Quote.

PR #95, opened 2020-05-26
Fix for TPM2 commands with more than one auth session

dgarske:
`TPM2_ActivateCredential`, `TPM2_EventSequenceComplete`, `TPM2_GetSessionAuditDigest`, `TPM2_GetCommandAuditDigest` and `TPM2_GetTime`

@tomoveu : This is for you. I hope it resolves `TPM2_GetTime`. Looking forward to your review/feedback.

### DIFF
--- a/src/tpm2_packet.c
+++ b/src/tpm2_packet.c
@@ -288,7 +288,7 @@ void TPM2_Packet_PlaceU32(TPM2_Packet* packet, int markSz)
     }
 }
 
-void TPM2_Packet_AppendAuth(TPM2_Packet* packet, TPMS_AUTH_COMMAND* auth, int authCount)
+void TPM2_Packet_AppendAuth(TPM2_Packet* packet, TPMS_AUTH_COMMAND* auth, const int authCount)
 {
     int tmpSz = 0;
     int i;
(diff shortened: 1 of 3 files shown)


tomoveu:
Hey @dgarske 

Yes, before `authorizationSize` was calculated wrong for commands like `GetTime` and the TPM was returning `TPM_RC_AUTH_MISSING`, meaning it does not see all `authValue` appended.

Now the TPM sees all (two) authorizations for `GetTime` and fails with `TPM_RC_BAD_AUTH`, which is another issue to do with enabling the use of the EH. Also, this code makes it clearer how the `authorizationSize` happens

I added this code in the AppendAuth for calculating authorizationSize to check the values.

```
   /* Calculate authorizationSize again */
    for (i = 0; i < authCount; i++) {
        authSize += sizeof(TPMI_SH_AUTH_SESSION);
        authSize += sizeof(UINT16) + auth[i].nonce.size;
        authSize += sizeof(UINT8);
        authSize += sizeof(UINT16) + auth[i].auth.size;
    }
    printf("AuthorizationSize....=%d\n", authSize);

```

As we saw, now AppendAuth calculates properly for both single auth and two auth commands. But I have not reviewed the code in detail. Will do it bit later, so we can merge this :-)

src/tpm2_packet.c
```diff
@@ -288,31 +288,32 @@ void TPM2_Packet_PlaceU32(TPM2_Packet* packet, int markSz)
...
-
-    TPM2_Packet_AppendU32(packet, auth->sessionHandle);
-
-    TPM2_Packet_AppendU16(packet, auth->nonce.size);
-    TPM2_Packet_AppendBytes(packet, auth->nonce.buffer, auth->nonce.size);
-    TPM2_Packet_AppendU8(packet, auth->sessionAttributes);
-    TPM2_Packet_AppendU16(packet, auth->auth.size);
-    TPM2_Packet_AppendBytes(packet, auth->auth.buffer, auth->auth.size);
-
+    for (i=0; i<authCount; i++) {
```
embhorn, line 300: Would it make sense to validate `authCount`?

tomoveu: Actually, yes. authCount should not be bigger than MAX_SESSION_NUM. Do you have some other validation in mind?

PR #96, opened 2020-05-26
Signed timestamp example (AIK and Attestation)

tomoveu:
This is the first step toward enabling wolfTPM to be used for System Integrity and Attestation.

The work here is continuation of #93 but stands on its own as it solves a very interesting problem.

This example creates EK, then sets up a TPM AuthSession for using the Endorsement Hierarchy(EH) under which an Attestation Identity Key(AIK) is to be created. Using this AIK we can attest the TPM time structure. A future use would be using another AIK to generate a quote of the system state.

Currently, the AIK creation fails with strange error for authPolicy or authValue unavailable. This is odd, because we are using PolicySecret to enable the use of the EH, in accordance with the TCG spec and turns out we do it the same way the Intel stack does in their [tpm2_createak](https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_createak.c#L228) tool.

```
TPM2 Demo of generating signed timestamp from the TPM
TPM2_ReadClock: success
TPM2_CreatePrimary: Endorsement 0x80000000 (314 bytes)
TPM2_policySecret success
TPM2_Create RSA failed 0x12f: TPM_RC_AUTH_UNAVAILABLE: The authValue or authPolicy is not available for selected entity
```

So, this example is pivotal for making Authorization Session for EK work with wolfTPM(#95), creating the special type of AIK keys and have the first ever use of wolfTPM for attestation. I am sure there is some tiny detail missing to have this fully working.

Note: I have an alternative version using EK with password authentication (in violation of the TCG spec) and then I am able to create AIK. But the TPM2_GetTime fails with badAuth (probably because of the violation). Still, this was a good test to determine that the params for AIK create are good and the TPM2_Create( AIK ) can succeed. Creating AIK under EH is where things hit an odd error.

In nutshell, it only remains to figure out why  TPM2_Create ( AIK ) is not satisfied by policySecret that  succeeds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ examples/bench/bench
 examples/csr/csr
 examples/tls/tls_client
 examples/pkcs7/pkcs7
+examples/timestamp/signed_timestamp
 pkcs7tpmsigned.p7s
 pkcs7tpmsignedex.p7s
 examples/tls/tls_server
--- a/examples/README.md
+++ b/examples/README.md
@@ -24,7 +24,7 @@ Demonstrates calling the wolfTPM2_* wrapper API's.
 
 Demonstrates creation of Attestation Identity Keys(AIK) and the generation of TPM signed timestamp that can be later used as protected report of the current system uptime.
 
-This example demonstrates the use of authSession(authorization Sesssion) and policySession(Policy authorization) to enable the Endorsement Hierarachy necessary for creating AIK. Then, the AIK is used to issue a TPM2_GetTime command using the wolfTPM2 native API. This provides us with TPM generated and signed timestamp that can be used as a system report of its uptime.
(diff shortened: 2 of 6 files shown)


tomoveu:
@dgarske  two weeks of fighting this thing and I am sure it has all the required components. 

1. Using the EK requires a policy session and we have that (startAuthSession)
2. Policy secret satisfies the policy session for using the EK primary key and creating AIK under the Endorsement Hierarchy (policySecret)
3. EK and AIK have the right key parameters.

So, the odd TPM2_Create() error could even not be TPM related. Do you spot something out of the ordinary in the code?

tomoveu:
OK, found what was the odd thing happening. I am going to quote my commit message of the updated PR, it has all the details:

> AIK is generated using the Endorsement Hierarchy but created under Owner's SRK
>     
>     Generating AIK happens by using the Endorsement Hierarchy, but still
>     the only Hierarchy that can allow loading the key is the Owner Hierarchy
>     through a Storage Key. Using NULL Password Auth Sessions for EK & EH.

@dgarske what do you think, maybe add also some prints for the got time and signatureVerify?

tomoveu:
In separate PR will add wrappers for AIK and SRK. This one I think to finish up by printing the attested time we got.

tomoveu:
@dgarske examples/README.md updated with information about attestation and the new example, and prints added. Ready for review and possibly merge.

PR #99, opened 2020-05-28
If creating an NV and it already exists, set auth and handle anyways

### LINKED ISSUE #98: Can't read from NV memory
Hello,

I want to store 32-bytes of data in TPM's NV memory, then read it back (using an SLB9670). I used your example in wrap_test.c with success. After refactoring my code, I got an error when reading from the NV memory. I tried to restore the changes, and I tried the example code too, but couldn't get it to work. 

Testing code:
```
void test(WOLFTPM2_DEV *dev)
{
    int rc = -1;
    WOLFTPM2_HANDLE parent;
    WOLFTPM2_NV nv;
    word32 nvAttributes = 0;
    WOLFTPM2_BUFFER message, plain;

    XMEMSET(&parent, 0, sizeof(parent));
    parent.hndl = TPM_RH_OWNER;

    rc = wolfTPM2_GetNvAttributesTemplate(parent.hndl, &nvAttributes);
    if (rc != 0) return;

    rc = wolfTPM2_NVCreateAuth(dev, &parent, &nv, 0x01800000,
        nvAttributes, 32, (byte*)gNvAuth, sizeof(gNvAuth)-1);
    if (rc != 0 && rc != TPM_RC_NV_DEFINED) return;
   
   if (rc != TPM_RC_NV_DEFINED) {
        message.size = 32; 
        XMEMSET(message.buffer, 0x42, message.size);
        rc = wolfTPM2_NVWriteAuth(dev, &nv, 0x01800000,
            message.buffer, message.size, 0);
        if (rc != 0) return;
    }

    plain.size = 32;
    rc = wolfTPM2_NVReadAuth(dev, &nv, 0x01800000,
        plain.buffer, (word32*)&plain.size, 0);
    if (rc != 0) return;
}
```

The output it produces:
```
TPM2_NV_DefineSpace failed 332: TPM_RC_NV_DEFINED: NV Index or persistent object already defined
TPM2_NV_Read failed 388: TPM_RC_VALUE: Value is out of range or is not correct for the context
```

However, after clearing the chip, its seems to work once:
```
TPM2_NV_DefineSpace: Auth 0x40000001, Idx 0x1800000, Attribs 0x33947654, Size 64
TPM2_NV_Write: Auth 0x1800000, Idx 0x1800000, Offset 0, Size 32
TPM2_NV_Read: Auth 0x1800000, Idx 0x1800000, Offset 0, Size 32
```
But after I ran my test code again, it will still produces the TPM_RC_VALUE error.

I compiled wolfTPM with v1.7;v1.8 and the latest commit.

What could be the problem?

dgarske:
Fixes #98

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -2040,7 +2040,12 @@ int wolfTPM2_NVCreateAuth(WOLFTPM2_DEV* dev, WOLFTPM2_HANDLE* parent,
     in.publicInfo.nvPublic.dataSize = (UINT16)maxSize;
 
     rc = TPM2_NV_DefineSpace(&in);
-    if (rc != TPM_RC_SUCCESS) {
+    if (rc == TPM_RC_NV_DEFINED) {
+    #ifdef DEBUG_WOLFTPM
+        printf("TPM2_NV_DefineSpace: handle already exists\n");
+    #endif
+    }
+    else if (rc != TPM_RC_SUCCESS) {
     #ifdef DEBUG_WOLFTPM
         printf("TPM2_NV_DefineSpace failed %d: %s\n", rc,
             wolfTPM2_GetRCString(rc));


embhorn:
Looks great, David. Thanks!

PR #100, opened 2020-05-29
Attestation key wrappers

tomoveu:
This PR adds new wolfTPM2 wrappers for easier use by application developers.

First two wrappers are for creating typical and always needed keys like the EK and SRK.
Second set of wrappers are for Attestation purposes, Attestation Identity Keys(AIK) and signed timestamp(GetTime). In addition, the signed_timestamp is updated to use the new wrappers. This has the benefit to highlight the required authSession and Policy details, which i think is important.

List of new wrappers:
* wolfTPM2_CreateEK
* wolfTPM2_CreateSRK
* wolfTPM2_CreateAndLoadAIK
* wolfTPM2_GetTime

Currently the wolfTPM2_GetTime returns the output structure from the TPM2 command as we do not know what the user would want to use the timestamp for since it has more information than just the TPM2/system uptime. I am looking for fedback on what the output structure or variable could be. Maybe we could create wolfTPM2_TIMESTAMP structure?

Looking forward to any and all feedback!

PR 100, Wow =)

### DIFF
--- a/examples/timestamp/signed_timestamp.c
+++ b/examples/timestamp/signed_timestamp.c
@@ -77,6 +77,7 @@ int TPM2_Timestamp_Test(void* userCtx)
         printf("wolfTPM2_Init failed 0x%x: %s\n", rc, TPM2_GetRCString(rc));
         goto exit;
     }
+    printf("wolfTPM2_Init: success\n");
 
 
     /* Define the default session auth that has NULL password */
@@ -100,11 +101,11 @@ int TPM2_Timestamp_Test(void* userCtx)
     /* Create Endorsement Key, also called EK */
     rc = wolfTPM2_CreateEK(&dev, &endorse, TPM_ALG_RSA);
     if (rc != TPM_RC_SUCCESS) {
-        printf("TPM2_CreateEK: Endorsement failed 0x%x: %s\n",
+        printf("wolfTPM2_CreateEK: Endorsement failed 0x%x: %s\n",
             rc, TPM2_GetRCString(rc));
         goto exit;
     }
-    printf("TPM2_CreateEK: Endorsement 0x%x (%d bytes)\n",
+    printf("wolfTPM2_CreateEK: Endorsement 0x%x (%d bytes)\n",
         (word32)endorse.handle.hndl, endorse.pub.size);
 
 
@@ -112,11 +113,11 @@ int TPM2_Timestamp_Test(void* userCtx)
     rc = wolfTPM2_CreateSRK(&dev, &storage, TPM_ALG_RSA, storagePwd,
         sizeof(storagePwd)-1);
     if (rc != TPM_RC_SUCCESS) {
-        printf("TPM2_CreatePrimary: Storage failed 0x%x: %s\n", rc,
+        printf("wolfTPM2_CreateSRK: Storage failed 0x%x: %s\n", rc,
             TPM2_GetRCString(rc));
         goto exit;
     }
-    printf("TPM2_CreatePrimary: Storage 0x%x (%d bytes)\n",
+    printf("wolfTPM2_CreateSRK: Storage 0x%x (%d bytes)\n",
         (word32)storage.handle.hndl, storage.pub.size);
 
 
@@ -172,10 +173,12 @@ int TPM2_Timestamp_Test(void* userCtx)
     rc = wolfTPM2_CreateAndLoadAIK(&dev, &rsaKey, TPM_ALG_RSA, &storage,
         usageAuth, sizeof(usageAuth)-1);
     if (rc != TPM_RC_SUCCESS) {
-        printf("TPM2_Create RSA failed 0x%x: %s\n", rc,
+        printf("wolfTPM2_CreateAndLoadAIK failed 0x%x: %s\n", rc,
             TPM2_GetRCString(rc));
         goto exit;
     }
+    printf("wolfTPM2_CreateAndLoadAIK: AIK 0x%x (%d bytes)\n",
+        (word32)rsaKey.handle.hndl, rsaKey.pub.size);
 
 
     /* set NULL password auth for using EK */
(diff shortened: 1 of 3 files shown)


tomoveu:
I rebased on top of master, so we are ready to merge at any time after review is in :)

Here is the output of the signed_timestamp example using the new wrappers:

```
TPM2 Demo of generating signed timestamp from the TPM
TPM2_ReadClock: success
TPM2_CreateEK: Endorsement 0x80000000 (314 bytes)
TPM2_CreatePrimary: Storage 0x80000001 (282 bytes)
TPM2_policySecret success
TPM2_GetTime: success
TPM2_GetTime: TPMS_TIME_ATTEST_INFO with signature attests:
* TPM Uptime (in ms) since power-up = 4892609020942611888

```

examples/timestamp/signed_timestamp.c
```diff
@@ -36,67 +36,48 @@
...
-        GetTime_In getTime;
-        /* For creating keys */
-        CreatePrimary_In createPri;
-        Create_In create;
         /* For managing TPM session */
         StartAuthSession_In authSes;
         PolicySecret_In policySecret;
-        /* Loading a key and removing objects */
-        Load_In load;
+        /* For removing keys after use */
```
dgarske, line 49: It looks like you can remove `cmdIn` and `cmdOut` now.

examples/timestamp/signed_timestamp.c
```diff
@@ -36,67 +36,48 @@
...
-        Load_Out load;
         byte maxOutput[MAX_RESPONSE_SIZE];
     } cmdOut;
 
     TPM_HANDLE sessionHandle = TPM_RH_NULL;
 
-    TpmKey endorse = { .handle = TPM_RH_NULL }; /* EK  */
-    TpmKey storage = { .handle = TPM_RH_NULL }; /* SRK */
-    TpmKey rsaKey = { .handle = TPM_RH_NULL };  /* AIK */
+    WOLFTPM2_KEY endorse = { .handle.hndl = TPM_RH_NULL }; /* EK  */
```
dgarske, line 64: An XMEMSET works too and ensures entire key is cleared.

tomoveu:
Here is the output of the signed_timestamp example using the new wrappers:

```
TPM2 Demo of generating signed timestamp from the TPM
wolfTPM2_Init: success
TPM2_ReadClock: success
wolfTPM2_CreateEK: Endorsement 0x80000000 (314 bytes)
wolfTPM2_CreateSRK: Storage 0x80000001 (282 bytes)
TPM2_policySecret success
wolfTPM2_CreateAndLoadAIK: AIK 0x80000002 (280 bytes)
wolfTPM2_GetTime: success
TPMS_TIME_ATTEST_INFO with signature attests:
	TPM Uptime (in ms) since power-up = 963972163842237615
```

examples/timestamp/signed_timestamp.c
```diff
@@ -337,27 +219,20 @@ int TPM2_Timestamp_Test(void* userCtx)
     }
 
     /* Close key handles */
-    if (rsaKey.handle != TPM_RH_NULL) {
-        cmdIn.flushCtx.flushHandle = rsaKey.handle;
+    if (rsaKey.handle.hndl != TPM_RH_NULL) {
```
dgarske, line 222: These can be refactored to use `wolfTPM2_UnloadHandle(&dev, &rsaKey.handle);`

src/tpm2_wrap.c
```diff
@@ -3249,6 +3294,110 @@ int wolfTPM2_GetNvAttributesTemplate(TPM_HANDLE auth, word32* nvAttributes)
...
+    getTimeCmd.qualifyingData.size = 0; /* optional */
+    rc = TPM2_GetTime(&getTimeCmd, &getTimeOut);
+    if (rc != TPM_RC_SUCCESS) {
+    #ifdef DEBUG_WOLFTPM
+        printf("TPM2_GetTime failed 0x%x: %s\n", rc, wolfTPM2_GetRCString(rc));
+    #endif
+    }
+
+    /* TODO: Decide what to return. For now, the output structure of GetTime */
+    out = &getTimeOut;
```
dgarske, line 3395: Returning pointer to stack variable is invalid. You should use `out` directly in the `TPM2_GetTime` call and delete `getTimeOut`.

examples/timestamp/signed_timestamp.c
tomoveu: Hm, I thought I did. Strange. Maybe between different branch versions it remained. Thanks for catching this leftover.

examples/timestamp/signed_timestamp.c
tomoveu: I followed the pattern used in native_tests, where we have handles explicitly set. Are you saying to use XMEMSET on all handles before wolfTPM2_Init()?

examples/timestamp/signed_timestamp.c
tomoveu: Good idea ! I like it

src/tpm2_wrap.c
tomoveu: Indeed. Thanks

examples/timestamp/signed_timestamp.c
dgarske: Your named init to TPM_RH_NULL works, but to ensure best forward compatiblity with code the memset is best. The memset should be done early, so any failure will know the handle hasn't been setup when `wolfTPM2_UnloadHandle` is called in the failure case at the end.

tomoveu: Indeed. Before we init just a handle, now we have a whole structure that has more members and sub(member) structures.

examples/timestamp/signed_timestamp.c
tomoveu: Wait @dgarske  I have removed all possible cmdIn and cmdOut. I can not remove it completely, because we need it for the policySecret, authSesssion (creating and flushing it).

PR #102, opened 2020-06-29
Added NPCT75x Nuvoton support and dynamic module detection support

dgarske:
* Added NPCT75x Nuvoton support (`--enable-nuvoton`).
* Added dynamic module detection at run-time when using internal TIS. (on by default when no module is specified `--enable-autodetect`.
* Added some missing --enable options and documentation.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -199,6 +199,16 @@ then
     AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_MCHP"
 fi
 
+# Nuvoton NPCT65x/NPCT75x
+AC_ARG_ENABLE([nuvoton],
+    [AS_HELP_STRING([--enable-nuvoton],[Enable Nuvoton NPCT65x/NPCT75x TPM Support (default: disabled)])],
+    [ ENABLED_NUVOTON=$enableval ],
+    [ ENABLED_NUVOTON=no ]
+    )
+if test "x$ENABLED_NUVOTON" = "xyes"
+then
+    AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_NUVOTON"
+fi
 
 # Infineon SLB9670
 ENABLED_INFINEON=no
@@ -243,6 +253,7 @@ AM_CONDITIONAL([BUILD_ST33], [test "x$ENABLED_ST33" = "xyes"])
 AM_CONDITIONAL([BUILD_MCHP], [test "x$ENABLED_MCHP" = "xyes"])
 AM_CONDITIONAL([BUILD_INFINEON], [test "x$ENABLED_INFINEON" = "xyes"])
 AM_CONDITIONAL([BUILD_DEVTPM], [test "x$ENABLED_DEVTPM" = "xyes"])
+AM_CONDITIONAL([BUILD_NUVOTON], [test "x$ENABLED_NUVOTON" = "xyes"])
 
 
 
@@ -361,3 +372,4 @@ echo "   * STM ST33:                  $ENABLED_ST33"
 echo "   * Microchip ATTPM20:         $ENABLED_MCHP"
 echo "   * I2C:                       $ENABLED_I2C"
 echo "   * Linux kernel TPM device:   $ENABLED_DEVTPM"
+echo "   * Nuvoton NPCT75x:           $ENABLED_NUVOTON"
--- a/configure.ac
+++ b/configure.ac
@@ -94,9 +94,14 @@ else
 fi
 
 # Verbose Logging
-if test "x$ax_enable_debug" = "xverbose"
+if test "x$ax_enable_debug" = "xverbose" || test "x$ax_enable_debug" = "xio"
 then
     AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_DEBUG_VERBOSE"
+fi
+# IO Logging
+if test "x$ax_enable_debug" = "xio"
+then
+    AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_DEBUG_IO"
 fi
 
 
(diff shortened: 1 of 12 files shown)


dgarske:
@tomoveu: Looking forward to your feedback on this PR.

examples/native/native_test.c
```diff
@@ -1195,17 +1187,19 @@ int TPM2_Native_Test(void* userCtx)
...
-    cmdIn.setCmdSet.authHandle = TPM_RH_PLATFORM;
-    cmdIn.setCmdSet.commandCode = TPM_CC_EncryptDecrypt2;
-    cmdIn.setCmdSet.enableFlag = 1;
-    rc = TPM2_SetCommandSet(&cmdIn.setCmdSet);
-    if (rc != TPM_RC_SUCCESS) {
-        printf("TPM2_SetCommandSet failed 0x%x: %s\n", rc,
-            TPM2_GetRCString(rc));
-        goto exit;
+#if defined(WOLFTPM_ST33) || defined(WOLFTPM_AUTODETECT)
+    if (TPM2_GetVendorID() == TPM_VENDOR_STM) {
```
tomoveu, line 1191: EncryptDecrypt2 native test is working only with ST33?

examples/tpm_io.c
```diff
@@ -62,15 +62,25 @@
...
+            #define TPM2_SPI_DEV_CS "0"
         #else
             /* OPTIGA SLB9670 and LetsTrust TPM use CE1 */
-            #define TPM2_SPI_DEV "/dev/spidev0.1"
+            #define TPM2_SPI_DEV_CS "1"
+        #endif
+
+        #ifdef WOLFTPM_AUTODETECT
+            #undef TPM2_SPI_DEV
+            static char TPM2_SPI_DEV[] = "/dev/spidev0.0";
```
tomoveu, line 80: Adding a short comment here that the autodetect changes the CS might be nice. Because initially I was confused why the autodetec has a fixed/hardcoded value, then saw the string manipulation later ;-)

examples/tpm_io.c
```diff
@@ -211,6 +221,10 @@
         int mode = 0; /* Mode 0 (CPOL=0, CPHA=0) */
         int bits_per_word = 8; /* 8-bits */
 
+    #ifdef WOLFTPM_AUTODETECT
+    tryagain:
```
tomoveu, line 225: This label & later goto make sense. Just to be thorough, what about having the count of supported modules (right now 4) and having a simpler for/while cycle going through detection?
My consideration is only one - With time, the number of supported modules will rise and having that list as an actual array(programming list) to go through might be useful in future.
So, this is more of an enhancement rather than request for change. Because the goto here makes sense, keeps things simple, changes CS0 and CS1, can detect everything and anything.

examples/tpm_io.c
```diff
@@ -272,6 +286,26 @@
...
+        if (!foundSpiDev) {
+            if (rxBuf[0] != 0xFF) {
+        #ifdef DEBUG_WOLFTPM
+                printf("Found TPM @ %s\n", TPM2_SPI_DEV);
+        #endif
+                foundSpiDev = 1;
+            }
+            else {
+                int devLen = (int)XSTRLEN(TPM2_SPI_DEV);
+                if (TPM2_SPI_DEV[devLen-1] == '0') {
```
tomoveu, line 301: A comment for this string manipulation would be nice here: "Change from CS0 to CS1 for next round of detection" or something even simpler :)

examples/native/native_test.c
```diff
@@ -1221,16 +1215,9 @@ int TPM2_Native_Test(void* userCtx)
         cmdIn.create.inSensitive.sensitive.userAuth.size);
     cmdIn.create.inPublic.publicArea.type = TPM_ALG_SYMCIPHER;
     cmdIn.create.inPublic.publicArea.nameAlg = TPM_ALG_SHA256;
-#ifdef WOLFTPM_MCHP
-    /* workaround for issue with both sign and decrypt being set */
```
tomoveu, line 1225: marking this for comment in the PR about EncryptDecrypt, TPMA_OBJECT_Sign/decrypt and MCHP

examples/native/native_test.c
```diff
@@ -1221,16 +1215,9 @@ int TPM2_Native_Test(void* userCtx)
...
     cmdIn.create.inPublic.publicArea.objectAttributes = (
         TPMA_OBJECT_sensitiveDataOrigin | TPMA_OBJECT_userWithAuth |
         TPMA_OBJECT_noDA | TPMA_OBJECT_decrypt);
-#else
-    cmdIn.create.inPublic.publicArea.objectAttributes = (
-        TPMA_OBJECT_sensitiveDataOrigin | TPMA_OBJECT_userWithAuth |
-        TPMA_OBJECT_noDA | TPMA_OBJECT_sign | TPMA_OBJECT_decrypt);
```
tomoveu, line 1232: Marking this for further comment in the PR - "Generating a key for both signing/encryption and decryption should be possible."

tomoveu:
@dgarske  **one thing for the configure script**

Autodetect requires wait states, correct?

But when autodetect is enabled the configure script shows wait states are not enabled.

Perhaps some more tweaking of the configure.ac script to print the correct info.

examples/tpm_io.c
```diff
@@ -272,6 +286,26 @@
...
+                    TPM2_SPI_DEV[devLen-1] = '1';
+                    goto tryagain;
+                }
```
tomoveu, line 304: There are whitespaces at this line.

examples/native/native_test.c
dgarske: This is a vendor specific command provided by ST for enabling the TPM2_EncryptDecrypt2 command...

examples/native/native_test.c
dgarske: I retested with all module and it did not need both sign/decrypt set. So this seems like a good change to make for all.

dgarske:
@tomoveu : Thanks for the peer review! I pushed your peer review feedback items. Let me know if you see anything else.

@embhorn : Would you mind taking the final pass on this one?

configure.ac
```diff
@@ -356,8 +433,13 @@ echo "   * Wrappers:                  $ENABLED_WRAPPER"
...
-echo "   * Infineon SLB9670           $ENABLED_INFINEON"
-echo "   * STM ST33:                  $ENABLED_ST33"
-echo "   * Microchip ATTPM20:         $ENABLED_MCHP"
 echo "   * I2C:                       $ENABLED_I2C"
 echo "   * Linux kernel TPM device:   $ENABLED_DEVTPM"
+echo "   * TIS/SPI Check Wait State:  $ENABLED_CHECKWAITSTATE"
+
+echo "   * Infineon SLB9670           $ENABLED_INFINEON"
+echo "   * STM ST33:                  $ENABLED_ST"
+echo "   * Microchip ATTPM20:         $ENABLED_MCHP"
```
embhorn, line 442: Should this be displaying the value of `ENABLED_MICROCHIP`?

dgarske:
@tomoveu : Would you mind testing this again? All tests are passing here for me, but @embhorn is having trouble. Curious if you are still seeing the /dev/spidev0.0 issue.

tomoveu:
@dgarske sure, I could do it tomorrow, because it is 1:40am right now here :)

Could you please remind me if I need to change something regarding the GPIO/CSx for LetsTrust on PI?

AFAIK and can see from the code, the autodetect should work out of the box as long as I have disabled the TPM TIS module for Infineon, correct?

I can confirm the TPM TIS is disabled. I am using PI4.

dgarske:
@tomoveu : No rush. Nothing I had to do besides disable /dev/tpm0. I'm just using the stock let's trust module with no modifications on the Pi 3. It appears on this Pi board all the modules I tested with responded on CS 0 (dev/spidev0.0), but that seems odd. Let me know if you find anything strange and I'll do some more testing.

PR #118, opened 2020-09-02
Fix examples to use same AIK for persistent key

dgarske:
Use the AIK template for the examples, so it can also be used for the signed timestamp example.
Added note about TPM2_Clear and lockout.

### DIFF
--- a/examples/csr/csr.c
+++ b/examples/csr/csr.c
@@ -199,12 +199,10 @@ int TPM2_CSR_Example(void* userCtx)
     }
 
 #ifndef NO_RSA
-    /* Create/Load RSA key for CSR */
+    /* Create/Load RSA key for CSR (AIK) */
     rc = wolfTPM2_ReadPublicKey(&dev, &rsaKey, TPM2_DEMO_RSA_KEY_HANDLE);
     if (rc != 0) {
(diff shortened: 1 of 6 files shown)


tomoveu:
@dgarske I am a bit confused by the commit message. 

"Use the AIK template for the examples." - I understand this.

"so it can also be used for the signed timestamp example" - Here I got lost, because it is already used. If I remmeber correctly, the AIK template came while making the signed timestamp example.

I think that making the AIK persistent is a really nice improvement. Would definitely save time for the user when running the examples. Making all applicable examples use the AIK template, also a nice improvement, so we have the same style in the examples. 

Could I recommend updating the commit message?

dgarske:
Hi @tomoveu , I could not get this to work. Still trying to find a solution.

tomoveu:
@dgarske  could you please update me on what was the end goal here?

Was it to have TPM Keys that link back to the TPM they were generated from and these same keys to be used for CSR operatoins?

I want to take a closer look.

dgarske:
The keys in these examples have different object attributes and a generic AIK cannot be used for the CSR. The goal was to use the already generated and stored key to improve run-time. I have a feeling going forward we won't be storing keys for the examples due to the limited NV on some TPM's. For now I am going to close this PR.

PR #119, opened 2020-09-15
Improve the TLS client mutual authentication example to clarify use of dummy key

dgarske:
ZD 10895.


embhorn (commented):
Building wolfssl with the CSR instructions does not work.

```
    printf("Wrapper/CertReq/CryptoDev code not compiled in\n");
    printf("Build wolfssl with ./configure --enable-certgen --enable-certreq --enable-certext --enable-cryptodev\n");
```
Could this be updated now that crypto-dev is deprecated?

I used:
`./configure --enable-certgen --enable-certreq --enable-certext --enable-cryptocb`

embhorn (approved):
Excellent, thanks David!

PR #122, opened 2020-09-22
Add XOR parameter encryption

tomoveu:
Hi, so here is lots to unload

Here are all the moving parts:
a) Upstream the new tpm2_param_enc.c/h layer together with the tpm2.c changes to wolfTPM's core
b) Upstream encrypt parameter in the request command
c) Upstream decrypt parameter in the response
d) TPM2.0 Quote with Parameter Encryption over the User Data
e) Key generation tool for creating and storing TPM2.0 Keys with Parameter Encryption
f) wolfTPM Tools, first one is flush to clear transient objects, such as TPM2.0 keys and sessions

**Step 1. a)+b)+d)**
Step 2. c)+f)
Step 3. e)

Originally, my wish was all in one take, but unfortunately I hit several unexpected issues
* TPM_ST_SESSION tag not matched + additional bug in helper (reported on GitHub)
* parsing of authSize while marshaled/appended
* parsing of Parameter Size also while marshaled and wrong type (U32->U16), twice the fun
* TPM_RC_SEQUENCE, error that no one seems to ever have had according to Google, but I managed to make the TPM produce it. The issue turned out to be related to the loading of existing SRK and TPM2_ReadPublic (to be reported on GitHub in detail). For the moment I just generate SRK every time like before, so the focus can be on parameter encryption. 

_Now that all of these are solved_, I hit a strange segfault on TPM2_CTX that has nothing to do with parameter encryption, but is preventing the completion of it. Hopeful, I juts have "window vision" and is something silly. Here is the final struggle:

> 
> using GetActiveCTX in TPM2_ParamEnc_CmdRequest works
> TPM2_ParamEnc_CmdRequest calls TPM2_ParamEnc_XOR
> using GetActiveCTX in TPM2_ParamEnc_XOR segfaults

The routines for generating encrypted session and encrypted parameter are there and working. It is just this bizarre segfault that occur yesterday after solving the TPM_RC_SEQUENCE error.

@dgarske lend me a fresh pair of eyes :)

### DIFF
--- a/examples/pcr/quote_paramenc.c
+++ b/examples/pcr/quote_paramenc.c
@@ -1,0 +1,300 @@
+/* quote_paramenc.c
+ *
+ * Copyright (C) 2006-2020 wolfSSL Inc.
+ *
+ * This file is part of wolfTPM.
+ *
+ * wolfTPM is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfTPM is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/* This example shows how to generate a TPM2.0 Quote that holds a signed
+ * PCR measurement. PCR values are used as basis for system integrity.
+ */
+
+#include <wolftpm/tpm2_wrap.h>
+
+#include <examples/pcr/quote.h>
+#include <examples/tpm_io.h>
+#include <examples/tpm_test.h>
+
+#include <stdio.h>
+#include <stdlib.h> /* atoi */
+
+const char defaultFilename[] = "quote.blob\0";
+const char userData[] = "ThisIsData";
+
+/******************************************************************************/
+/* --- BEGIN TPM2.0 Quote Test with Parameter Encryption over user data   --- */
+/******************************************************************************/
+
+static void usage(void)
+{
+    printf("Expected usage:\n");
+    printf("./examples/pcr/quote [pcr] [filename]\n");
+    printf("* pcr is a PCR index between 0-23 (default %d)\n", TPM2_TEST_PCR);
+    printf("* filename for saving the TPMS_ATTEST structure to a file\n");
+    printf("Demo usage without parameters, generates quote over PCR%d and\n"
+           "saves the output TPMS_ATTEST structure to \"quote.blob\" file.\n",
+           TPM2_TEST_PCR);
+}
+
+int TPM2_Quote_Test(void* userCtx, int argc, char *argv[])
+{
+    int pcrIndex, rc = -1;
+    const char *filename = NULL;
+    BYTE *data = NULL;
+    FILE *quoteBlob = NULL;
+    WOLFTPM2_DEV dev;
+    TPMS_ATTEST attestedData;
+
+    WOLFTPM2_KEY storage; /* SRK */
+    WOLFTPM2_KEY rsaKey;  /* AIK */
+
+    TPMS_AUTH_COMMAND session[MAX_SESSION_NUM];
+    TPM_HANDLE sessionHandle = TPM_RH_NULL;
+
+    union {
+        Quote_In quoteAsk;
+        StartAuthSession_In authSes;
+        FlushContext_In flushCtx;
+        byte maxInput[MAX_COMMAND_SIZE];
+    } cmdIn;
+    union {
+        Quote_Out quoteResult;
+        StartAuthSession_Out authSes;
+        byte maxOutput[MAX_RESPONSE_SIZE];
+    } cmdOut;
+
+    XMEMSET(&storage, 0, sizeof(storage));
+    XMEMSET(&rsaKey, 0, sizeof(rsaKey));
+
+
+    if (argc == 1) {
+        /* Demo usage */
+        pcrIndex = TPM2_TEST_PCR;
+        filename = defaultFilename;
+    }
+    else if (argc == 3) {
+        /* Advanced usage */
+        pcrIndex = atoi(argv[1]);
+        if (pcrIndex < 0 || pcrIndex > 23 || *argv[1] < '0' || *argv[1] > '9') {
+            printf("PCR index is out of range (0-23)\n");
+            usage();
+            goto exit_badargs;
+        }
+        filename = argv[2];
+    }
+    else {
+        printf("Incorrect arguments\n");
+        usage();
+        goto exit_badargs;
+    }
+
+    quoteBlob = fopen(filename, "wb");
+    if (quoteBlob == NULL) {
+        printf("Error opening file %s\n", filename);
+        usage();
+        goto exit_badargs;
+    }
+
+    printf("Demo of TPM2.0 Quote with parameter encryption over user data\n");
+    rc = wolfTPM2_Init(&dev, TPM2_IoCb, userCtx);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("wolfTPM2_Init failed 0x%x: %s\n", rc, TPM2_GetRCString(rc));
+        goto exit;
+    }
+    printf("wolfTPM2_Init: success\n");
+
+
+    /* Define the default session auth that has NULL password */
+    XMEMSET(session, 0, sizeof(session));
+    session[0].sessionHandle = TPM_RS_PW;
+    session[0].auth.size = 0; /* NULL Password */
+    TPM2_SetSessionAuth(session);
+
+
+    /* Create RSA Storage Key, also called SRK */
+    rc = wolfTPM2_CreateSRK(&dev, &storage, TPM_ALG_RSA,
+        (byte*)gStorageKeyAuth, sizeof(gStorageKeyAuth)-1);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("wolfTPM2_CreateSRK: Storage failed 0x%x: %s\n", rc,
+            TPM2_GetRCString(rc));
+        goto exit;
+    }
+    printf("wolfTPM2_CreateSRK: Storage 0x%x (%d bytes)\n",
+        (word32)storage.handle.hndl, storage.pub.size);
+
+
+    /* set session auth for storage key */
+    session[0].auth.size = sizeof(gStorageKeyAuth)-1;
+    XMEMCPY(session[0].auth.buffer, gStorageKeyAuth, session[0].auth.size);
+
+
+    /* Create an RSA key for Attestation purposes */
+    rc = wolfTPM2_CreateAndLoadAIK(&dev, &rsaKey, TPM_ALG_RSA, &storage,
+        (const byte*)gUsageAuth, sizeof(gUsageAuth)-1);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("wolfTPM2_CreateAndLoadAIK failed 0x%x: %s\n", rc,
+            TPM2_GetRCString(rc));
+        goto exit;
+    }
+    printf("wolfTPM2_CreateAndLoadAIK: AIK 0x%x (%d bytes)\n",
+        (word32)rsaKey.handle.hndl, rsaKey.pub.size);
+
+
+   /* Start Auth Session - Unbounded, Unsalted */
+    XMEMSET(&cmdIn.authSes, 0, sizeof(cmdIn.authSes));
+    cmdIn.authSes.sessionType = TPM_SE_POLICY;
+    cmdIn.authSes.tpmKey = TPM_RH_NULL;
+    cmdIn.authSes.bind = TPM_RH_NULL;
+    cmdIn.authSes.symmetric.algorithm = TPM_ALG_XOR;
+    cmdIn.authSes.symmetric.keyBits.xorr = TPM_ALG_SHA256;
+    cmdIn.authSes.symmetric.mode.sym = TPM_ALG_NULL;
+    cmdIn.authSes.encryptedSalt.size = 0;
+    cmdIn.authSes.authHash = TPM_ALG_SHA256;
+    cmdIn.authSes.nonceCaller.size = TPM_SHA256_DIGEST_SIZE;
+    rc = TPM2_GetNonce(cmdIn.authSes.nonceCaller.buffer,
+                       cmdIn.authSes.nonceCaller.size);
+    if (rc < 0) {
+        printf("TPM2_GetNonce failed 0x%x: %s\n", rc,
+            TPM2_GetRCString(rc));
+        goto exit;
+    }
+    rc = TPM2_StartAuthSession(&cmdIn.authSes, &cmdOut.authSes);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("\nTPM2_StartAuthSession failed 0x%x: %s\n", rc, TPM2_GetRCString(rc));
+        goto exit;
+    }
+    sessionHandle = cmdOut.authSes.sessionHandle;
+    printf("TPM2_StartAuthSession: sessionHandle 0x%x\n", (word32)sessionHandle);
+
+
+    /* set auth for using the AIK */
+    session[0].auth.size = sizeof(gUsageAuth)-1;
+    XMEMCPY(session[0].auth.buffer, gUsageAuth, session[0].auth.size);
+
+    /* set session for XOR parameter encryption of the TPM Command */
+    session[1].sessionHandle = sessionHandle;
+    session[1].sessionAttributes = TPMA_SESSION_decrypt |
+                                   TPMA_SESSION_encrypt |
+                                   TPMA_SESSION_continueSession;
+    session[1].symmetric.algorithm = TPM_ALG_XOR;
+    session[1].symmetric.keyBits.xorr = TPM_ALG_SHA256;
+    session[1].authHash = TPM_ALG_SHA256;
+    session[1].auth.size = sizeof(gXorAuth)-1;
+    XMEMCPY(session[1].auth.buffer, gXorAuth, session[1].auth.size);
+    session[1].nonce.size = TPM_SHA256_DIGEST_SIZE;
+    rc = TPM2_GetNonce(session[1].nonce.buffer,
+                       session[1].nonce.size);
+    if (rc < 0) {
+        printf("TPM2_GetNonce failed\n");
+        goto exit;
+    }
+
+    /* Prepare Quote request */
+    XMEMSET(&cmdIn.quoteAsk, 0, sizeof(cmdIn.quoteAsk));
+    XMEMSET(&cmdOut.quoteResult, 0, sizeof(cmdOut.quoteResult));
+    cmdIn.quoteAsk.signHandle = rsaKey.handle.hndl;
+    cmdIn.quoteAsk.inScheme.scheme = TPM_ALG_RSASSA;
+    cmdIn.quoteAsk.inScheme.details.rsassa.hashAlg = TPM_ALG_SHA256;
+    /* This specifies the size of the parameter that will be encrypted */
+    cmdIn.quoteAsk.qualifyingData.size = sizeof(userData);
+    /* This is the data that will be encrypted */
+    XMEMCPY(cmdIn.quoteAsk.qualifyingData.buffer, userData,
+            cmdIn.quoteAsk.qualifyingData.size);
+   /* Choose PCR for signing */
+    TPM2_SetupPCRSel(&cmdIn.quoteAsk.PCRselect, TPM_ALG_SHA256, pcrIndex);
+
+#ifdef WOLFTPM_DEBUG_VERBOSE
+    printf("\n\tCommand with ParamEnc\n");
+#endif
+
+    /* Get the PCR measurement signed by the TPM using the AIK key */
+    rc = TPM2_Quote(&cmdIn.quoteAsk, &cmdOut.quoteResult);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("TPM2_Quote failed 0x%x: %s\n", rc, TPM2_GetRCString(rc));
+        goto exit;
+    }
+    printf("TPM2_Quote: success\n");
+
+    rc = TPM2_ParseAttest(&cmdOut.quoteResult.quoted, &attestedData);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("TPM2_Packet_ParseAttest failed 0x%x: %s\n", rc,
+            TPM2_GetRCString(rc));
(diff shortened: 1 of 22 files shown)


tomoveu:
Once this is solved, and Step1) is upstreamed, I am ready to add Step2) and then Step3) to this PR ;)

Here is my gdb output```

eading symbols from src/.libs/libwolftpm.so.9...
(gdb) break TPM2_ParamEnc_XOR
Breakpoint 1 at 0x1f71b: file src/tpm2_param_enc.c, line 222.
(gdb) r
Starting program: /home/trm/wolfTPM/examples/pcr/.libs/quote_paramenc 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Demo of TPM2.0 Quote with parameter encryption over user data
wolfTPM2_Init: success
TPM2_CreatePrimary: 0x80000000 (282 bytes)
wolfTPM2_CreateSRK: Storage 0x80000000 (282 bytes)
TPM2_Create key: pub 280, priv 222
TPM2_Load Key Handle 0x80000002
wolfTPM2_CreateAndLoadAIK: AIK 0x80000002 (280 bytes)

	Command with ParamEnc

TPM2_Quote: Packet->Pos=49
TPM2_Quote: Parameter Size=11


Breakpoint 1, TPM2_ParamEnc_XOR (sessionParamEnc=32767, encryptedData=0x7ffff7fae07c, paramData=0x0 <usage>, 
    paramSz=0) at src/tpm2_param_enc.c:222
222	{
(gdb) s
223	    TPM_RC rc = TPM_RC_FAILURE;
(gdb) 
228	    UINT32 bits = paramSz * 8;
(gdb) 
231	    ctx = TPM2_GetActiveCtx();
(gdb) 
TPM2_Quote failed 0xb: Unknown
TPM2_FlushContext: Closed handle 0x80000002
TPM2_FlushContext: Closed handle 0x80000000
[Inferior 1 (process 26658) exited with code 013]
(gdb) q

```

If I access ctx>authCmd through a pointer, which was my original code, then it straight segfaults. 

This is bizarre, because in the caller function CTX and authCmd are accessible.

**Comment by tomoveu** [comment]
Last, but not least, I have tried to document the parameter encryption process in the code using comments. There is also a nice intro in tpm2_param_enc about some param enc specifics.

File: src/tpm2.c [hunk 1]
```diff
@@ -2250,7 +2164,9 @@ TPM_RC TPM2_Quote(Quote_In* in, Quote_Out* out)
 
         TPM2_Packet_AppendAuth(&packet, ctx->authCmd, 1);
 
+        printf("\nTPM2_Quote: Packet->Pos=%d\n", packet.pos);
```
dgarske, line 2167: Cleanup to remove debugging. Or if it should stay please add DEBUG_WOLFTPM around.

src/tpm2_param_enc.c
```diff
@@ -0,0 +1,282 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+
+#include <wolftpm/tpm2_param_enc.h>
+#include <wolftpm/tpm2_packet.h>
+
+#include <wolfssl/wolfcrypt/hmac.h>
```
dgarske, line 26: This file needs to be wrapped in `#ifndef WOLFTPM2_NO_WOLFCRYPT`

src/tpm2_param_enc.c
```diff
@@ -0,0 +1,282 @@
...
+    Hmac hmac;
+    word32 counter = 0;
+    int hLen, outLen, lLen = 0;
+    byte *outStream;
+    byte uint32Buf[sizeof(UINT32)];
+
+    if (key == NULL || keyStream == NULL)
+        return BAD_FUNC_ARG;
+
+    if (doOnce != FALSE && (sizeInBits & 7) != 0)
```
dgarske, line 94: Let's avoid using `FALSE` and `TRUE`. Stick with 0 and 1. If you want you can stick with the TPM's def of it "YES" and "NO" in tpm2_types.h.

src/tpm2_param_enc.c
```diff
@@ -0,0 +1,282 @@
...
+#include <wolftpm/tpm2_param_enc.h>
+#include <wolftpm/tpm2_packet.h>
+
+#include <wolfssl/wolfcrypt/hmac.h>
+
+/* Routines for performing TPM Parameter Encryption
+ *
+ * NB: Only TPM2B_DATA parameters can be encrypted
+ *
+ * Only the first parameter of a TPM commmand can be encrypted.
```
dgarske, line 32: Check command spelling.

wolftpm/tpm2_param_enc.h
```diff
@@ -0,0 +1,46 @@
...
+#define _TPM2_PARAM_ENC_H_
+
+#include <wolftpm/tpm2.h>
+#include <wolftpm/tpm2_packet.h>
+
+#ifdef __cplusplus
+    extern "C" {
+#endif
+
+int TPM2_ParamEnc_FindDecryptSession(TPM2_CTX *ctx);
```
dgarske, line 32: Please add `WOLFTPM_LOCAL` on these new API's.

src/tpm2_param_enc.c
```diff
@@ -0,0 +1,282 @@
...
+    TPM2B_MAX_BUFFER mask;
+    TPM2B_DATA authValueKDF;
+    TPM2B_DATA nonceNewer, nonceOlder;
+    UINT32 bits = paramSz * 8;
+    UINT32 i;
+
+    ctx = TPM2_GetActiveCtx();
+
+    authValueKDF.size = ctx->authCmd[sessionParamEnc].auth.size;
+    XMEMCPY(authValueKDF.buffer, ctx->authCmd[sessionParamEnc].auth.buffer, authValueKDF.size);
```
dgarske, line 234: I recommend adding a sanity check on the `authValueKDF.size` before the memcpy. Same with `nonceNewer.size` below.

src/tpm2_param_enc.c
```diff
@@ -0,0 +1,282 @@
...
+{
+    TPM_RC rc = TPM_RC_FAILURE;
+    TPM2_CTX *ctx;
+    TPM2B_MAX_BUFFER mask;
+    TPM2B_DATA authValueKDF;
+    TPM2B_DATA nonceNewer, nonceOlder;
+    UINT32 bits = paramSz * 8;
+    UINT32 i;
+
+    ctx = TPM2_GetActiveCtx();
```
dgarske, line 231: For clarify can we just pass in `TPM2_CTX *ctx` from caller?

examples/pcr/quote_paramenc.c
```diff
@@ -0,0 +1,298 @@
...
+    XMEMCPY(cmdIn.quoteAsk.qualifyingData.buffer, userData,
+            cmdIn.quoteAsk.qualifyingData.size);
+   /* Choose PCR for signing */
+    TPM2_SetupPCRSel(&cmdIn.quoteAsk.PCRselect, TPM_ALG_SHA256, pcrIndex);
+
+
+    /* Get the PCR measurement signed by the TPM using the AIK key */
+    rc = TPM2_Quote(&cmdIn.quoteAsk, &cmdOut.quoteResult);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("TPM2_Quote failed 0x%x: %s\n", rc, TPM2_GetRCString(rc));
```
dgarske, line 225: The example appears to return `0xb: Unknown`. Is that really an error or just positive return code?

dgarske (changes requested):
Excellent progress! Some early feedback.

src/tpm2_param_enc.c
tomoveu: This is what I was originally doing and getting segfault. Therefore, I refactored the code with GetActiveCTX to clearly show the problem. I will push my original code to emerge the segfault again.

src/tpm2_param_enc.c
tomoveu: OK. Will do.

src/tpm2_param_enc.c
tomoveu: Thank you.

src/tpm2_param_enc.c
tomoveu: OK

src/tpm2_param_enc.c
tomoveu: Indeed. Thank you

src/tpm2.c
tomoveu: Intentionally left for now, but you are right I can leave it with DEBUG_WOLFTPM even upstream. Thanks.

examples/pcr/quote_paramenc.c
tomoveu: Because access to CTX fails, it returns 0xb that is not a TPM defined error. I will push my original code for this section to emerge the segfault again.

tomoveu:
> Excellent progress! Some early feedback.

Thank you, David! I appreciate the positive feedback.
We are really close to having parameter encryption :) 

I added back my original approach - Passing directly the authSession for Parameter Encryption instead of Context. Funny enough, I no longer get the segfault, but the program does break upon access at the pointer to the authSession as before.
```

	Command with ParamEnc

TPM2_Quote: Packet->Pos=49
TPM2_Quote: Parameter Size=11


Breakpoint 1, TPM2_ParamEnc_XOR (session=0x0 <usage>, encryptedData=0xffffffffffffffff, 
    paramData=0x7fff00000000 <error: Cannot access memory at address 0x7fff00000000>, 
    paramSz=32767) at src/tpm2_param_enc.c:222
warning: Source file is more recent than executable.
222	{
(gdb) s
223	    TPM_RC rc = TPM_RC_FAILURE;
(gdb) s
227	    UINT32 bits = paramSz * 8;
(gdb) s
230	    authValueKDF.size = session->auth.size;
(gdb) s
TPM2_Quote failed 0xb: Unknown
TPM2_FlushContext: Closed handle 0x80000002
TPM2_FlushContext: Closed handle 0x80000000
[Inferior 1 (process 6005) exited with code 013]
(gdb) 

```

Why is this bizarre?

Because in the caller of TPM2_XOR_ParamEnc, we have **if (session->symmetric.algorithm == TPM_ALG_XOR)** and access is OK. The moment we enter TPM2_XOR_ParamEnc , access to ctx OR authCMD or session pointer just does not work as expected. Before it even segfaults. I am surprised that it ends with an error instead.

tomoveu:
After quick screen share with @dgarske , it turned out that (a) there might be a stack issue and (b) when I was stripping down code to pin point the TPM_RC_SEQUENCE & segfault issue,  I left one error check that aborts execution.

The error code 0xb is actually the length of encrypted data, which is exactly 11 (userData is a string of length 11), used as arbitrary User Data for the Quote with Parameter Encryption example.

Will add back the rest of the code for Step1.

Thank you @dgarske for the fresh eyes and useful feedback!

tomoveu:
Hey @dgarske , just added back the code for replacing the parameter with its encrypted data and here is the successful result of a TPM2.0 Quote with XOR Parameter Encryption :) 

"ThisIsData" is the UserData passed as First Parameter to TPM2.0 Quote command

[log.txt](https://github.com/wolfSSL/wolfTPM/files/5269227/log.txt)

The data is being encrypted with different keys and nonce, as seen from this diff/Meld screenshot, where I took the complete output of two different runs of the new example/pcr/quote_paramenc

Screenshot from 2020-09-23 18-23-48

Once you review and confirm, I will also add the early feedback and then move to adding back the code for Step2. I just wanted to keep the code relative to param enc as it was until I get the complete Step1 in the PR ;-) 

For completeness I am attaching and the two output logs that I compared. 
* [log2.txt](https://github.com/wolfSSL/wolfTPM/files/5269228/log2.txt)
* [log.txt](https://github.com/wolfSSL/wolfTPM/files/5269229/log.txt)

tomoveu:
Early feedback taken in. Output when wolfcrypt is disabled (or not available):
```
$ sudo examples/pcr/quote_paramenc
Demo of TPM2.0 Quote with parameter encryption over user data
wolfTPM2_Init: success
TPM2_CreatePrimary: 0x80000000 (282 bytes)
wolfTPM2_CreateSRK: Storage 0x80000000 (282 bytes)
TPM2_Create key: pub 280, priv 222
TPM2_Load Key Handle 0x80000002
wolfTPM2_CreateAndLoadAIK: AIK 0x80000002 (280 bytes)
Encrypted data differs in size = -1
Parameter data original size is = 11
TPM2_Quote failed 0x101: TPM_RC_FAILURE: Commands not being accepted because of a TPM failure
TPM2_FlushContext: Closed handle 0x80000002
TPM2_FlushContext: Closed handle 0x80000000
```

I added a KDFa dummy that returns always -1 to generate an error. Probably good to make some print about wolfcrypt. Open to ideas.

src/tpm2_param_enc.c
```diff
@@ -0,0 +1,306 @@
...
+
+    for (; outLen > 0; outLen -= hLen)
+    {
+        if (hLen > outLen)
+            hLen = outLen;
+
+        counter++;
+
+        /* start HMAC */
+        ret = wc_HmacSetKey(&hmac, hashType, &key->buffer[0], key->size);
```
dgarske, line 128:
```
src/tpm2_param_enc.c:128:15: warning: Function call argument is an uninitialized value
        ret = wc_HmacSetKey(&hmac, hashType, &key->buffer[0], key->size);
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/tpm2_param_enc.c
```diff
@@ -0,0 +1,306 @@
...
+        /* add label */
+        if (label != NULL) {
+            ret = wc_HmacUpdate(&hmac, (byte*)label, lLen);
+            if (ret != 0)
+                break;
+        }
+
+        /* add contextU */
+        if (contextU != NULL) {
+            ret = wc_HmacUpdate(&hmac, contextU->buffer, contextU->size);
```
dgarske, line 147:
```
src/tpm2_param_enc.c:147:19: warning: Function call argument is an uninitialized value
            ret = wc_HmacUpdate(&hmac, contextU->buffer, contextU->size);
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dgarske (approved):
@tomoveu ,

The current state is approved. This PR correctly sends XOR encrypted commands using the `examples/pcr/quote_paramenc`. Tested with /dev/tpm0 and /dev/spidev and TPM simulator. 

Note: I pushed a fix to correct the `packet->pos` that will resolve a future issue when the TPM simulator PR is adding.

As requested, not merged yet. You will be adding the response decryption support and a key gen example. Then AES CFB.

Thanks

tomoveu:
The parameter encryption for command response is pushed together with new tool 'flush' for clearing TPM transient objects.

The current state is that the XOR encryption as is works, however there seemed to be a stack issue. I was getting randomly _segfault_ at end of execution or _stack smashing detected_ from gdb. 

Refactoring the code and minimizing the stack usage for KDFa about three times did help to eliminate the stack issues. Now, the error is TPM_RC_SEQUENCE at StartAuthSession, although the parameters are not changed. Only the stack usage was changed.

Pushing the code for review, in case there is something I missed @dgarske

src/tpm2_param_enc.c
tomoveu: please check if you get this with after the latest commits

src/tpm2_param_enc.c
tomoveu: please check if you are still getting this with the latest commits

tomoveu:
@dgarske  just found something interesting.

Error is
`TPM2_StartAuthSession failed 0x903: TPM_RC_SEQUENCE: Improper use of a sequence handle`

But according to TPM spec 0x903 is TPM_RC_SESSION_MEMORY

Checked in wolfTPM headers
```
src/tpm2.c:            TPM_RC_STR(TPM_RC_SESSION_MEMORY, "Out of memory for session contexts");
wolftpm/tpm2.h:    TPM_RC_SESSION_MEMORY   = RC_WARN + 0x003,
```
Wonder why 0x903 is interpret as TPM_RC_SEQUENCE when it is TPM_RC_SESSION_MEMORY?

tomoveu:
@dgarske I can confirm the error code reported is correct, but the interpretation is wrong.

The actual error is **TPM_RC_SESSION_MEMORY**, because after segfault the session handles are not freed and the TPM runs out of session memory. As discussed, this is a logic bug somewhere in the error parsing of wolfTPM.

As for the response param encryption, I will now investigate this segfault.

I pushed minor fix to the flush tool, so we can clear any TPM loaded sessions remaining after execution.

tomoveu:
I traced the segfault to the finish of the parameter encryption of the Response parameter.

Strangely, TPM2_AppendBytes is used exactly the same way for the Request parameter and causes on issue. Stack usage is reduced. I did one workaround with authValueKDF to save even more stack. Any ideas @dgarske ?


```

Starting program: /home/trm/wolfTPM/examples/pcr/quote_paramenc 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Demo of TPM2.0 Quote with parameter encryption over user data
wolfTPM2_Init: success
TPM2_CreatePrimary: 0x80000000 (282 bytes)
wolfTPM2_CreateSRK: Storage 0x80000000 (282 bytes)
TPM2_Create key: pub 280, priv 222
TPM2_Load Key Handle 0x80000004
wolfTPM2_CreateAndLoadAIK: AIK 0x80000004 (280 bytes)

Breakpoint 1, TPM2_SendCommandAuth (ctx=0x7fffffffd0b0, packet=0x7fffffffa390, authCnt=1, inHandleCnt=1, outHandleCnt=0) at src/tpm2.c:189
189	            TPM2_Packet_AppendBytes(packet, encryptedParameter.buffer,
(gdb) s
TPM2_Packet_AppendBytes (packet=0xda975800, buf=0x18400000000 <error: Cannot access memory at address 0x18400000000>, size=32767) at src/tpm2_packet.c:245
245	{
(gdb) 
246	    if (packet && (packet->pos + size <= packet->size)) {
(gdb) 
247	        if (buf)
(gdb) 
248	            XMEMCPY(&packet->buf[packet->pos], buf, size);
(gdb) 

Program received signal SIGSEGV, Segmentation fault.
__memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:468
468	../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S: No such file or directory.
(gdb) 


```

tomoveu:
Adding two debug commits that helped me investigate. Third one is a one-liner fix that has no relation to the segfault.

@dgarske Didn't the KDFa have constrains on the input data size. For example, input data size has to be ^2 or no more than 1024 bits in one take. Or along those lines. Have to check, but I remember there was something on the original or TCG KDFa.

I look forward to your review

dgarske (approved):
All tests out and looks good now. Holding to merge as I believe you will be adding another example.

tomoveu:
Hey @dgarske 

Thank you! Good catch. I used TPM2B_MAX_BUFFER for the mask, but not for encryptedParameter. 

I pushed the final changes. Just had to make sure the decrypted response parameter is copied in the correct byte order, using Parse instead of Append and everything works like a charm!

Rebasing right now and will ping you for final review and merge :-)

tomoveu:
@dgarske rebased done. Readme files updated. Force pushed the rebased commits. Please do a final review and merge :)

TPM2.0 Quote with param encryption output and visual diff of the two examples (the one in green has Parameter Encryption session and user data that is encrypted).

1. The TPM generates Quote after successful decryption of our encrypted request parameter
2. We decrypt the response parameter that was encrypted by the TPM
3. Output passes the checks for Quote, magic, type, etc.

```

sudo examples/pcr/quote_paramenc 
Demo of TPM2.0 Quote with parameter encryption over user data
wolfTPM2_Init: success
TPM2_CreatePrimary: 0x80000000 (282 bytes)
wolfTPM2_CreateSRK: Storage 0x80000000 (282 bytes)
TPM2_Create key: pub 280, priv 222
TPM2_Load Key Handle 0x80000004
wolfTPM2_CreateAndLoadAIK: AIK 0x80000004 (280 bytes)
TPM2_Quote: success
TPM with signature attests (type 0x8018):
	TPM signed 1 count of PCRs
	PCR digest:
	66 68 7a ad f8 62 bd 77 6c 8f c1 8b 8e 9f 8e 20 | fhz..b.wl...... 
	08 97 14 85 6e e2 33 b3 90 2a 59 1d 0d 5f 29 25 | ....n.3..*Y.._)%
	TPM generated signature:
	54 38 21 a6 17 50 43 14 b5 23 3a ec 44 cc 42 60 | T8!..PC..#:.D.B`
	c4 b3 0c 84 7a 5f 92 48 e3 e8 d0 47 fd bb d4 c6 | ....z_.H...G....
	44 e7 24 d8 48 12 2a b2 49 7d cf 31 1a f6 4a 3e | D.$.H.*.I}.1..J>
	5a 9e 94 e0 71 8d 0f 94 98 80 68 db 45 a6 32 8e | Z...q.....h.E.2.
	d7 fe 97 55 f0 af f4 35 37 01 42 7c ee 2f d6 cc | ...U...57.B|./..
	11 ce 87 2c 3d fd 40 77 fa 93 56 7f 47 3b 37 90 | ...,=.@w..V.G;7.
	0f 1c 99 32 3b b9 25 1a 28 bc 3c 54 cb e1 a3 28 | ...2;.%.(.<T...(
	6b 84 8a 2c 09 b8 ef 02 42 2f de 79 f4 12 bd e6 | k..,....B/.y....
	8d a9 3e b2 77 fb b8 31 1f 24 5b c0 9a 54 82 5d | ..>.w..1.$[..T.]
	29 3f 02 b5 b6 42 5a 82 f2 f8 86 5e ce ce 2d 63 | )?...BZ....^..-c
	ca dc ac ef 9f 86 c6 b8 c5 2a 35 b1 5d 03 6c 00 | .........*5.].l.
	ed bf 75 50 9b 36 bb a2 9d e7 b6 e0 5e 13 78 c4 | ..uP.6......^.x.
	7b d7 42 37 ef f6 9f e2 e0 56 b7 d2 60 08 11 c0 | {.B7.....V..`...
	54 ca 50 9e 52 b4 4e b8 44 38 c2 4f f9 d0 73 ae | T.P.R.N.D8.O..s.
	ec 59 55 6e 0a 71 18 7a 68 77 32 e0 e8 22 1a 49 | .YUn.q.zhw2..".I
	51 d9 c8 ba 27 f8 2f 76 a2 12 a3 a7 4f 1d c8 1f | Q...'./v....O...
TPM2_FlushContext: Closed handle 0x80000004
TPM2_FlushContext: Closed handle 0x80000000

```

Visual diff of `examples/pcr/quote` and `examples/pcr/quote_paramenc`:

Screenshot from 2020-10-01 00-44-17

Will add the keygen example in separate PR. This one is already quite large and complete with the quote_paramenc example.

dgarske (approved):
Thanks Dimi @tomoveu . Excellent work

tomoveu:
Thank you @dgarske Teamwork also played its part here :1st_place_medal: Parameter encryption is a big step forward. Now onto adding AES CFB.

PR #123, opened 2020-09-28
Change public API TPM2_Init_minimal to take user context

elms:


src/tpm2.c
```diff
@@ -388,7 +388,7 @@ TPM_RC TPM2_Init_ex(TPM2_CTX* ctx, TPM2HalIoCb ioCb, void* userCtx,
     return rc;
 }
 
-TPM_RC TPM2_Init_minimal(TPM2_CTX* ctx)
+TPM_RC TPM2_Init_minimal(TPM2_CTX* ctx, void* userCtx)
```
dgarske, line 391: Changes to public API's are not allowed. However this was just added 4/28/20 for the `--enable-devtpm` support and is unlikely to be an issue. The `TPM2_Init_minimal` is used by the wrapper only.

dgarske (approved):
Thanks Jeff. I like this change, but need to confirm breaking this public API is okay before merge.

src/tpm2.c
elms: I misunderstood how `userCtx` is used. My new understanding is that `userCtx` is only used when `ioCb` is called, which is only for `TIS` functions. When `TPM2_Init_minimal` is used by wrap, TIS is not used. @dgarske is that correct?

If so, then I should close this PR and instead open a PR with additional comments and not consider not even building TIS and erroring if `ioCb` or `userCtx` are anything besides NULL when `devtpm` or `swtpm` enabled.

dgarske: That is correct. Agree we can close. Thanks

PR #124, opened 2020-09-28
scripts: Add script to run unit tests with SWTPM

elms:


dgarske (changes requested):
Please add usage instructions to top of script. Please add the script to include.am.

On my Mac when running this I get:

```
$ ./scripts/swtpm_sim.test
./scripts/swtpm_sim.test: line 10: [: -v: unary operator expected
./scripts/swtpm_sim.test: line 13: [: -v: unary operator expected
./scripts/swtpm_sim.test: line 16: [: -v: unary operator expected
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
realpath: missing operand
Try 'realpath --help' for more information.
usage: which [-as] program ...
fatal: cannot change to 'describe': No such file or directory
Building SWTPM:
make: *** /src: No such file or directory.  Stop.
Failed to build
```

Looks like the bash syntax isn't fully portable.

On the Pi when I run it clones, builds and executes the tests successfully.

scripts/swtpm_sim.test
```diff
@@ -0,0 +1,50 @@
...
+    exit 1
+}
+
+build() {
+    if [ ! -d "$SWTPM_DIR" ]; then
+        echo "Cloning SWTPM from $SWTPM_GIT_URL to $SWTPM_DIR";
+        git clone $SWTPM_GIT_URL $SWTPM_DIR || die "unable to clone $SWTPM_GIT_URL for simualtor";
+    fi
+    echo "Building SWTPM: $(git -C "$SWTPM_DIR" describe --long)";
+    make -C $SWTPM_DIR/src > /dev/null || die "Failed to build $SWTPM_DIR";
```
dgarske, line 23: On the Mac this uses the `make -f makefile.mac` file. Can you update, so it builds for Mac too?

```
if [[ "$OSTYPE" == "darwin"* ]]; then
    make -f makefile.mac
fi
```

dgarske: Also might consider "-j4" or something to boost build time.

scripts/swtpm_sim.test
```diff
@@ -0,0 +1,50 @@
...
+
+# check if on path or clone and build
+which ${SWTPM_EXEC:=tpm_server} > /dev/null
+if [ 0 -ne $? ]; then
+    build 
+    SWTPM_EXEC=$SWTPM_DIR/src/$SWTPM_EXEC
+fi;
+
+# start swtpm server
+$SWTPM_EXEC >/dev/null &
```
dgarske, line 39: It might be good to check the return code on this... Although it was nice because I was able to start `tpm_server` myself and run the script and the error was ignored.

dgarske (changes requested):

scripts/swtpm_sim.test
elms: Adding a warning if it exits immediately. Not sure if there is a better way to show return code.

scripts/swtpm_sim.test
elms: Good idea. I can't test on Mac. Hope my approach works.

dgarske (approved):
Thanks Jeff! All looks good and I like the updates for Mac.

PR #126, opened 2020-09-30
Reduce Init_ex code and add documention on `ioCb` and `userCtx`

elms:


dgarske (approved):
Excellent cleanup! Thanks Jeff. Verified all three build modes (spi, devtpm, spidev) with various compilers/analyzers (gcc, scan-build and g++.

Let me know when you are ready to merge.

PR #127, opened 2020-10-07
Windows TBS support

elms:


src/tpm2_wrap.c
```diff
@@ -105,8 +105,8 @@ static int wolfTPM2_Init_ex(TPM2_CTX* ctx, TPM2HalIoCb ioCb, void* userCtx,
 #else
     rc = TPM_RC_SUCCESS;
 #endif /* WOLFTPM_MCHP || WOLFTPM_PERFORM_SELFTEST */
-
-#endif /* ! WOLFTPM_LINUX_DEV */
+    (
```
dgarske, line 108: Was this `(` intentional?

examples/tls/tls_common.h
```diff
@@ -33,6 +33,13 @@
 #include <examples/tls/tls_client.h>
 
 #include <wolfssl/ssl.h>
+#include <wolfssl/wolfio.h>
+
+/* TODO: HACKY for win32 */
```
dgarske, line 38: This workaround looks good.

dgarske (changes requested):

src/tpm2_wrap.c
elms: Nope, definitely an accident. Good catch. Thanks.

examples/timestamp/signed_timestamp.c
```diff
@@ -282,6 +294,10 @@ int TPM2_Timestamp_Test(void* userCtx)
     }
 
     /* Close key handles */
+    if (!tryNVkey) {
+        wolfTPM2_UnloadHandle(&dev, &storage.handle);
```
dgarske, line 298: I would expect you'd still need to unload the handle in either case. If you call  `wolfTPM2_CreateSRK` it will return a valid transitent (temp) handle in `storage.handle.hndl`.

wolftpm/visibility.h
```diff
@@ -31,7 +31,8 @@
 */
 
 #if defined(BUILDING_WOLFTPM)
-    #if defined(_MSC_VER) || defined(__CYGWIN__)
+    #if defined(_MSC_VER) || defined(__CYGWIN__) || \
+      defined(__MINGW32__) || defined(__MINGW64__)
```
dgarske, line 35: Can you add MinGW build instructions to `README.md`?

For ref here is what you sent me:

```
cd wolfssl
export PREFIX=$PWD/../tmp
configure --host=i686 CC=i686-w64-mingw32-gcc CFLAGS="-DWIN32 -DMINGW -D_WIN32_WINNT=0x0600" LIBS="-lws2_32" --prefix=$PREFIX
make
make install
cd ../wolftpm/
./configure --host=i686 CC=i686-w64-mingw32-gcc CFLAGS="-DWIN32 -DMINGW -D_WIN32_WINNT=0x0600" LIBS="-lws2_32 -ltbs" --prefix=$PREFIX --enable-winapi
make
```

examples/tls/tls_common.h
```diff
@@ -311,6 +322,240 @@ static inline void CloseAndCleanupSocket(SockIoCbCtx* sockIoCtx)
         sockIoCtx->listenFd = -1;
     }
 }
+#else /* USE_WOLFSSL_IO */
```
dgarske, line 325: Can we combine this section with the one above? A lot of duplicate code there. I'd prefer we pull in the `USE_WINDOWS_API` logic and socket mappings here directly. I have code for this.

dgarske (changes requested):

wolftpm/visibility.h
elms: I added the winapi specific instructions in `docs/WindowTBS.md` https://github.com/wolfSSL/wolfTPM/pull/127/files/62ff638dc7937d94f19e2ebf96e4b769fcf59113#diff-ff9c2ef8f041e518689e41a95875e1ec7789ae59d7c4bf0560a33df554f99759R41-R52

Do also want those in the top level `README.md`?

examples/tls/tls_common.h
elms: Will do.

examples/timestamp/signed_timestamp.c
elms: It wasn't being unloaded before unless `wolfTPM2_NVStoreKey()` failed. Does that call essentially invalidate the handle?
See above: https://github.com/wolfSSL/wolfTPM/pull/127/files/62ff638dc7937d94f19e2ebf96e4b769fcf59113#diff-be84e1391f63202a3163ecb899bcddc655a28ce91b092f20ab55971aac712190L131-L136

dgarske: I read that wrong, all good here. You are unloading it in the Win TBS case. Awesome

wolftpm/visibility.h
dgarske: Nope, that is perfect. I missed that!

examples/tls/tls_common.h
```diff
@@ -311,6 +322,240 @@ static inline void CloseAndCleanupSocket(SockIoCbCtx* sockIoCtx)
...
+static inline int SockIORecv(WOLFSSL* ssl, char* buff, int sz, void* ctx)
+{
+    SockIoCbCtx* sockCtx = (SockIoCbCtx*)ctx;
+    int recvd;
+    int err;
+
+    (void)ssl;
+
+    /* Receive message from socket */
+    if ((recvd = wolfIO_Send(sockCtx->fd, buff, sz, 0)) == -1) {
```
dgarske, line 348: As discussed this is wrong, should be `wolfIO_Recv`, however you'll be refactoring away the requirement to rely on wolfIO here.

configure.ac
```diff
@@ -196,6 +196,23 @@ then
...
+    )
+
+if test "x$ENABLED_WINAPI" = "xyes"
+then
+    if test "x$ENABLED_DEVTPM" = "xyes" -o "x$ENABLED_SWTPM" = "xyes"
+    then
+        AC_MSG_ERROR([Cannot enable swtpm or devtpm with windows API])
+    fi
+
+    AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_WINAPI -DNO_WOLFSSL_SERVER -DNO_WOLFSSL_CLIENT"
```
elms, line 213: @dgarske I think it will be cleaner to disable the tls examples for `--enable-winapi` in this PR and have a follow on PR. I have a lot of it done, but I was running into permission issues testing with real hardware and realize it may be better to handle as a separate PR.

src/tpm2_winapi.c
```diff
@@ -0,0 +1,104 @@
...
+
+/* Talk to a TPM device using Windows TBS */
+int TPM2_WinApi_SendCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
+{
+    TBS_CONTEXT_PARAMS2 tbs_params;
+    tbs_params.version = TBS_CONTEXT_VERSION_TWO;
+    tbs_params.includeTpm12 = 0;
+    tbs_params.includeTpm20 = 1;
+
+    int rc = 0;
```
dgarske, line 68: Variables need to be declared at top.

src/tpm2_winapi.c
```diff
@@ -0,0 +1,104 @@
...
+
+    if (ctx->winCtx.tbs_context == NULL) {
+        rc = Tbsi_Context_Create((TBS_CONTEXT_PARAMS*)&tbs_params,
+                                 &ctx->winCtx.tbs_context);
+  }
+
+    /* send the command to the device.  Error if the device send fails. */
+    if (rc == 0) {
+        uint32_t tmp = packet->size;
+	rc = Tbsip_Submit_Command(ctx->winCtx.tbs_context,
```
dgarske, line 80: indentation off here and above at 75.

dgarske (changes requested):
Excellent work and thanks for the fixup's. Just a couple minor items. Let me know when its ready for final.

dgarske (approved):
Awesome work on this Jeff. Assigning over to @tomoveu for another set of eyes and he can test as well.

configure.ac
tomoveu: @dgarske  is this an open topic? :)

dgarske: Yes this will be addressed in a new / separate PR that Jeff is already working on here https://github.com/elms/wolfTPM/tree/mingw/tls

examples/timestamp/signed_timestamp.c
```diff
@@ -113,31 +120,36 @@ int TPM2_Timestamp_Test(void* userCtx)
...
+            storage.handle.hndl = TPM2_DEMO_STORAGE_KEY_HANDLE;
+            rc = wolfTPM2_NVDeleteKey(&dev, TPM_RH_OWNER, &storage);
+            if (rc != 0) goto exit;
+            rc = TPM_RC_HANDLE; /* mark handle as missing */
+        }
 #endif
-    if (rc != 0) {
+    }
+
+    if (!tryNVkey || (tryNVkey && rc != 0)) {
```
tomoveu, line 135: maybe we need to document(comment) the tryNVkey variable or this if check here. Took me few seconds to realize this solves the NV persistence issue, but I have context. I wonder if someone without context will understand the idea behind tryNVkey and the rule. Just thinking out loud. This is not a change request.

tomoveu (approved):
@elms very clean and nice implementation. 

@dgarske other than one open question from Jeff to you, I think this is good to merge.

examples/timestamp/signed_timestamp.c
dgarske: @elms , I agree about adding a comment here for what `tryNVkey` is. I had the same thought during my review. Had to double take for what its doing.

Also this flow for `tryNVkey` should be added to the other examples. If you want to follow this up with a new PR that is okay with me.

PR #128, opened 2020-10-08
Fix for properly handling the request parameter size

dgarske:
Fix for properly handling the request parameter size where there isn't one. Fix `TPM2_PrintBin` so its not a recursive function call.

### DIFF
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -86,7 +86,7 @@ static TPM_RC TPM2_SendCommandAuth(TPM2_CTX* ctx, TPM2_Packet* packet,
     TPM_ST tag;
     BYTE *cmd, *param;
     UINT32 cmdSz, authSz;
-    UINT16 requestParamSz;
+    UINT16 requestParamSz = 0;
     UINT32 responseParamSz;
     TPM2B_MAX_BUFFER encryptedParameter;
     int sessionParamEnc;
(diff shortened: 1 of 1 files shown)


tomoveu (approved):
Fix for parameter handling when missing is good. I will think about additional sanity checks for packet size for parameter encryption that can be added.

Change to TPM2_PrintBin looks okay. Definitely better to avoid recursion in this one. I see there was also some whitespace fixes.

src/tpm2.c
```diff
@@ -119,8 +119,10 @@ static TPM_RC TPM2_SendCommandAuth(TPM2_CTX* ctx, TPM2_Packet* packet,
         TPM2_Packet_ParseU32(packet, &authSz);
         /* Pass the Auth Session area, size field already passed by ParseU32 */
         packet->pos += authSz;
-        /* Extract the size of the first parameter */
-        TPM2_Packet_ParseU16(packet, &requestParamSz);
+        if (packet->pos + sizeof(requestParamSz) <= cmdSz) {
```
elms, line 122: If I understand it is valid to have no parameters and therefore have `requestParamSz=0` and this case want' handled previously?

src/tpm2.c
```diff
@@ -5072,34 +5074,39 @@ UINT16 TPM2_GetVendorID(void)
...
         printf("\tNULL");
         return;
     }
 
-    sprintf(line, "\t");
+    while (length > 0) {
+        sz = length;
+        if (sz > LINE_LEN)
+            sz = LINE_LEN;
+        sprintf(line, "\t");
```
elms, line 5089: By using a temp `char*` pointer, I think `sprintf()` is a little less error prone in case someone changes the format strings. Each `sprintf` then `+=` to the pointer for fewer equations.

```suggestion
        line_ptr = line;
        line_ptr += sprintf(line, "\t");
```

elms (commented):
A question and and an idea for further printbin cleanup.
Forgot to actually hit comment before lunch... :grimacing:

src/tpm2.c
dgarske: This was broken in #122 where the assumption was made that a request size was always present, which it is not.

src/tpm2.c
dgarske: Excellent suggestion. Fixes pushed.

elms (approved):

Issue #120, opened 2020-09-18
TPM2_GetTag does not return TPM_ST_SESSIONS when required


tomoveu:
While working on #110 I came across the _"TPM2_GetTag"_ helper function. It is used for about 20 TPM2 commands that can be executed with or without Authorization Session. _"TPM2_GetTag"_  checks for an indication whether the Authorization Area is present. This would then reqiure the TPM_ST_SESSIONS tag. Otherwise, we will get the TPM_ST_NO_SESSIONS tag.

The issue here is that _"TPM2_GetTag"_ will not return TPM_ST_SESSIONS, even when the TPM context(user) has set it. Because there are extra checks for parameter encryption attributes that are not actually needed for Authorization Session check.

_"TPM2_GetTag"_ is always used like this:

```
  st = TPM2_GetTag(ctx);
        if (st == TPM_ST_SESSIONS) {
            TPM2_Packet_AppendAuth(&packet, ctx->authCmd, 1);
        }
```

This is the helper in its current form _"TPM2_GetTag"_ 
```
static TPM_ST TPM2_GetTag(TPM2_CTX* ctx)
{
    TPM_ST st = TPM_ST_NO_SESSIONS;
    if (ctx && ctx->authCmd &&
        (ctx->authCmd->sessionAttributes &
            (TPMA_SESSION_decrypt | TPMA_SESSION_encrypt))) {
        st = TPM_ST_SESSIONS;
    }
    return st;
}
```

The fix is straightforward, making the if rule `"if (ctx && ctx->authCmd)"`. However, when the proper check is made, the native tests break. Therefore, I do not submit a patch with this fix. Right now, this bug makes no user impact, because none of the 20 TPM2 commands has been used with Auth Session in our examples and AFAIK no user has reported the issue. So, we are catching this on time.

@dgarske  After #110 is done, I would like to fix this. Currently, I think we can put it as low to medium priority. Just marking it as found :)

tomoveu:
Solved in #121

tomoveu:
This issue is solved in the new version of TPM2_GetTag that uses the new auth session handling introduced by #133  and additional check added by @dgarske in #129

PR #129, opened 2020-10-09
Add Parameter Encryption support

tomoveu:
Added AES CFB parameter support, salted unbound TPM session

* Added AES CFB support
* Added calculation of command hash and HMAC for sessions.
* Added response HMAC validation.
* Fixes and cleanups for KDFa
* Added KDFa unit test (passes)
* Inlined the param encryption buffers
* Added "-aes" or "-xor" option to some examples to enable parameter encryption.
* Refactor of the session authentication.
* Fixup for the `getPrimaryStoragekey` to allow persisting if not `WOLFTPM_WINAPI`.
* Fixes for nonce and auth count.
* Fix for AES CFB decrypt set key (requires `AES_ENCRYPTION` on SetKey).
* Added support for encrypted RSA salt and salted-unbounded session
* Add innerWrap and outerWrap support for sensitive to private.


tomoveu:
@dgarske  I wanted to make this PR a draft, but I see the option did not apply. Could you please make the change, because I do not have write permission to the wolfTPM repo?

https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/

src/tpm2_param_enc.c
```diff
@@ -44,10 +45,170 @@
  *
  */
 
+/******************************************************************************/
+/* --- Local Variables -- */
+/******************************************************************************/
+wc_Sha sha1digest;
```
dgarske, line 51: Let's not use static global memory. 
Use stack and pass pointer. 

We have wrappers for all this in `wolfssl/wolfcrypt/hash.h`.

```
wc_HashAlg
WOLFSSL_API int wc_HashInit_ex(wc_HashAlg* hash, enum wc_HashType type,
    void* heap, int devId);
WOLFSSL_API int wc_HashInit(wc_HashAlg* hash, enum wc_HashType type);
WOLFSSL_API int wc_HashUpdate(wc_HashAlg* hash, enum wc_HashType type,
    const byte* data, word32 dataSz);
WOLFSSL_API int wc_HashFinal(wc_HashAlg* hash, enum wc_HashType type,
    byte* out);
WOLFSSL_API int wc_HashFree(wc_HashAlg* hash, enum wc_HashType type);
```

dgarske (changes requested):

src/tpm2_param_enc.c
tomoveu: Most definitely will use the wolfCrypt wrappers. 
The global memory was to make it easier to focus on the KDFa issue.
Will fix this. Thank you for the feedback!

src/tpm2_param_enc.c
```diff
@@ -109,82 +107,77 @@ static int TPM2_KDFa(
...
         /* get result */
-        ret = wc_HmacFinal(&hmac, outStream);
+        ret = wc_HmacFinal(&hmac, keyStream);
         if (ret != 0)
-            break;
+            goto exit;
 
-        outStream = &outStream[hLen];
+        counter++;
+        keyStream += hLen;
```
dgarske, line 178: Something missing here for handling the remainder portion that is not digest sized. Let’s discuss Tue.

src/tpm2_param_enc.c
tomoveu: Reverted back to the prior implementation using wolfcrypt HMAC

tomoveu:
@dgarske now that I understand what do to you mean by something is missing here, it makes sense.

If the data is not digest size, it would be unaligned and then this would result in *improper* marshaling(appending), thus the TPM_RC_SENSITVE error that we are now getting would make sense.

Let's look at this together, because I really did get stuck with this KDFa issue.

tomoveu:
Pushed my latest changes to the keygen example. There might be something of value for resemblance between the keygen examples with and without parameter encryption #131

tomoveu:
Opting out from unsalted-unbounded session in favor of salted-bounded session that adds more entropy.

This also helps to simplify the KDFa.

tomoveu:
@dgarske Hi David, after all the fixes and debugging, this should now be working.

There is one enhancement that I need to add in the tpm2.c for handling the different TPM2 commands. For the moment I have only created handling for TPM2_Create and TPM2_Load.

Please review, so we can confirm this stage and then I would just add the enhancement.

tomoveu:
Found issue in the handling of authSessions when we have more than the default/minimum required by a command. Pushed DEBUG commit to demonstrate the issue.

Will add helper to check the set sessions and provide session count that will be used in all TPM2_{Command} helpers.

Once this is solved will check again if the keygen/keyload examples fail, as there could still be an issue with the inner/outer integrity check.

tomoveu:
**Intermediate status update** - Today I added handling of the maximum possible number of TPM sessions. I tried using WOLFTPM2_SESSION and significantly change how auth sessions are handled in wolfTPM, but then I went with this much simpler and cleaner implementation.

**This solved the issue** with not appending the correct number of auth sessions for TPM2_Create and TPM2_Load. And after this fix another auth session issue emerged for parameter encryption. I have been working on solving that new auth session issue and is still ongoing.

**Additionally**, while working on this new issue I noticed that we use dev->session in wolfTPM2_CreateKey and wolfTPM2_LoadKey although everywhere else ctx->authCmd is used. For example, wolfTPM2_CreateKey manipulates dev->authCmd while the later called TPM2_Create uses ctx->authCmd.

tomoveu:
status update: cause for the RC_BAD_AUTH is that the calculated HMAC over the parameter encryption session differs from what the TPM expects. investigating why the difference occurs

tomoveu:
there is something strange happening with the HMAC calculation from the sessionKey on the TPM simulator side.

The authValue prepared for the AIK key generation is compared with the HMAC resulting from the sessionKey.
I am continuing to investigate.

tomoveu:
OK, I think I found the issue.

Our implementation does not use TPM2_PolicyAuthValue and does not include the authValue for the sessionKey calculation. But at least one of these is expected from the TPM (simulator) side. So, when there is no PolicyAuthValue set and no authValue used for setting up parameter encryption, the check on TPM side operates under conditions that must not be and results in an error. 

Ken Goldman helped with pointing out this note in the spec:

TCG Spec, Part 1, Chapter 21 Session-based encryption
> If a session is also being used for authorization, sessionValue (see 21.2 and 21.3) is sessionKey || authValue. The binding of the session is ignored. If the session is not being used for authorization, sessionValue is sessionKey.
> 
> NOTE 1 A policy session that is used for parameter encryption uses authValue to calculate sessionValue even if the policy does not include TPM2_PolicyAuthValue().

I have read Chapter 21 from the spec so many times and somehow glimpsed over Note 1 ... 

Screenshot from 2020-11-18 14-03-05

tomoveu:
While working the new issue, I came up with a simpler way of handling the maximum possible TPM sessions.

I think that using authSession from active TPM context offloads the TPM2 Command wrappers and minimizes the room for error Therefore, I changed all TPM2 wrappers to be using new way of appending the authSession.

tomoveu:
> there is something strange happening with the HMAC calculation from the sessionKey on the TPM simulator side.
> 
> The authValue prepared for the AIK key generation is compared with the HMAC resulting from the sessionKey.
> I am continuing to investigate.

For traceability: After switching from unsalted-unbounded session to a salted session, the hmac/auth field in the TPMS_AUTH_COMMAND on the stack side should also be populated. This explains the behavior observed above. 

Good catch @dgarske :100:

src/tpm2.c
```diff
@@ -191,21 +288,38 @@ static TPM_RC TPM2_SendCommandAuth(TPM2_CTX* ctx, TPM2_Packet* packet,
...
-                return TPM_RC_FAILURE;
+                return rc;
             }
+
             /* packet->pos already points at the first command parameter */
             TPM2_Packet_AppendBytes(packet, encryptedParameter.buffer,
-                                    encryptedParameter.size);
+                                            encryptedParameter.size);
+
+        #ifndef WOLFTPM2_NO_WOLFCRYPT
```
tomoveu, line 308: Could parameter encryption work without this hmac field calculation?
My read: if this hmac filed is indeed necessary for policy or hmac session, then it is necessary for parameter encryption. Because parameter encryption requires at policy session or hmac. I would recommend not just adding build option for param enc, but adding a comment here about this relation.

tomoveu (commented):
@dgarske I like the way you replace the Authorization Area in the packet in TPM2_CommandSendAuth. Regarding the hmac calculation, I could not find the chapter in the TCG spec for the sequence. I would be interested to give you a cross-check on that.

The refactoring looks great.

ps: The hmac field calculation does look like the last missing piece.

src/tpm2.c
```diff
@@ -290,103 +300,68 @@ static int TPM2_CalcHmac(TPM2_CTX* ctx, int inHandleCnt,
...
-        /* Parse first parameter */
-        TPM2_Packet_ParseU16(&packet, &tempSz);
-        encParam += sizeof(UINT16);
+        TPM2_Packet_ParseU16(packet, &tempSz);
+        encParam = param + sizeof(UINT16);
+        encParamSz = tempSz;
+    }
+    else if (info->flags & CMD_FLAG_ENC4) {
+        UINT32 tempSz;
+        TPM2_Packet_ParseU32(packet, &tempSz);
```
tomoveu, line 341: I think this should be also U16, but with offset of U16(+2).
Because I believe this is the case when multiple TPM2B parameters are wrapped as one. Still, the size field of the TPM2B structures is always UINT16. However, we sometimes must extract the inner size field, i.e the size field of the first wrapped parameter
tpm2b-struct{ 
tpm2b.size;
tpm2b.firstparam.size; < we want this
tpm2b.firstparam;
tpm2b.secondparam.size;
tpm2b.secondparam;
}

tomoveu (commented):
@dgarske  I like the way you split the request and response command process. Very logical and easier to parse the code. I pushed some minor fixes so I could test and review.

One comment about the **CmdFlags** use. I think the type for the second case of dec/enc4 should be u16, because the size field of tpm2b structures is always u16.

dgarske:
@tomoveu : This is ready for final review.

tomoveu:
@dgarske OK, I am starting final review and testing

tomoveu:
I have tested and found the following issues:
- native tests must be made compatible with the new session handling **(fixed)**
- some of the issues are introduced due to default auth session set in wolfTPM2 Init wrapper **(fixed)**
* The above completely fixes #120 TPM2_GetTag issue that was also addressed with adding the TPM2_GetSessionAuthCount to the TPM2_GetTag **(#120 can be closed)**
- keyload example fails when parameter encryption is not used (**investigating**, I am close to solving it)

tomoveu:
I pushed my latest changes and all but two native tests should be now working

I am investigating issues with _(the native tests of)_ TPM2_LoadExternal and TPM2_Create when ECDH key is created.

tomoveu:
Fixed TPM2_LoadExternal and HMAC native tests.

I compared the command request of the previously working native tests with the current state and went down the rabbit whole debugging. The storage key auth is set even though it is not needed for TPM2_LoadExternal. Simply disabling the auth session for the TPM2_LoadExternal solved the problem. 

* Although the storage auth is not in a perfect place in the code around LoadExternal, it is later used for other native tests, when TPM_RS_PW is re-enabled in code. The length of the native tests make it difficult to read at times.

tomoveu:
All native tests fixed, but I noticed this diff when comparing the tests output @dgarske I will take a closer look and let's sync

Now
```

TPM2_ECDH_KeyGen: zPt 2, pubPt 60533
TPM2_ECDH_ZGen: zPt 2
...
TPM2_RSA_Encrypt: 256
TPM2_RSA_Decrypt: 2

```
Before
```
TPM2_ECDH_KeyGen: zPt 68, pubPt 68
TPM2_ECDH_ZGen: zPt 68

TPM2_RSA_Encrypt: 256
TPM2_RSA_Decrypt: 68
```

tomoveu:
I finished testing the parameter encryption examples. They all seem to work flawlessly with just one detail.

I noticed that for keyimport the EncSz is 0 and yet param enc is set and our implementation is executed. 

I digged into the TCG Spec and found that for TPM2_Import the first TPM2B parameter is optional. And we are not using it, therefore our keyimport example does not demonstrate in reality parameter encryption when option '-e' is enabled.

```
encryptionKey the optional symmetric encryption key used as the inner
wrapper for duplicate
If symmetricAlg is TPM_ALG_NULL, then this
parameter shall be the Empty Buffer.
```
Awesome work @dgarske I am glad you approve changing core design choices of wolfTPM to get param encryption. We also now have the handling for the maximum auth sessions that opens all doors to enhanced authorization! Just awesome :100:

dgarske (approved):
Approved. Going to assign to @elms for review.

examples/tls/tls_common.h
```diff
@@ -60,7 +60,7 @@
 /* force use of a TLS cipher suite */
 #if 0
     #ifndef TLS_CIPHER_SUITE
-        #define TLS_CIPHER_SUITE "ECDHE-RSA-AES128-SHA256"
+        #define TLS_CIPHER_SUITE "ECDHE-rsa-AES128-SHA256"
```
embhorn, line 63: Was this intentional?

examples/tls/tls_server.c
```diff
@@ -319,7 +374,7 @@ int TPM2_TLS_Server(void* userCtx)
 
 #if 0
     /* Optionally choose the cipher suite */
-    rc = wolfSSL_CTX_set_cipher_list(ctx, "ECDHE-RSA-AES128-GCM-SHA256");
+    rc = wolfSSL_CTX_set_cipher_list(ctx, "ECDHE-rsa-AES128-GCM-SHA256");
```
embhorn, line 377: Was this intentional?

tests/unit_tests.c
```diff
@@ -255,6 +256,35 @@ static void test_wolfTPM2_Cleanup(void)
...
+    };
+    const byte keyExp[TEST_KDFA_KEYSZ] = {
+        0xbb, 0x02, 0x59, 0xe1, 0xc8, 0xba, 0x60, 0x7e, 0x6a, 0x2c,
+        0xd7, 0x04, 0xb6, 0x9a, 0x90, 0x2e, 0x9a, 0xde, 0x84, 0xc4};
+    byte key[TEST_KDFA_KEYSZ];
+
+    rc = TPM2_KDFa(TPM_ALG_SHA256, &keyIn, label, &contextU, &contextV, key, keyIn.size);
+    AssertIntEQ(sizeof(keyExp), rc);
+
+    AssertIntEQ(XMEMCMP(key, keyExp, sizeof(keyExp)), 0);
```
embhorn, line 285: This assert is failing when I run `make check`.

```
./configure --enable-debug=verbose
make && make check
```

src/tpm2_param_enc.c
```diff
@@ -51,11 +52,11 @@
...
  *
- * 'key' or 'authValue' are used to generate the session key.
- * - either 'key' or 'authValue' must be provided.
+ * 'key' input data is used together with the label, ContextU and ContextV to
+ * generate the session key.
  *
- * 'keyStream' points to the buffer storing the generated session key
- * - 'keyStream' can not be NULL.
+ * 'keyStream' points to the buffer storing the generated session key, and
+ * 'keyStream' can not be NULL.
```
embhorn, line 59: `keystream` param was removed

src/tpm2_param_enc.c
```diff
@@ -71,34 +72,29 @@
...
+    int hLen, lLen = 0;
+    byte uint32Buf[sizeof(UINT32)];;
+    UINT32 sizeInBits = keySz * 8, pos;
+    BYTE* keyStream = key;
 
-    if ((key == NULL && authValue == NULL) || keyStream == NULL)
-        return BAD_FUNC_ARG;
-
-    if (doOnce != 0 && (sizeInBits & 7) != 0)
+    if (key == NULL || keyStream == NULL)
```
embhorn, line 97: Check either, not both

embhorn (changes requested):
Wow, this is a great addition! Couple minor changes. Also, seeing failure on `make check` (could be my config).

embhorn (approved):
Excellent work, Dimi and David!

elms:
Testing with `--winapi` and seeing failures. Attempting to compare to head of tree, but will have to come back to that.

`keygen`/`keyload`, `signed_timestamp`, and `quote` work.

<details>
<summary>CSR</summary>

```
./examples/csr/csr.exe
TPM2 CSR Example
TPM2_ReadPublic failed 139: TPM_RC_HANDLE: The handle is not correct for the use
TPM2_CreatePrimary: 0x80ffffff (282 bytes)
Loading SRK: Storage 0x80ffffff (282 bytes)
Reading 727 bytes from rsa_test_blob.raw
TPM2_Load Key Handle 0x80fffffe
Loaded key to 0x80fffffe
CryptoDevCb Hash: Type 4
CryptoDevCb Hash: Type 4
CryptoDevCb Hash: Type 6
CryptoDevCb Hash: Type 6
Failure 0xffffff7d: RSA buffer error, output too small or input too big
TPM2_FlushContext: Closed handle 0x80ffffff
TPM2_FlushContext: Closed handle 0x80fffffe
```
</details>

<details>
<summary>wrap</summary>
```
 ./examples/wrap/wrap_test.exe
TPM2 Demo for Wrapper API's
Mfg STM  (2), Vendor , Fw 73.20 (1151343132), FIPS 140-2 1, CC-EAL4 0
TPM2_FlushContext failed 139: TPM_RC_HANDLE: The handle is not correct for the use
TPM2_FlushContext failed 139: TPM_RC_HANDLE: The handle is not correct for the use
TPM2_FlushContext failed 139: TPM_RC_HANDLE: The handle is not correct for the use
TPM2_CreatePrimary: 0x80ffffff (314 bytes)
TPM2_FlushContext: Closed handle 0x80ffffff
TPM2_ReadPublic failed 139: TPM_RC_HANDLE: The handle is not correct for the use
TPM2_CreatePrimary: 0x80fffffe (282 bytes)
TPM2_EvictControl failed -2144861184: unknown error number
Failure 0x80280400: unknown error number
TPM2_Shutdown failed 0x80280400: unknown error number
wolfTPM2_Shutdown complete
```
</details>

dgarske:
@elms : On the winapi test of the param enc PR. The wrapper is trying to use TPM2_EvictControl, which is not allowed on winapi... That is an easy fix. For the csr one I would need verbose logs to debug cause (--enable-debug=verbose).

elms:
> @elms : On the winapi test of the param enc PR. The wrapper is trying to use TPM2_EvictControl, which is not allowed on winapi... That is an easy fix. For the csr one I would need verbose logs to debug cause (--enable-debug=verbose).

oops I had used `--enable-debug=VERBOSE`

```
./examples/csr/csr.exe
TPM2 CSR Example
TPM2_ReadPublic failed 139: TPM_RC_HANDLE: The handle is not correct for the use
Found 1 auth sessions
CommandProcess: Handles (Auth 1, In 1), CmdSz 355, AuthSz 9, ParamSz 328, EncSz 36
ResponseProcess: Handles (Out 1), RespSz 474, ParamSz 451, DecSz 282, AuthSz 5
TPM2_CreatePrimary: 0x80ffffff (282 bytes)
Loading SRK: Storage 0x80ffffff (282 bytes)
Reading 825 bytes from rsa_test_blob.raw
Found 1 auth sessions
CommandProcess: Handles (Auth 1, In 1), CmdSz 550, AuthSz 41, ParamSz 491, EncSz 207
ResponseProcess: Handles (Out 1), RespSz 59, ParamSz 36, DecSz 34, AuthSz 5
TPM2_Load Key Handle 0x80fffffe
Loaded key to 0x80fffffe
CryptoDevCb Hash: Type 4
CryptoDevCb Hash: Type 4
CryptoDevCb Hash: Type 6
CryptoDevCb Hash: Type 6
CryptoDevCb Pk: Type 1
Found 1 auth sessions
CommandProcess: Handles (Auth 1, In 1), CmdSz 304, AuthSz 24, ParamSz 262, EncSz 256
TPM2_RSA_Decrypt failed 162: TPM_RC_BAD_AUTH: Authorization failure without DA implications
wolfTPM2_CryptoDevCb failed rc = 162
Failure 0xffffff08: Error with hardware crypto use
TPM2_FlushContext: Closed handle 0x80ffffff
TPM2_FlushContext: Closed handle 0x80fffffe
```

Works on head of tree although the keyblobs didn't seem compatible, failed integrity check. Is that expected? Maybe an issue with unexpected auth change?

dgarske:
@dgarske : Feedback from @elms, which I am investigating.
1. The CSR example required re-running the keygen. Did the .blob format change in this PR?
2. The TLS client / server examples with `-aes` are not working.

elms (approved):
I did tests using `--enable-winapi` (far from exhaustive) and they behave the same as before. Read over code (again not exhaustive) and looks good to me.

Awesome work @tomoveu and @dgarske.

PR #132, opened 2020-11-02
Added support for importing external private key to get a key blob

dgarske:
ZD 11163.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -193,7 +193,7 @@ Creating new RSA key...
 Created new key (pub 280, priv 222 bytes)
 Wrote 840 bytes to keyblob.bin
 
-$ ./examples/keygen/keyload keyblob.bin RSA
+$ ./examples/keygen/keyload keyblob.bin
 TPM2.0 Key load example
 Loading SRK: Storage 0x81000200 (282 bytes)
 Reading 840 bytes from keyblob.bin
@@ -207,7 +207,7 @@ Creating new ECC key...
 Created new key (pub 88, priv 126 bytes)
 Wrote 744 bytes to keyblob.bin
 
-$ ./examples/keygen/keyload keyblob.bin ECC
+$ ./examples/keygen/keyload keyblob.bin
 TPM2.0 Key load example
 Loading SRK: Storage 0x81000200 (282 bytes)
 Reading 744 bytes from keyblob.bin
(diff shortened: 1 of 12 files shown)


examples/README.md
```diff
@@ -186,20 +186,59 @@ Performance benchmarks.
...
-$ ./examples/keygen/keyload
+$ ./examples/keygen/keyload keyblob.bin RSA
 TPM2.0 Key load example
-wolfTPM2_Init: success
+Loading SRK: Storage 0x81000200 (282 bytes)
+Reading 840 bytes from keyblob.bin
+Loaded key to 0x80000001
+
 
+$ ./examples/keygen/keyimport keyblob.bin ECC
```
tomoveu, line 233: Once I looked at the code it became clear that we are importing hardcoded RSA key from tpm_test.h, but originally I thought we are important a plain key from keyblob.bin and then overwritting keyblob.bin with the same pair but as a TPM key. Perhaps one additional printf and/or usage when running keyload without parameters.

examples/keygen/keyimport.c
```diff
@@ -0,0 +1,159 @@
...
+    TPMI_ALG_PUBLIC alg = TPM_ALG_RSA; /* TPM_ALG_ECC */
+
+#if !defined(WOLFTPM2_NO_WOLFCRYPT) && !defined(NO_FILESYSTEM)
+    XFILE f;
+    const char* outputFile = "keyblob.bin";
+    size_t fileSz = 0;
+
+    if (argc >= 2) {
+        outputFile = argv[1];
+        if (argc >= 3) {
```
tomoveu, line 53: There is no information to the user what is the default key scheme, when not specified. Maybe adding informative print about the options the tool is running with :)

tomoveu, line 53: There is no information to the user what is the default key scheme, when not specified. Maybe adding informative print about the options the tool is running with

tomoveu (commented):
I could not spot anything out of the ordinary. My only two comments are about missing usage() prints for the user in the two new tools. Functionality wise they work and the code is easy to follow.

tomoveu (approved):
Great addition to the wolfTPM examples! I only had to do some minor changes about the keyload usage prints, whitespaces and README file. This is ready for merge :100:

elms (changes requested):
Some compile isssues with `--disable-wolfcrypt` around unused variables and functions.

elms (approved):

PR #133, opened 2020-11-18
New auth session handling for TPM2 command handlers

tomoveu:
In the run of #129 , we found that wolfTPM currently does not handle the maximum possible TPM sessions. 

I came up with two simple solutions, this one I think is better, because it uses the active TPM context.

It is best to make this change upstream regardless of the status of 129, because the current authSession handling results in TPM2 command handlers to include only the default amount of TPM sessions for a particular command.

For example, for TPM2_Create the handler would include only 1 session at all times regardless of how many sessions the user has set. For TPM2_Quote the handler would include only 2 sessions at all times, regardless if there are 3.

### DIFF
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -279,6 +279,39 @@ TPM_RC TPM2_SetSessionAuth(TPMS_AUTH_COMMAND* cmd)
     return rc;
 }
 
+/* Finds the number of active Auth Session in the given TPM2 context */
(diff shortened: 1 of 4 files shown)


tomoveu:
Native tests pass with flying colors, but I would prefer if we get two reviews on this before merging.

dgarske (approved):
Awesome work on this Dimi. Thanks!

PR #134, opened 2020-12-03
Preparation for wolfTPM v2.0 release

dgarske:


tomoveu:
@dgarske looks good! 

I do not see #133 mentioned. Could be something like: Added support for handling the maximum number of TPM auth sessions. Although, it may not be needed, because 129 extends this change and new auth session handling is mentioned in the releas notes. Just my 2 cents.

embhorn (approved):
Looks great! Thanks, David.

dgarske:
FYI: There is an issue with the PKCS 7 example on the verify. I debugged a bit, but suspect it may be a change in the wolfSSL side. I will debug post release.

PR #135, opened 2020-12-08
Add lpc bus support

tomoveu:
Most servers and data centers use TPM on LPC Bus. These systems have native LPC support and when run under Linux, there would be /dev/tpmX interface. However, with the rise of edge gateways and IIoT(Industrial IoT), there is increasing use of ARM SoC without native LPC support. These systems could not use TPM on the LPC bus.

This PR aims to address that and add LPC support for embedded Linux systems that want to use LPC TPM
- Adding LPC bus state machine
- Adding LPC cycle handling for TPM Locality Read/Write (similar to LPC IO Read/Write)
- Integrating with the existing tpm_io and preserving backward comparability
- Enable future extension to support bare metal LPC bus

There are various LPC bus specifics, here is a list of the design choices made to address those specifics:
1. LPC needs free running LCLK
Generation a free running clock manually under linux is not a good idea, therefore I am using the SPIdev CLK as LCKL :) 
2. LPC needs a dedicated LFRAME pin to signal the start of every new LPC Cycle
Manual control of that pin would be possible, however I think I could use the SPIdev CS for that purpose and simplify our LPC bus logic
3. LPC needs 4-bite wide shared data, control and address bus
Because there is not native LPC support, the natural choice is a bit bang of the already slow-by-design LPC protocol interface.
* Current code is written to support Linux GPIO Sysfs layer
* The code allows adding future support of other GPIO layers. For example, adding glue code for STM32 GPIO HAL, because STM32 also does not have native LPC hardware interface.

Future identified challenges:
- LPC SYNC cycle is used for TPM wait states and does not have a defined maximum, it is defined as multiple LPC cycles, as long as the TPM responds with 0x0101 or 0x0110 in the response to LPC_SYNC.
- LPC TAR cycle changed direction of pins

My test setup is RPI4 with Nuvoton NPC650 TPM2.0 module on LPC bus.

Originally, I tried extending the WOLFTPM_SPI to support LPC, but the 4-bit wide data lenght and LPC cycle specifics just did not allow for such simple solution. Therefore, I chose to add tpm_lpc.c layer to extend tpm_io.c in wolftpm/examples/... 

Looking forward to feedback @dgarske on the design choices and code so far, before continuing forward

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ autogen.sh requires: automake and libtool: `sudo apt-get install automake libtoo
 --enable-wolfcrypt      Enable wolfCrypt hooks for RNG, Auth Sessions and Parameter encryption (default: enabled) - WOLFTPM2_NO_WOLFCRYPT
 --enable-advio          Enable Advanced IO (default: disabled) - WOLFTPM_ADV_IO
 --enable-i2c            Enable I2C TPM Support (default: disabled, requires advio) - WOLFTPM_I2C
+--enable-lpc            Enable LPC TPM Support (default: disabled, requires advio) - WOLFTPM_LPC
 --enable-checkwaitstate Enable TIS / SPI Check Wait State support (default: depends on chip) - WOLFTPM_CHECK_WAIT_STATE
 --enable-smallstack     Enable options to reduce stack usage
 --enable-tislock        Enable Linux Named Semaphore for locking access to SPI device for concurrent access between processes - WOLFTPM_TIS_LOCK
--- a/configure.ac
+++ b/configure.ac
@@ -154,6 +154,19 @@ then
 fi
 
 
+# LPC Support
+AC_ARG_ENABLE([lpc],
+    [AS_HELP_STRING([--enable-lpc],[Enable LPC TPM Support (default: disabled)])],
+    [ ENABLED_LPC=$enableval ],
+    [ ENABLED_LPC=no ]
+    )
+
+if test "x$ENABLED_LPC" = "xyes"
+then
+    AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_LPC"
+fi
+
+
 # Advanced IO
 AC_ARG_ENABLE([advio],
     [AS_HELP_STRING([--enable-advio],[Enable Advanced IO (default: disabled)])],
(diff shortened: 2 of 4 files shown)


tomoveu:
This design is also meant to support x86 bare metal LPC bus by using outb() in the place of the GPIO calls.

This is possible thanks to the support in the x86 southridge/northbridge depending on the motherboard chipset. 

There might be some differences due to AMD/Intel chipsets, but the access to the LPC bus would be writing to IO mapped memory space when running as bare metal firmware on x86.

tomoveu:
Some notes relevant to byte order, when x86 LPC bus (southbridge/northbridge) is used.

Quote from the TCG TIS Specification:
> The driver is required to handle the byte ordering of TPM command fields vs. the byte ordering of the TPM registers, LPC bridge and LPC bus. Software may do Double Word (DW) (4 bytes) or Word (2 bytes) or Byte accesses to the TPM. Standard PC platforms 3030will  have  bridges  that  translate  these  4-byte  or  2-byte  accesses  into  single-byte accesses on LPC. PC platforms will always break up the request so that they send the least significant address of that request first on the LPC bus.As  an  example,  the  TPM  data  structure  TPM_PROTOCOL_ID  is  defined  in  the TPM Library Specificationas having a value of 0x0005 for PID_OWNER, where 0x00 is the MSB  in  the  array.  To  ensure  the  TPM  receives  the  correct  command  payload,  this structure must be sent to the TPM with 0x00 as the first byte and 0x05 as the second byte.  However,  the  bridge  between  the  CPU  and  the  TPM  sends  the  LSB first  and therefore, a CPU write of 0x0005 would send the 0x05 to the TPM first, then the 0x00 which  is  not  the  correct  sequence  for  the  TCG_PROTOCOL_ID  for  PID_OWNER.  The driver could perform two 1-byte accesses to the TPM, the first write would be with data 0x00,  and  the  second  write  would  be  with  data  0x05.    Or  software  could  do  a  Word access and send 0x0500, which would result in the bridge issuing the 0x00 cycle first on the LPC bus followed by an LPC cycle of 0x05.

tomoveu:
Some more info about the SYNC cycle, because TPM wait states are important for successful communication

**Chapter 4.2.1.8 from Intel LPC specification**

SYNC  - This is used to add wait states.  It can be several clocks in length.

> 
> If a peripheral needs to assert wait states, it does so by driving ‘0101b’ (short SYNC) or ‘0110b’(long SYNC) on LAD[3:0] until it is ready.  When ready, it may choose to drive ‘0000b’ (Ready),‘1010b’ (Error), or, in the case of DMA transfers, ‘1001b’ (Ready More).On any particular cycle, if the host or peripheral chooses to insert wait states, it must choose onetype of SYNC wait value (‘0101b’ or ‘0110b’), and must not change it until it asserts one of theready SYNC values.The short SYNC is used for normal wait-states.  This is where the cycle will complete within a fewclocks
>

tomoveu:
Dropping this PR, because LPC bus seems to be legacy and available only on TPM1.2 chips. It seems there are no TPM2.0 chips with LPC bus support.

PR #136, opened 2020-12-17
Cleanup of the examples to use the key templates

dgarske:
The PKCS7 example is failing because of a key mismatch between the TPM private key and the public key used in the certificate setup using the CSR and ./certs/certreq/sh`.

ZD 11163


dgarske:
@elms , can you review this when you get a chance? You should be able to now use the PKCS7 examples correctly even on Windows TBS.

elms (approved):
Looks good! Tested with Windows TBS.

PR #137, opened 2021-01-03
Add support for LetsTrust TPM2.0 on USB stick

tomoveu:
LetsTrust USB2GO uses Cypress USB2SPI bridge that requires (Cypress) libusb.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -213,6 +213,22 @@ then
     AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_WINAPI"
 fi
 
+# LetsTrust TPM2.0 on UBS stick / Cypress USB2SPI Support
+AC_ARG_ENABLE([cyusb],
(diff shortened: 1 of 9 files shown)


tomoveu:
ping @dgarske @PaulKissinger 

This is to be tested.

tomoveu:
This is not going to be as simple as I thought it would be, so I am taking a step back to rethink this.

Here are my findings so far:

* Not a blocker: Having the TPM on USB means it has to be initialised once and detected when unplugged. Otherwise, we have to do USB detect and Library init every time we want to send a command and that's far from good. It works, but it is not good.

* Blocker: Modifying only TPM2_SendCommandAuth is not enough. Because the Cypress chip is just a USB2SPI bridge, we need to treat the TPM2.0 USB stick as SPI device. Meaning TPM2_TIS_Write/Read must be used. Perhaps we could use the Advanced IO case and keep it simple?

While the first issue could be overlooked for initial support, the second issue is vital.

dgarske:
This device is not available, so closing this PR.

PR #142, opened 2021-02-10
Add STM32 CubeMX I2C support

tomoveu:
This PR adds support for he STM32 CubeMX HAL for I2C

To enable CubeMX I2C HAL

1. wolfTPM must be build using `--enable-i2c`
2. OpenSTM or STM32IDE should have the following define in their project settings: `WOLFSSL_STM32_CUBEMX`. 

**NB: To be tested and confirmed working**

ZD 11681 and https://www.wolfssl.com/forums/topic1680-tpm2iocbstcubemxi2c.html

### DIFF
--- a/IDE/OPENSTM32/Inc/user_settings.h
+++ b/IDE/OPENSTM32/Inc/user_settings.h
@@ -22,8 +22,8 @@ extern "C" {
 #undef  WOLFSSL_STM32F4
 #define WOLFSSL_STM32F4
 
-#undef  WOLFSSL_STM32_CUBEMX_SPI
-#define WOLFSSL_STM32_CUBEMX_SPI
+#undef  WOLFSSL_STM32_CUBEMX
+#define WOLFSSL_STM32_CUBEMX
 
 /* Optionally Disable Hardware Hashing Support */
 //#define NO_STM32_HASH
--- a/IDE/OPENSTM32/Inc/user_settings.h
+++ b/IDE/OPENSTM32/Inc/user_settings.h
@@ -22,8 +22,8 @@ extern "C" {
 #undef  WOLFSSL_STM32F4
 #define WOLFSSL_STM32F4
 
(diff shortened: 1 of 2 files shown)


IDE/OPENSTM32/Inc/user_settings.h
```diff
@@ -22,8 +22,8 @@ extern "C" {
 #undef  WOLFSSL_STM32F4
 #define WOLFSSL_STM32F4
 
-#undef  WOLFSSL_STM32_CUBEMX
-#define WOLFSSL_STM32_CUBEMX
+#undef  WOLFTPM_STM32_CUBEMX_SPI
```
dgarske, line 25: You need to keep `WOLFSSL_STM32_CUBEMX`. That is a define we already use in wolfSSL to indicate using STM32 with the CubeMX HAL. We already have defines in wolfTPM to indicate I2C using `WOLFTPM_I2C`. Please revert this and default to SPI unless `WOLFTPM_I2C` is defined for Cube.

dgarske: Add the code flow will be:

```
#ifdef WOLFSSL_STM32_CUBEMX
#ifdef WOLFTPM_I2C
/* Use STM32CubeHAL with I2C */
#else
/* Use STM32CubeHAL with SPI */
#endif /* WOLFTPM_I2C */
#endif /* WOLFSSL_STM32_CUBEMX */
```

dgarske (changes requested):

IDE/OPENSTM32/Inc/user_settings.h
tomoveu: Thansk, David. It is now fixed.

I did not know WOLFSSL_STM32_CUBEMX is used across wolfSSL and wolfTPM.

I also got lost in the long file and thought WOLFTPM_I2C is nested with the `__linux__` ifdefs.

dgarske (approved):

PR #143, opened 2021-02-11
Add symmetric key option to examples

tomoveu:
This PR is part of ZD11566

I decided to split the changes in two PRs
- examples/keygen
- examples/nvram

PR #145, opened 2021-02-12
Add examples for NVRAM usage

tomoveu:
This PR is the second part of ZD11566 and adds two new examples:
* examples/nvram/store
* examples/nvram/read

Examples work without parameter encryption.

With parameter encryption, NV DefineSpace breaks execution (working on this issue)

1. Generate a TPM key (can be any key, RSA, ECC or AES symmetric)
```
pi@raspberrypi:~/wolfTPM $ sudo ./examples/keygen/keygen -rsa
TPM2.0 Key generation example
	Key Blob: keyblob.bin
	Algorithm: RSA
	Template: AIK
	Use Parameter Encryption: NULL
wolfSSL Entering wolfCrypt_Init
TPM2_ReadPublic Handle 0x81000200: pub 282, name 34, qualifiedName 34
Loading SRK: Storage 0x81000200 (282 bytes)
RSA AIK template
Creating new RSA key...
Created new key (pub 280, priv 222 bytes)
Wrote 840 bytes to keyblob.bin
wolfSSL Entering wolfCrypt_Cleanup
```

2. Store the key in the TPM's NVRAM and set password (the string gNvAuth from tpm_test.h)

```
pi@raspberrypi:~/wolfTPM $ sudo examples/nvram/store
wolfSSL Entering wolfCrypt_Init
Reading 840 bytes from keyblob.bin
TPM2_NV_DefineSpace: Auth 0x40000001, Idx 0x1800203, Attribs 0x33947654, Size 1024
Storing key at TPM NV index 0x1800203 with password protection
Public part = 616 bytes
TPM2_NV_Write: Auth 0x1800203, Idx 0x1800203, Offset 0, Size 616
Private part = 222 bytes
TPM2_NV_Write: Auth 0x1800203, Idx 0x1800203, Offset 616, Size 222
NV write succeeded
wolfSSL Entering wolfCrypt_Cleanup
```

3. Read out the key from NVRAM using the password (requires to specify size of the private and public part)

```
pi@raspberrypi:~/wolfTPM $ sudo examples/nvram/read 616 222
wolfSSL Entering wolfCrypt_Init
Trying to read 616 bytes of public key part from NV
TPM2_NV_Read: Auth 0x1800203, Idx 0x1800203, Offset 0, Size 616
Trying to read 222 bytes of private key part from NV
TPM2_NV_Read: Auth 0x1800203, Idx 0x1800203, Offset 616, Size 222
TPM2_NV_UndefineSpace: Auth 0x40000001, Idx 0x1800203
Extraction of key from NV RAm at index 0x1800203
wolfSSL Entering wolfCrypt_Cleanup
```

First part of Zd11566 is #143

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -44,6 +44,14 @@ More information about how to test and use PCR attestation can be found in the i
 
 ## Parameter Encryption
 
+### Key generation with encrypted authorization
+
+Detailed information can be found in this file under section "Key generation"
+
+### Secure vault for keys with encrypted NVRAM authorization
+
+Detailed information can be found in this file under section "Storing keys into the TPM's NVRAM"
+
 ### TPM2.0 Quote with encrypted user data
 
 Example for demonstrating how to use parameter encryption to protect the user data between the Host and the TPM.
@@ -260,3 +268,29 @@ Loaded key to 0x80000001
 ```
 
 The `keyload` tool takes only one argument, the filename of the stored key. Because the information what is key scheme (RSA or ECC) is contained within the key blob.
+
+## Storing keys into the TPM's NVRAM
+
+These examples demonstrates how to use the TPM as secure vault for keys. There are two programs, one to store a TPM key into the TPM's NVRAM and another to extract the key from the TPM's NVRAM. Both examples can use parameter encryption to protect from MITM attacks. The Non-volatile memory location is protected with a password authorization that is passed in encrypted form, when "-aes" or "-xor" is given on the commmand line.
+
+Before running the examples, make sure there is a keyblob.bin generated using the keygen tool. The key can be of any type, RSA, ECC or symmetric. The example will store the private and public part. In case of a symmetric key the public part is meta data from the TPM. How to generate a key you can see above, in the description of the keygen example.
+
+Typical output for storing and then reading an RSA key using parameter encryption:
+
+```
+$ ./examples/nvram/store -aes
+TPM2_StartAuthSession: sessionHandle 0x2000000
+Reading 840 bytes from keyblob.bin
+Storing key at TPM NV index 0x1800203 with password protection
+Public part = 616 bytes
+Private part = 222 bytes
+NV write succeeded
+
+$ ./examples/nvram/read 616 222 -aes
+TPM2_StartAuthSession: sessionHandle 0x2000000
+Trying to read 616 bytes of public key part from NV
+Trying to read 222 bytes of private key part from NV
+Extraction of key from NVRAM at index 0x1800203 succeeded
+```
+
+The read example takes as first argument the size of the public part and as second argument the private part. This information is given from the store example. The "-aes" swiches triggers the use of parameter encryption.
--- a/examples/README.md
+++ b/examples/README.md
@@ -281,7 +281,7 @@ Typical output for storing and then reading an RSA key using parameter encryptio
 $ ./examples/nvram/store -aes
 TPM2_StartAuthSession: sessionHandle 0x2000000
 Reading 840 bytes from keyblob.bin
-Storing key at TPM NV index 0x1800203 with password protection
+Storing key at TPM NV index 0x1800202 with password protection
 Public part = 616 bytes
 Private part = 222 bytes
 NV write succeeded
@@ -290,7 +290,7 @@ $ ./examples/nvram/read 616 222 -aes
 TPM2_StartAuthSession: sessionHandle 0x2000000
 Trying to read 616 bytes of public key part from NV
 Trying to read 222 bytes of private key part from NV
-Extraction of key from NVRAM at index 0x1800203 succeeded
+Extraction of key from NVRAM at index 0x1800202 succeeded
 ```
 
 The read example takes as first argument the size of the public part and as second argument the private part. This information is given from the store example. The "-aes" swiches triggers the use of parameter encryption.
--- a/examples/nvram/read.c
+++ b/examples/nvram/read.c
@@ -120,28 +120,28 @@ int TPM2_NVRAM_Read_Example(void* userCtx, int argc, char *argv[])
 
     /* Prepare auth for NV Index */
     XMEMSET(&nv, 0, sizeof(nv));
-    nv.handle.hndl = TPM2_DEMO_NVRAM_STORE_PRIV_INDEX;
+    nv.handle.hndl = TPM2_DEMO_NVRAM_STORE_INDEX;
     nv.handle.auth.size = auth.size;
     XMEMCPY(nv.handle.auth.buffer, auth.buffer, auth.size);
 
     readSize = keyBlob.pub.size;
     printf("Trying to read %d bytes of public key part from NV\n", readSize);
-    rc = wolfTPM2_NVReadAuth(&dev, &nv, TPM2_DEMO_NVRAM_STORE_PRIV_INDEX,
+    rc = wolfTPM2_NVReadAuth(&dev, &nv, TPM2_DEMO_NVRAM_STORE_INDEX,
         (byte*)&keyBlob.pub.publicArea, &readSize, 0);
     if (rc != 0) goto exit;
 
     readSize = keyBlob.priv.size;
     printf("Trying to read %d bytes of private key part from NV\n", readSize);
-    rc = wolfTPM2_NVReadAuth(&dev, &nv, TPM2_DEMO_NVRAM_STORE_PRIV_INDEX,
+    rc = wolfTPM2_NVReadAuth(&dev, &nv, TPM2_DEMO_NVRAM_STORE_INDEX,
         (byte*)&keyBlob.priv.buffer, &readSize, keyBlob.pub.size);
     if (rc != 0) goto exit;
 
     parent.hndl = TPM_RH_OWNER;
-    rc = wolfTPM2_NVDeleteAuth(&dev, &parent, TPM2_DEMO_NVRAM_STORE_PRIV_INDEX);
+    rc = wolfTPM2_NVDeleteAuth(&dev, &parent, TPM2_DEMO_NVRAM_STORE_INDEX);
     if (rc != 0) goto exit;
 
     printf("Extraction of key from NVRAM at index 0x%x succeeded\n" ,
-        TPM2_DEMO_NVRAM_STORE_PRIV_INDEX);
+        TPM2_DEMO_NVRAM_STORE_INDEX);
 
 exit:
 
--- a/examples/nvram/read.c
+++ b/examples/nvram/read.c
@@ -1,0 +1,178 @@
+/* read.c
+ *
+ * Copyright (C) 2006-2021 wolfSSL Inc.
+ *
+ * This file is part of wolfTPM.
+ *
+ * wolfTPM is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfTPM is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/* Tool and example for extracting a TPM key from the TPM's NVRAM
+ *
+ * NB: This example uses Parameter Encryption to protect
+ *     the Password Authorization of the TPM NVRAM Index
+ *
+ **/
+
+#include <wolftpm/tpm2_wrap.h>
+
+#include <examples/nvram/store.h>
+#include <examples/tpm_io.h>
+#include <examples/tpm_test.h>
+#include <examples/tpm_test_keys.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#ifndef WOLFTPM2_NO_WRAPPER
+
+#define ARGC_PUB_SIZE  1
+#define ARGC_PRIV_SIZE 2
+#define ARGC_PARAM_ENC 3
+
+/******************************************************************************/
+/* --- BEGIN TPM Keygen Example -- */
+/******************************************************************************/
+static void usage(void)
+{
+    printf("Expected usage:\n");
+    printf("./examples/nvram/read priv pub [-aes/-xor]\n");
+    printf("* priv: size in bytes of the private part of the key in NV\n");
+    printf("* pub: size in bytes of the public part of the key in NV\n");
+    printf("* -aes/xor: Use Parameter Encryption\n");
+}
+
+int TPM2_NVRAM_Read_Example(void* userCtx, int argc, char *argv[])
+{
+    int rc;
+    WOLFTPM2_DEV dev;
+    WOLFTPM2_KEYBLOB keyBlob;
+    WOLFTPM2_SESSION tpmSession;
+    WOLFTPM2_HANDLE parent;
+    WOLFTPM2_NV nv;
+    TPM2B_AUTH auth;
+    word32 readSize;
+    int paramEncAlg = TPM_ALG_NULL;
+
+    if (argc > 2) {
+        XMEMSET(&keyBlob, 0, sizeof(keyBlob));
+        keyBlob.pub.size = atoi(argv[ARGC_PUB_SIZE]);
+        keyBlob.priv.size = atoi(argv[ARGC_PRIV_SIZE]);
+
+        if (argc == 4) {
+            if (XSTRNCMP(argv[ARGC_PARAM_ENC], "-aes", 4) == 0) {
+                paramEncAlg = TPM_ALG_CFB;
+            }
+            if (XSTRNCMP(argv[ARGC_PARAM_ENC], "-xor", 4) == 0) {
+                paramEncAlg = TPM_ALG_XOR;
+            }
+            argc--;
+        }
+    }
+    else {
+        usage();
+        return 0;
+    }
+
+    if (keyBlob.priv.size == 0 || keyBlob.pub.size == 0) {
+        printf("Specify size of private and public part of the key\n");
+        usage();
+        return 0;
+    }
+
+    XMEMSET(&tpmSession, 0, sizeof(tpmSession));
+    XMEMSET(&parent, 0, sizeof(parent));
+    XMEMSET(&auth, 0, sizeof(auth));
+
+    rc = wolfTPM2_Init(&dev, TPM2_IoCb, userCtx);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("\nwolfTPM2_Init failed\n");
+        goto exit;
+    }
+
+    if (paramEncAlg != TPM_ALG_NULL) {
+        /* Start TPM session for parameter encryption */
+        rc = wolfTPM2_StartSession(&dev, &tpmSession, NULL, NULL,
+                TPM_SE_HMAC, TPM_ALG_CFB);
+        if (rc != 0) goto exit;
+        printf("TPM2_StartAuthSession: sessionHandle 0x%x\n",
+            (word32)tpmSession.handle.hndl);
+        /* Set TPM session attributes for parameter encryption */
+        rc = wolfTPM2_SetAuthSession(&dev, 1, &tpmSession,
+            (TPMA_SESSION_decrypt | TPMA_SESSION_encrypt | TPMA_SESSION_continueSession));
+        if (rc != 0) goto exit;
+    }
+
+    auth.size = sizeof(gNvAuth)-1;
+    XMEMCPY(auth.buffer, gNvAuth, auth.size);
+
+    /* Prepare auth for NV Index */
+    XMEMSET(&nv, 0, sizeof(nv));
+    nv.handle.hndl = TPM2_DEMO_NVRAM_STORE_PRIV_INDEX;
+    nv.handle.auth.size = auth.size;
+    XMEMCPY(nv.handle.auth.buffer, auth.buffer, auth.size);
+
+    readSize = keyBlob.pub.size;
+    printf("Trying to read %d bytes of public key part from NV\n", readSize);
+    rc = wolfTPM2_NVReadAuth(&dev, &nv, TPM2_DEMO_NVRAM_STORE_PRIV_INDEX,
+        (byte*)&keyBlob.pub.publicArea, &readSize, 0);
+    if (rc != 0) goto exit;
+
+    readSize = keyBlob.priv.size;
+    printf("Trying to read %d bytes of private key part from NV\n", readSize);
+    rc = wolfTPM2_NVReadAuth(&dev, &nv, TPM2_DEMO_NVRAM_STORE_PRIV_INDEX,
+        (byte*)&keyBlob.priv.buffer, &readSize, keyBlob.pub.size);
+    if (rc != 0) goto exit;
+
+    parent.hndl = TPM_RH_OWNER;
+    rc = wolfTPM2_NVDeleteAuth(&dev, &parent, TPM2_DEMO_NVRAM_STORE_PRIV_INDEX);
+    if (rc != 0) goto exit;
+
+    printf("Extraction of key from NVRAM at index 0x%x succeeded\n" ,
+        TPM2_DEMO_NVRAM_STORE_PRIV_INDEX);
+
+exit:
+
+    if (rc != 0) {
+        printf("\nFailure 0x%x: %s\n\n", rc, wolfTPM2_GetRCString(rc));
+    }
+
+    wolfTPM2_UnloadHandle(&dev, &tpmSession.handle);
+    wolfTPM2_Cleanup(&dev);
+
+    return rc;
+}
+
+/******************************************************************************/
+/* --- END TPM NVRAM Store Example -- */
+/******************************************************************************/
+#endif /* !WOLFTPM2_NO_WRAPPER */
+
+#ifndef NO_MAIN_DRIVER
+int main(int argc, char *argv[])
+{
+    int rc = NOT_COMPILED_IN;
+
+#ifndef WOLFTPM2_NO_WRAPPER
+    rc = TPM2_NVRAM_Read_Example(NULL, argc, argv);
+#else
+    printf("NVRAM code not compiled in\n");
+    (void)argc;
+    (void)argv;
+#endif
+
+    return rc;
+}
+#endif
--- a/examples/nvram/store.c
+++ b/examples/nvram/store.c
@@ -1,0 +1,170 @@
+/* store.c
+ *
+ * Copyright (C) 2006-2021 wolfSSL Inc.
+ *
+ * This file is part of wolfTPM.
+ *
+ * wolfTPM is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfTPM is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/* Tool and example for storing a TPM key into the TPM's NVRAM
+ *
+ * NB: This example uses Parameter Encryption to protect the password of the
+ *     TPM NVRAM Index, where the private and public parts of a TPM key is stored
+ *
+ **/
+
+#include <wolftpm/tpm2_wrap.h>
+
+#include <examples/nvram/store.h>
+#include <examples/tpm_io.h>
+#include <examples/tpm_test.h>
+#include <examples/tpm_test_keys.h>
+
+#include <stdio.h>
+
+#ifndef WOLFTPM2_NO_WRAPPER
+
+/******************************************************************************/
+/* --- BEGIN TPM Keygen Example -- */
+/******************************************************************************/
+static void usage(void)
+{
+    printf("Expected usage:\n");
+    printf("./examples/nvram/store [filename] [-aes/-xor]\n");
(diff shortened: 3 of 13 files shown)


tomoveu:
I found an issue with some TPM2 Commands who get their packet data overwritten by the HMAC field calculated during parameter encryption. This was happening because no space was reserved for the HMAC field (hmac.Size/auth.Size was 0 during AppendAuth), so packet->pos was pointing to an incorrect place in the packet buffer later during ParamEnc functions.

After figuring out the tricky situation, I found a nice place in SetAuthSession for the fix. Because this issue only impacts Parameter Encryption sessions at the moment. In future, it could have impact on HMAC sessions without parameter encryption.

tomoveu:
AUTH HMAC issue comes from the Handle Names used during authorization.  The reason is that we are extracting the names incorrectly.

To complete the fix and extract(compute) the correct names we might need to start managing information about TPM objects that the stack creates.

I pushed my draft code for the fix with comment explaining the TPM object info we miss.

tomoveu:
I just noticed that some of my new code was lost after rebase. I pushed it back as a new commit.

And this made me realize that **we might also need the original TPM2_GetName, but as TPM2_GetSessionName for calcCpHash().**

tomoveu:
I just noticed that some of my new code was lost after rebase. I pushed it back as a new commit.

src/tpm2.c
```diff
@@ -187,9 +188,16 @@ static int TPM2_CommandProcess(TPM2_CTX* ctx, TPM2_Packet* packet,
             }
 
         #ifndef WOLFTPM2_NO_WOLFCRYPT
-            rc =  TPM2_GetName(ctx, info->inHandleCnt, 0, &name1);
-            rc |= TPM2_GetName(ctx, info->inHandleCnt, 1, &name2);
-            rc |= TPM2_GetName(ctx, info->inHandleCnt, 2, &name3);
+            UINT32 handleValue;
```
dgarske, line 191: Please move variable declaration to top of function or put within a brace.

src/tpm2_wrap.c
```diff
@@ -2560,9 +2582,42 @@ int wolfTPM2_NVCreateAuth(WOLFTPM2_DEV* dev, WOLFTPM2_HANDLE* parent,
         return rc;
     }
 
+    /* Compute NV Index name in case of parameter encryption */
+#ifndef WOLFTPM2_NO_WOLFCRYPT
+    rc = TPM2_GetHashType(in.publicInfo.nvPublic.nameAlg);
```
dgarske, line 2587: Please move this logic into its own function that takes in an nvPublic struct computes the hash digest (name for NV).

src/tpm2.c
```diff
@@ -5686,6 +5706,20 @@ void TPM2_PrintBin(const byte* buffer, word32 length)
...
     }
 }
+
+void TPM2_PrintAuth(const TPMS_AUTH_COMMAND* authCmd)
+{
+    if (authCmd == NULL)
+        return;
+
+    printf("authCmd:\n");
+    printf("sessionHandle=0x%7X\n", authCmd->sessionHandle);
```
dgarske, line 5716: Why not `0x%08X` or `0x%8X`?

dgarske (changes requested):

tomoveu:
@dgarske  thanks, I pushed new changes

* Hashing over the whole infoPublic instead of just nvPublic
* added generic function for Hashing TPM2B_NV_PUBLIC

Strangely, the issue persists, my name computation differs from what the tpm expects.

tomoveu:
@dgarske  ready for review :)

Once done, I will rebase the changes to 2-3 commits for better history readability. 

And for GitHub issue traceability:
- Fixed TPM2_GetName computation
- Fixed HMAC auth issue due nvPublic changing with NV_WRITTEN flag between NVWrite commands, thus Name chagned and HMAC Auth failed 
- Fixed unneeded Auth slot clearing in wolfTPM2_NVDeleteAuth wrapper and added new helper

**Output with parameter encryption**

```
dimitartomov@Dimitars-MacBook-Pro wolfTPM % ./examples/nvram/store -aes
...
TPM2_StartAuthSession: handle 0x2000000, algorithm AES
TPM2_StartAuthSession: sessionHandle 0x2000000
Reading 840 bytes from keyblob.bin
TPM2_NV_DefineSpace: Auth 0x40000001, Idx 0x1800203, Attribs 0x33947654, Size 1024
Storing key at TPM NV index 0x1800203 with password protection
Public part = 616 bytes
TPM2_NV_Write: Auth 0x1800203, Idx 0x1800203, Offset 0, Size 616
Private part = 222 bytes
TPM2_NV_Write: Auth 0x1800203, Idx 0x1800203, Offset 616, Size 222
**NV write succeeded**
TPM2_FlushContext: Closed handle 0x2000000

dimitartomov@Dimitars-MacBook-Pro wolfTPM % ./examples/nvram/read 616 222 -aes
...
TPM2_StartAuthSession: handle 0x2000000, algorithm AES
TPM2_StartAuthSession: sessionHandle 0x2000000
Trying to read 616 bytes of public key part from NV
TPM2_NV_Read: Auth 0x1800203, Idx 0x1800203, Offset 0, Size 616
Trying to read 222 bytes of private key part from NV
TPM2_NV_Read: Auth 0x1800203, Idx 0x1800203, Offset 616, Size 222
TPM2_NV_UndefineSpace: Auth 0x40000001, Idx 0x1800203
Extraction of key from NVRAM at index 0x1800203 succeeded
TPM2_FlushContext: Closed handle 0x2000000
dimitartomov@Dimitars-MacBook-Pro wolfTPM % 

```

Output without parameter encryption
```
dimitartomov@Dimitars-MacBook-Pro wolfTPM % ./examples/nvram/store     
...
Reading 840 bytes from keyblob.bin
TPM2_NV_DefineSpace: Auth 0x40000001, Idx 0x1800203, Attribs 0x33947654, Size 1024
Storing key at TPM NV index 0x1800203 with password protection
Public part = 616 bytes
TPM2_NV_Write: Auth 0x1800203, Idx 0x1800203, Offset 0, Size 616
Private part = 222 bytes
TPM2_NV_Write: Auth 0x1800203, Idx 0x1800203, Offset 616, Size 222
NV write succeeded

dimitartomov@Dimitars-MacBook-Pro wolfTPM % ./examples/nvram/read 616 222     
...
Trying to read 616 bytes of public key part from NV
TPM2_NV_Read: Auth 0x1800203, Idx 0x1800203, Offset 0, Size 616
Trying to read 222 bytes of private key part from NV
TPM2_NV_Read: Auth 0x1800203, Idx 0x1800203, Offset 616, Size 222
TPM2_NV_UndefineSpace: Auth 0x40000001, Idx 0x1800203
Extraction of key from NVRAM at index 0x1800203 succeeded
dimitartomov@Dimitars-MacBook-Pro wolfTPM % 
```

src/tpm2.c
dgarske: This is still an issue. Variables must be declared at top of function or brace.

src/tpm2.c
```diff
@@ -5278,10 +5287,28 @@ int TPM2_GetName(TPM2_CTX* ctx, int handleCnt, int idx, TPM2B_NAME* name)
         return TPM_RC_SUCCESS;
 
     session = &ctx->session[idx];
+
+#if 0
```
dgarske, line 5291: Please remove dead code.

src/tpm2.c
```diff
@@ -5648,6 +5675,67 @@ UINT16 TPM2_GetVendorID(void)
...
+        nameAlgSize = sizeof(nvPublic->nameAlg);
+        XMEMCPY(buffer, (byte*)&nameAlgValue, nameAlgSize);
+        XMEMCPY(&buffer[2], digest.buffer, hashSize);
+        /* account for nameAlg concatenation */
+        *size = hashSize + nameAlgSize;
+    }
+
+    wc_HashFree(&hash, hashType);
+
+    return TPM_RC_SUCCESS;
```
dgarske, line 5733: Return code is not being checked or returned... Where is `rc` checked?

src/tpm2_wrap.c
```diff
@@ -26,6 +26,11 @@
 
 /* For some struct to buffer conversions */
 #include <wolftpm/tpm2_packet.h>
+#ifndef WOLFTPM2_NO_WOLFCRYPT
+/* Required in wolfTPM2_CreateAuth() for name computation of NV handles */
+#include <wolfssl/wolfcrypt/hash.h>
+#endif
```
dgarske, line 32: I believe this section is no longer required because its in tpm2.c now.

src/tpm2_wrap.c
```diff
@@ -369,6 +374,20 @@ int wolfTPM2_GetCapabilities(WOLFTPM2_DEV* dev, WOLFTPM2_CAPS* cap)
     return wolfTPM2_GetCapabilities_NoDev(cap);
 }
 
+int wolfTPM2_UnsetAuth(WOLFTPM2_DEV* dev, int index)
+{
+    TPM2_AUTH_SESSION* session;
+
+    if (dev == NULL || index >= MAX_SESSION_NUM) {
```
dgarske, line 381: What about a negative index? I see other cases where this could happen. Please correct.

src/tpm2_wrap.c
```diff
@@ -2524,6 +2570,7 @@ int wolfTPM2_NVCreateAuth(WOLFTPM2_DEV* dev, WOLFTPM2_HANDLE* parent,
 {
     int rc;
     NV_DefineSpace_In in;
+    TPM2B_NAME name;
```
dgarske, line 2573: You can use `nv->handle.name` directly. A separate TPM2B_NAME is not required.

src/tpm2_wrap.c
```diff
@@ -2605,6 +2661,27 @@ int wolfTPM2_NVWriteAuth(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv,
         wolfTPM2_SetAuthHandle(dev, 0, &nv->handle);
     }
 
+    XMEMSET((byte*)&inPublic, 0, sizeof(inPublic));
```
dgarske, line 2664: Why not use `wolfTPM2_NVReadPublic`?

src/tpm2_wrap.c
```diff
@@ -2665,6 +2745,27 @@ int wolfTPM2_NVReadAuth(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv,
...
```
dgarske, line 2748: Why not use `wolfTPM2_NVReadPublic`?

src/tpm2.c
```diff
@@ -5648,6 +5675,67 @@ UINT16 TPM2_GetVendorID(void)
...
+/* Stores nameAlg + the digest of nvPublic in buffer, total size in size */
+int TPM2_HashNvPublic(TPMS_NV_PUBLIC* nvPublic, byte* buffer, UINT16* size)
+{
+    int rc;
+#ifndef WOLFTPM2_NO_WOLFCRYPT
+    int hashSize, nameAlgValue, nameAlgSize;
+    wc_HashAlg hash;
+    enum wc_HashType hashType;
+    byte appending[sizeof(TPMS_NV_PUBLIC)];
+    TPM2B_DIGEST digest;
```
dgarske, line 5687: This is not guaranteed to be max digest + 2. Use TPM2B_NAME with .name. That is... `TPMT_HA`.

src/tpm2.c
```diff
@@ -5648,6 +5675,67 @@ UINT16 TPM2_GetVendorID(void)
     return vid;
 }
 
+/* Stores nameAlg + the digest of nvPublic in buffer, total size in size */
+int TPM2_HashNvPublic(TPMS_NV_PUBLIC* nvPublic, byte* buffer, UINT16* size)
+{
+    int rc;
+#ifndef WOLFTPM2_NO_WOLFCRYPT
+    int hashSize, nameAlgValue, nameAlgSize;
```
dgarske, line 5683: Please use `UINT16` for `nameAlgValue`.

dgarske (changes requested):
Please correct feedback. Please add section in README.md with example output and usage. Please rebase and squash merge. The testing all passed and otherwise looks good.

src/tpm2.c
tomoveu: Indeed. Fixed

src/tpm2_wrap.c
tomoveu: Already move into TPM2_HashNvPublic

src/tpm2.c
tomoveu: session Handles are 7 digits, other TPM objects are loaded into 8 digit handles

```
wolftpm/tpm2.h:    TPM_HT_PCR              = 0x00,
wolftpm/tpm2.h:    TPM_HT_NV_INDEX         = 0x01,
wolftpm/tpm2.h:    TPM_HT_HMAC_SESSION     = 0x02,
wolftpm/tpm2.h:    TPM_HT_LOADED_SESSION   = 0x02,
wolftpm/tpm2.h:    TPM_HT_POLICY_SESSION   = 0x03,
wolftpm/tpm2.h:    TPM_HT_ACTIVE_SESSION   = 0x03,

notice the leading zero, of course we could print that too to show it is a 7 digit session handle

wolftpm/tpm2.h:    TPM_HT_PERMANENT        = 0x40,
wolftpm/tpm2.h:    TPM_HT_TRANSIENT        = 0x80,
wolftpm/tpm2.h:    TPM_HT_PERSISTENT       = 0x81,

```

I will fix it.

src/tpm2.c
tomoveu: Removed. Thanks!

src/tpm2_wrap.c
tomoveu: Indeed. Good catch

src/tpm2.c
tomoveu: this is changed to return rc; now

and in the case when wolfcrypt is not enabled(param encryption is not enabled) I return TPM_RC_SUCCESS, because the computation will not be used without param encryption.

tomoveu: the RC result is returned from the wolfTPM2_wrappers to the caller

this is why the RC is not checked, I will add comment so it is clear

src/tpm2_wrap.c
tomoveu: I took this rule from SetAuth, so I will apply a fix there as well.

src/tpm2_wrap.c
tomoveu: Indeed. I was already doing it in WriteAuth and ReadAuth. Good catch. Thank you

src/tpm2_wrap.c
tomoveu: Wow. I've forgot we have wolfTPM2 wrapper for that. Changed the code to use it - Awesome :)

src/tpm2_wrap.c
tomoveu: Likewise, awesome to use our own wrappers and make the code even better and more neat.

src/tpm2.c
tomoveu: done. thanks

src/tpm2.c
tomoveu: Very good point - indeed TPM2B_DIGEST buffer can not fit digest+2, but only digest size.

I will use TPM2B_DATA , because this is the type with buffer of size TPMT_HA
```

typedef struct TPM2B_DATA {
    UINT16 size;
    BYTE buffer[sizeof(TPMT_HA)];
} TPM2B_DATA;

```

tomoveu:
Awesome review @dgarske  Thank you 🥇 

All feedback is applied. Commits are squashed. Readme updated.

Please re-review and mark resolved issues :)

examples/tpm_test.h
```diff
@@ -43,6 +42,8 @@
 
 #define TPM2_DEMO_NV_TEST_INDEX         0x01800200
 #define TPM2_DEMO_NV_TEST_AUTH_INDEX    0x01800201
+#define TPM2_DEMO_NVRAM_STORE_PUB_INDEX 0x01800202
```
dgarske, line 45: The pub index is not used... Please remove and rename `TPM2_DEMO_NVRAM_STORE_PRIV_INDEX`, since it contains both private and public portions.

src/tpm2.c
```diff
@@ -5648,6 +5665,69 @@ UINT16 TPM2_GetVendorID(void)
...
+int TPM2_HashNvPublic(TPMS_NV_PUBLIC* nvPublic, byte* buffer, UINT16* size)
+{
+    int rc;
+#ifndef WOLFTPM2_NO_WOLFCRYPT
+    int hashSize, nameAlgSize;
+    UINT16 nameAlgValue;
+    wc_HashAlg hash;
+    enum wc_HashType hashType;
+    byte appending[sizeof(TPMS_NV_PUBLIC)];
+    TPM2B_DATA digest;
```
dgarske, line 5678: This is an oversized buffer and will use more stack than required. Either use the correctly sized buffer `TPMU_NAME`, which is max digest + 2 bytes (handle) or use the incoming 'buffer` directly and remove this.

src/tpm2.c
```diff
@@ -5648,6 +5665,69 @@ UINT16 TPM2_GetVendorID(void)
...
+    }
+    hashSize = rc;
+
+    rc = wc_HashInit(&hash, hashType);
+    if (rc == 0) {
+        rc = wc_HashUpdate(&hash, hashType, packet.buf, packet.pos);
+    }
+
+    if (rc == 0) {
+        rc = wc_HashFinal(&hash, hashType, digest.buffer);
```
dgarske, line 5710: Put the hash digest directly into `&buffer[2]`.

src/tpm2.c
```diff
@@ -5648,6 +5665,69 @@ UINT16 TPM2_GetVendorID(void)
...
+    }
+
+    if (rc == 0) {
+        nameAlgValue = TPM2_Packet_SwapU16(nvPublic->nameAlg);
+        nameAlgSize = sizeof(nvPublic->nameAlg);
+        XMEMCPY(buffer, (byte*)&nameAlgValue, nameAlgSize);
+        XMEMCPY(&buffer[2], digest.buffer, hashSize);
```
dgarske, line 5717: Then remove this.

src/tpm2_wrap.c
```diff
@@ -2665,6 +2744,30 @@ int wolfTPM2_NVReadAuth(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv,
...
+                            &nv->handle.name.size);
+    if (rc != TPM_RC_SUCCESS) {
+        return rc;
+    }
+
+    /* Necessary, because NVWrite has two handles, second is NV Index */
+    rc = wolfTPM2_SetNameHandle(dev, 0, &nv->handle);
+    rc |= wolfTPM2_SetNameHandle(dev, 1, &nv->handle);
+    if (rc != TPM_RC_SUCCESS) {
+        printf("Storing NV Index Name failed\n");
```
dgarske, line 2767: Please wrap this printf in `DEBUG_WOLFTPM`. The `rc` return code will not be valid with your above or'ing of the return code. Consider just returning `TPM_RC_FAILURE`.

src/tpm2_wrap.c
```diff
@@ -2605,6 +2658,30 @@ int wolfTPM2_NVWriteAuth(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv,
...
```
dgarske, line 2681: Please wrap this printf in `DEBUG_WOLFTPM`. The `rc` return code will not be valid with your above or'ing of the return code. Consider just returning `TPM_RC_FAILURE`.

src/tpm2_wrap.c
```diff
@@ -415,6 +430,24 @@ int wolfTPM2_SetAuthHandle(WOLFTPM2_DEV* dev, int index,
     return wolfTPM2_SetAuth(dev, index, TPM_RS_PW, auth, 0, name);
 }
 
+int wolfTPM2_SetNameHandle(WOLFTPM2_DEV* dev, int index,
```
dgarske, line 433: Please rename this to `wolfTPM2_SetAuthHandleName`. It is just setting the name for an existing auth handle.

src/tpm2_wrap.c
```diff
@@ -2560,9 +2602,19 @@ int wolfTPM2_NVCreateAuth(WOLFTPM2_DEV* dev, WOLFTPM2_HANDLE* parent,
...
+    /* Compute NV Index name in case of parameter encryption */
+    rc = TPM2_HashNvPublic(&in.publicInfo.nvPublic,
+                            (byte*)&nv->handle.name.name,
+                            &nv->handle.name.size);
+    if (rc != TPM_RC_SUCCESS) {
+        return rc;
+    }
+
     /* return new NV handle */
+    XMEMSET(nv, 0, sizeof(*nv));
```
dgarske, line 2614: This is broken. It will memset and clear the `nv->handle.name` done above. The memset needs moved above `TPM2_HashNvPublic`.

dgarske (changes requested):

src/tpm2.c
tomoveu: I was not sure if this is okay. I will do it like this using &buffer[2].

src/tpm2.c
tomoveu: I removed the XMEMCPY regarding digest.buffer, the concatenation with nameAlg remains as required.

examples/tpm_test.h
tomoveu: fixed. Thanks

src/tpm2_wrap.c
tomoveu: OK. I think this name fits better, however the justification for me is different - We use the AUTH structure to pass the NAME of a TPM Handle down to our parameter encryption implementation.  Typically, the NAME of a TPM Handle is not part of a AUTH structure. In our case, it is a good solution to fetch necessary info for our parameter encryption.

src/tpm2.c
tomoveu: I went with using directly buffer[2] , thanks.

src/tpm2_wrap.c
tomoveu: good point. Thank you

src/tpm2_wrap.c
tomoveu: Done. Thanks

src/tpm2_wrap.c
tomoveu: Indeed, leftover after making the Hash Name computation a generic function. 

This actually had no impact over the examples, because we are making the Name computation here just in case some later command that will be using parameter encryption needs it.

Good catch!

tomoveu:
@dgarske  all feedback taken in, good points

examples/README.md
```diff
@@ -260,3 +268,29 @@ Loaded key to 0x80000001
...
+```
+$ ./examples/nvram/store -aes
+TPM2_StartAuthSession: sessionHandle 0x2000000
+Reading 840 bytes from keyblob.bin
+Storing key at TPM NV index 0x1800202 with password protection
+Public part = 616 bytes
+Private part = 222 bytes
+NV write succeeded
+
+$ ./examples/nvram/read 616 222 -aes
```
dgarske, line 289: Having to supply the size of each public/private part seems problematic. Why not just used `sizeof(TPM2B_PUBLIC)` and `sizeof(TPM2B_PRIVATE)` like the `readKeyBlob` does? If you want to followup with this enhancement in PR #143 that is fine.

examples/nvram/read.c
```diff
@@ -0,0 +1,178 @@
...
+    readSize = keyBlob.pub.size;
+    printf("Trying to read %d bytes of public key part from NV\n", readSize);
+    rc = wolfTPM2_NVReadAuth(&dev, &nv, TPM2_DEMO_NVRAM_STORE_INDEX,
+        (byte*)&keyBlob.pub.publicArea, &readSize, 0);
+    if (rc != 0) goto exit;
+
+    readSize = keyBlob.priv.size;
+    printf("Trying to read %d bytes of private key part from NV\n", readSize);
+    rc = wolfTPM2_NVReadAuth(&dev, &nv, TPM2_DEMO_NVRAM_STORE_INDEX,
+        (byte*)&keyBlob.priv.buffer, &readSize, keyBlob.pub.size);
```
dgarske, line 136: After loading the `keyBlob` can you add code to demonstrate loading the key?

examples/README.md
```diff
@@ -260,3 +268,29 @@ Loaded key to 0x80000001
...
+## Storing keys into the TPM's NVRAM
+
+These examples demonstrates how to use the TPM as secure vault for keys. There are two programs, one to store a TPM key into the TPM's NVRAM and another to extract the key from the TPM's NVRAM. Both examples can use parameter encryption to protect from MITM attacks. The Non-volatile memory location is protected with a password authorization that is passed in encrypted form, when "-aes" or "-xor" is given on the commmand line.
+
+Before running the examples, make sure there is a keyblob.bin generated using the keygen tool. The key can be of any type, RSA, ECC or symmetric. The example will store the private and public part. In case of a symmetric key the public part is meta data from the TPM. How to generate a key you can see above, in the description of the keygen example.
+
+Typical output for storing and then reading an RSA key using parameter encryption:
+
+```
+$ ./examples/nvram/store -aes
```
dgarske, line 281: Can we add support for specifying the key type? Or are you waiting for this to merge so you can do that as part of PR #143 ?

examples/nvram/store.c
```diff
@@ -0,0 +1,170 @@
...
+{
+    int rc;
+    WOLFTPM2_DEV dev;
+    WOLFTPM2_KEYBLOB keyBlob;
+    WOLFTPM2_SESSION tpmSession;
+    WOLFTPM2_HANDLE parent;
+    WOLFTPM2_NV nv;
+    TPM2B_AUTH auth;
+    word32 nvAttributes;
+    const char* filename = "keyblob.bin";
```
dgarske, line 62: Can we make this a command line argument? Or followup with another PR?

dgarske (changes requested):
Looks good, but there are some build errors with `--disable-wolfcrypt`:

```
examples/tpm_test_keys.c:256:33: error: unused parameter 'tpmDevId' [-Werror,-Wunused-parameter]
                            int tpmDevId,
                                ^
examples/tpm_test_keys.c:290:33: error: unused parameter 'tpmDevId' [-Werror,-Wunused-parameter]
                            int tpmDevId,
                                ^
2 errors generated.

src/tpm2.c:109:20: error: unused variable 'handleValue' [-Werror,-Wunused-variable]
    UINT32 authSz, handleValue;
                   ^
src/tpm2.c:112:21: error: unused variable 'handlePos' [-Werror,-Wunused-variable]
    int i, authPos, handlePos;
                    ^
src/tpm2.c:5671:9: error: unused variable 'rc' [-Werror,-Wunused-variable]
    int rc;
        ^
src/tpm2.c:5669:39: error: unused parameter 'nvPublic' [-Werror,-Wunused-parameter]
int TPM2_HashNvPublic(TPMS_NV_PUBLIC* nvPublic, byte* buffer, UINT16* size)
                                      ^
src/tpm2.c:5669:55: error: unused parameter 'buffer' [-Werror,-Wunused-parameter]
int TPM2_HashNvPublic(TPMS_NV_PUBLIC* nvPublic, byte* buffer, UINT16* size)
                                                      ^
src/tpm2.c:5669:71: error: unused parameter 'size' [-Werror,-Wunused-parameter]
int TPM2_HashNvPublic(TPMS_NV_PUBLIC* nvPublic, byte* buffer, UINT16* size)
                                                                      ^
6 errors generated.
```

examples/README.md
tomoveu: Good question. As far as I know, TPM2B_PUBLIC is of size capable to take in the largest possible public key. Therefore, I thought having an argument to specify the size is a good approach. Similar for TPM2B_PRIVATE. I will check this and if I can avoid these manual arguments I will as addition to 143 Thanks for the hint!

examples/nvram/read.c
tomoveu: Good point. I will address that in 143 as well

examples/README.md
tomoveu: Good idea, we could generate RSA or ECC. I would address this in 143 and we can have the parameter encryption fix merged here.

examples/nvram/store.c
tomoveu: Let's follow-up with another PR for NVRAM examples enhancements

tomoveu:
@dgarske good recommendations for enhancements on the NVRAM examples, I will do them in follow-up commit.

I added fixes for when wolfCrypt is disabled. Thank you for catching this 💯

dgarske (approved):

examples/README.md
tomoveu: @dgarske coming back to this and related to #143 

I noticed that writeKeyBlob and readKeyBlob used in nvram/read use sizeof(key->pub) instead of the actual size of the Key Public Part. While this is okay for storing a file to disk, I am not sure if this is okay for using the NVRAM that on some TPMs is few kB only.

Example below:

In both cases, writeKeyBlob will write 616 bytes of public area, while the actual public area for the symmetric key is only 50 bytes and for RSA is only 278 bytes.

Should I just use sizeof(TPM2B_PUBLIC) for nvram/read or I should try to change the write/readKeyBlob helpers?

```
dimitartomov@Dimitars-MacBook-Pro wolfTPM % ./examples/keygen/keygen -sym
TPM2.0 Key generation example
	Key Blob: keyblob.bin
	Algorithm: SYMCIPHER
		AES, 256bits, CTR mode
	Template: Default
	Use Parameter Encryption: NULL
Loading SRK: Storage 0x81000200 (282 bytes)
AES Symmetric template with 256 bits, Counter mode
Creating new SYMCIPHER key...
Created new key (pub 50, priv 158 bytes)
Wrote 776 bytes to keyblob.bin
dimitartomov@Dimitars-MacBook-Pro wolfTPM % ./examples/keygen/keygen -t  
TPM2.0 Key generation example
	Key Blob: keyblob.bin
	Algorithm: RSA
	Template: Default
	Use Parameter Encryption: NULL
Loading SRK: Storage 0x81000200 (282 bytes)
RSA template
Creating new RSA key...
Created new key (pub 278, priv 222 bytes)
Wrote 840 bytes to keyblob.bin
dimitartomov@Dimitars-MacBook-Pro wolfTPM %
```

examples/README.md
tomoveu: @dgarske RSA ECC or Symmetric key all work with nvram/store,

because nvram/store uses readKeyBlob

and whatever is written in the KeyBlob->pub/priv.size that is then stored into the NVRAM (size-wise).

Did you mean to be able to store only the private material (-priv option perhaps)?

Could be useful for symmetric keys. Also -pub in cases when only the pub part has to be stored in NVRAM.

examples/nvram/store.c
tomoveu: nvram/store already accepts filename, but nvram/read does not. I will make sure to add that.

PR #147, opened 2021-03-04
KDFa avoid overrun

elms:
Error reproduced by `CFLAGS="-fsanitize=address" ./configure --enable-swtpm`
and running `./scripts/swtpm_sim.test`

### DIFF
--- a/tests/unit_tests.c
+++ b/tests/unit_tests.c
@@ -61,7 +61,7 @@
 #define AssertInt(x, y, op, er) do {                                           \
     int _x = (int)x;                                                           \
     int _y = (int)y;                                                           \
-    Assert(_x op _y, ("%s " #op " %s", #x, #y), ("%d " #er " %d", _x, _y));    \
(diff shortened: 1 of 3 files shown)


dgarske (approved):

tests/unit_tests.c
```diff
@@ -198,10 +199,9 @@ static void test_wolfTPM2_ReadPublicKey(void)
     /* Test success: read storage primary key */
     rc = wolfTPM2_ReadPublicKey(&dev, &storageKey,
         TPM2_DEMO_STORAGE_KEY_HANDLE);
-    if (rc == TPM_RC_HANDLE) {
+    if ((rc & 0xff) == TPM_RC_HANDLE) {
```
elms, line 202: This is one work around that, may need a closer look. ibmswtpm returns `0x18b`

elms: maybe?
```suggestion
    if ((rc & RC_MAX_FMT1) == TPM_RC_HANDLE) {
```

elms: Done

dgarske (approved):

PR #148, opened 2021-03-04
wrap: Fix older NV Read/Write API

elms:
use nvIndex for nv handle ignore authHandle

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -2723,7 +2723,8 @@ int wolfTPM2_NVWrite(WOLFTPM2_DEV* dev, TPM_HANDLE authHandle,
 {
     WOLFTPM2_NV nv;
     XMEMSET(&nv, 0, sizeof(nv));
-    nv.handle.hndl = authHandle;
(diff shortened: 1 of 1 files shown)


elms:
I'm not confident about this change, but examples runs now with swtpm and I think jenkins test was failing on the same example.

tomoveu (approved):
Good fix @elms

@ nv->handle.hndl = nvIndex

TPM2_NV_Write has two Handle slots. One for the NV index to be written and second from where an authorization will be used. When the auth is associated with the NV Handle and is not provided through other means, then both handles have nvIndex as their value. Therefore, your fix is good. It would be only in the case of providing auth using other means that this will differ and currently we do not have such use case, nor it is common. So, I am 100% comfortable with that change. 

@ ignore authHandle  

This was so since commit 4f1fb41d. I think the reason we did not see this earlier is because:
1. authHandle is used in an expression with nv.handle.hndl
2. No one was using the older API since we moved to the new API with Auth.

It is great we have complete backward compatibility of old and new API. 

@dgarske I confirm that this PR is a good fix

dgarske (approved):

PR #155, opened 2021-03-26
Add ST33 extra GPIO support

tomoveu:
This PR adds support for the extra GPIO available on some TPMs, such as the ST33

My intention is to have
* gpio/config - configuring TPM's extra GPIO
* gpio/read - reading pin level
* gpio/set - setting pin level


examples/gpio/gpio.c
```diff
@@ -0,0 +1,140 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/* This examples demonstrates the use of GPIO available on some TPM modules */
+
+#include <wolftpm/tpm2_wrap.h>
+
+#ifndef WOLFTPM2_NO_WRAPPER
```
dgarske, line 26: You'll get a build error with missing TPM2_GPIO_Config if these are not also defined `#if defined(WOLFTPM_ST33) || defined(WOLFTPM_AUTODETECT)`. Might consider making `TPM2_GPIO_Config` always available and add the ST33 specific stuff inside the function. When not compiled in use `NOT_COMPILED_IN` return code.

tomoveu:
Code is refactored

[ready]
* gpio/config - Configures GPIO and creates NV index for R/W access to the GPIO

* gpio/set - Sets GPIO level (low/high)

* gpio/unconfig - Destroy  NV Index

[ongoing]

* gpio/read - Reads GPIO level

For some reason the read operation fails on the NV index with NV uninitialised. Investigating

dgarske (changes requested):

tomoveu:
@dgarske thank you for the review. All feeedback is taken in. PR is ready for review.

tomoveu:
NV Index for GPIO input requires and extra step - added and tested. Working good.

Test uses GPIO0 as input in pull-up configuration
and we successfully read the high GPIO level.

Ready for final review @dgarske 

```
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/config 0 2
GPIO num is: 0
GPIO mode is: 2
Demo how to use extra GPIO on a TPM 2.0 modules
wolfTPM2_Init: success
Trying to configure GPIO0...
TPM2_GPIO_Config success
NV Index for GPIO access created
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/read 
GPIO0 is High.
```

dgarske (changes requested):
All looks great and tests correctly. The one annoying thing was not being able to call config without first calling unconfig (gets `TPM_RC_NV_DEFINED`). I would prefer config did the NV undefine first. Add option to config app to allow "unconfig / remove NV" and eliminate the unconfig app.

tomoveu:
@dgarske makes sense. Good idea! I will do it right away.

tomoveu:
@dgarske Please review and if all is good I will squash the commits before you merge.

examples/gpio/config.c
```diff
@@ -0,0 +1,248 @@
...
+    printf("./examples/gpio/gpio [num] [mode]\n");
+    printf("* num is a GPIO number between 0-3 (default %d)\n", TPM_GPIO_A);
+    printf("* mode is a number selecting the GPIO mode between 0-5 (default %d):\n", TPM_GPIO_MODE_STANDARD);
+    printf("\t0. standard - reset to the GPIO's default mode\n");
+    printf("\t1. floating - input in floating configuration.\n");
+    printf("\t2. pullup   - input with pull up enabled\n");
+    printf("\t3. pulldown - input with pull down enabled\n");
+    printf("\t4. opendrain - output in open drain configuration\n");
+    printf("\t5. pushpull  - output in push pull configuration\n");
+    printf("Demo usage, without parameters, configures GPIO%d as pushpull output.\n", TPM_GPIO_A);
```
dgarske, line 51: Please add another option `6` to `unconfigure` GPIO.

examples/gpio/config.c
```diff
@@ -0,0 +1,248 @@
...
+
+#ifdef DEBUG_WOLFTPM
+    printf("Trying to remove NV index 0x%8.8X used for GPIO\n", nvIndex);
+#endif
+    /* Make sure NV Index for this GPIO is cleared before use
+     * This way we make sure a new GPIO config can be set
+     */
+    rc = wolfTPM2_NVDelete(&dev, TPM_RH_OWNER, nvIndex);
+#ifdef DEBUG_WOLFTPM
+    if (rc == TPM_RC_SUCCESS) {
```
dgarske, line 157: Please do not report an error if the NV does not exist. scan-build will report warning `rc` set but not read. Please keep some logic in here for non-debug.

examples/gpio/config.c
```diff
@@ -0,0 +1,248 @@
...
+            if (gpioMode > 0 && gpioMode < 4) {
+                gpioInput = 1;
+            }
+            gpioMode += TPM_GPIO_MODE_STANDARD;
+            /* Preparing to process next argument */
+            argc--;
+        }
+        if (argc == 2) {
+            gpioNum = atoi(argv[1]);
+            if (gpioNum < 0 || gpioNum > 3) {
```
dgarske, line 94: Can we make the GPIO pin number min/max macros? These 0-3 are hard coded several places and should have option to change at build-time. Actually looks like min is always `TPM_GPIO_A` and max just needs added in tpm2.h.

examples/gpio/read.c
```diff
@@ -0,0 +1,137 @@
...
+    /* Prep NV Index and its auth */
+    nv.handle.hndl = nvIndex;
+    nv.handle.auth.size = sizeof(gNvAuth)-1;
+    XMEMCPY(nv.handle.auth.buffer, (byte*)gNvAuth, nv.handle.auth.size);
+    parent.hndl = TPM_RH_OWNER;
+    /* Read GPIO state */
+    readSize = sizeof(pinState);
+    rc = wolfTPM2_NVReadAuth(&dev, &nv, nvIndex, &pinState, &readSize, 0);
+    if (rc != 0) {
+            printf("Error while reading GPIO state\n");
```
dgarske, line 93: Check indent.

examples/gpio/config.c
```diff
@@ -0,0 +1,248 @@
...
+            TPM2_GetRCString(rc));
+        goto exit;
+    }
+    printf("TPM2_GPIO_Config success\n");
+
+    /* Configure NV Index for access to this GPIO */
+    XMEMSET(&nv, 0, sizeof(nv));
+    XMEMSET(&parent, 0, sizeof(parent));
+    /* Prep NV attributes */
+    parent.hndl = TPM_RH_OWNER;
```
dgarske, line 196: Curious the GPIO command uses platform, but the NV one uses OWNER. Have you tried platform? Might be good to add a comment here re this.

examples/gpio/config.c
```diff
@@ -0,0 +1,248 @@
...
+    rc = wolfTPM2_GetNvAttributesTemplate(parent.hndl, &nvAttributes);
+    if (rc != 0) {
+        printf("Setting NV attributes failed\n");
+        goto exit;
+    }
+    /* Define NV Index for GPIO */
+    rc = wolfTPM2_NVCreateAuth(&dev, &parent, &nv, nvIndex, nvAttributes,
+                               sizeof(BYTE), (byte*)gNvAuth, sizeof(gNvAuth)-1);
+    if (rc != 0 && rc != TPM_RC_NV_DEFINED) {
```
dgarske, line 205: Does it make sense to still skip on the `TPM_RC_NV_DEFINED` error here?

examples/gpio/config.c
```diff
@@ -0,0 +1,248 @@
...
+        printf("Creating NV Index for GPIO acccess failed\n");
+        goto exit;
+    }
+    printf("NV Index for GPIO access created\n");
+
+    /* GPIO configured as an input, requires an extra configuration step */
+    if (gpioInput) {
+        rc = wolfTPM2_NVWriteAuth(&dev, &nv, nvIndex, &dummy, sizeof(dummy), 0);
```
dgarske, line 213: Please add note that "dummy" value is discarded for GPIO and is required for the NV read to work on GPIO.

dgarske (changes requested):

examples/gpio/config.c
tomoveu: Platform auth is required to enable Platform specific feature. NV indexes are user owned aka owner. For example, the user would not want unauthorized change of a GPIO signal from the TPM to another subsystem. I will add a comment explaining these details briefly.

examples/gpio/config.c
```diff
@@ -0,0 +1,260 @@
...
+    printf("* num is a GPIO number between %d-%d (default %d)\n", GPIO_NUM_MIN, GPIO_NUM_MAX, TPM_GPIO_A);
+    printf("* mode is a number selecting the GPIO mode between 0-5 (default %d):\n", TPM_GPIO_MODE_STANDARD);
+    printf("\t0. standard - reset to the GPIO's default mode\n");
+    printf("\t1. floating - input in floating configuration.\n");
+    printf("\t2. pullup   - input with pull up enabled\n");
+    printf("\t3. pulldown - input with pull down enabled\n");
+    printf("\t4. opendrain - output in open drain configuration\n");
+    printf("\t5. pushpull  - output in push pull configuration\n");
+    printf("\t6. unconfigure - delete the NV index for the selected GPIO\n");
+    printf("Demo usage, without parameters, configures GPIO%d as pushpull output.\n", TPM_GPIO_A);
```
dgarske, line 55: Can we change the "demo" words to example? There are a few places.

dgarske (changes requested):
Also from your notes... please squash into two commits:

```
1) The first commit should be maintained 9f7c80 "Added vendor command ..." as it works only on tpm2.c/h
2) Everything else can be squashed into "Added new example to demo TPM 2.0 extra GPIO control"
```

tomoveu:
@dgarske commits squashed, ready for merge.

tomoveu:
@dgarske commits squashed and ready for merge.

PR #156, opened 2021-03-26
Support for QNX with wolfTPM

dgarske:


src/tpm2_wrap.c
```diff
@@ -1170,12 +1188,21 @@ int wolfTPM2_SensitiveToPrivate(TPM2B_SENSITIVE* sens, TPM2B_PRIVATE* priv,
         Aes enc;
         TPM2B_IV ivField;
         TPM2B_SYM_KEY symKey;
+        TPM2_CTX* ctx = TPM2_GetActiveCtx();
 
         /* Generate IV */
         ivField.size = digestSz;
-        rc = TPM2_GetNonce(ivField.buffer, ivField.size);
-        if (rc != 0)
+        rc = TPM2_AcquireLock(ctx);
```
tomoveu, line 1195: I do not understand why is active context needed here? 
Is it because of TPM2_GetNonce? 
If yes, shouldn't then the mutex be acquired inside TPM2_GetNonce?

tomoveu (commented):
I think the lock for TPM2_GetNonce should be inside the function. I could be missing something.

@elms could you please take a second look and review the new QNX callback

Thanks,
Dimi

configure.ac
```diff
@@ -341,7 +341,13 @@ then
     AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_SMALL_STACK"
 
     # Reduces max packet and buffer sizes to 1024 bytes
-    AM_CFLAGS="$AM_CFLAGS -DMAX_COMMAND_SIZE=1024 -DMAX_RESPONSE_SIZE=1024 -DWOLFTPM2_MAX_BUFFER=1500 -DMAX_SESSION_NUM=1 -DMAX_DIGEST_BUFFER=973"
+    AM_CFLAGS="$AM_CFLAGS -DMAX_COMMAND_SIZE=1024 -DMAX_RESPONSE_SIZE=1024 -DWOLFTPM2_MAX_BUFFER=1500 -DMAX_DIGEST_BUFFER=973"
+
+    # If parameter encryption is not used then maximum session count is one
```
elms, line 346: Not sure I follow this, perhaps the comment can explain why max sessions is related to parameter encryption.

IDE/QNX/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+# The QNX SPI Driver
+LIBS += -lspi-master
+```
+
+4) Create `inc/user_settings.h` for all wolf specific settings:
+
+Here is a template:
+
+```
```
elms, line 29: nit: as a consumer of this I'd prefer the templates and diffs as files that I could copy and patches I could apply directly.

IDE/QNX/README.md
```diff
@@ -0,0 +1,147 @@
...
+#define NO_DES3
+
+#endif /* !WOLF_USER_SETTINGS_H */
+```
+
+5) wolfTPM HAL: Use either `tpm_io.c` directly or copy the required HAL interface into your own .c.
+
+6) Add wolfTPM example code into your own .c.
+
+## QNX SPI Master Patch for Manual Chip Select
```
elms, line 88: Please add a explanation of why manual chip select is necessary

examples/tpm_io.c
```diff
@@ -623,6 +630,95 @@
...
+            spi_close(fd);
+            return ret;
+        }
+
+    #ifdef WOLFTPM_CHECK_WAIT_STATE
+        /* Send Header */
+        status = spi_xchange(fd, SPI_DEV_DEFAULT, 
+            (byte*)txBuf, rxBuf, TPM_TIS_HEADER_SZ);
+        if (status == -1) {
+            spi_close(fd);
```
elms, line 674: If I understand correctly, in the error cases, the CS could still be asserted. As part of clean up should the config be cleared as at the end on success?
```
        cfg.mode |= (SPI_MODE_MAN_CS | SPI_MODE_CLEAR_CS);
        (void)spi_setcfg(fd, SPI_DEV_DEFAULT, &amp;cfg);
```

elms (commented):
I'm with @tomoveu on not quite tracking the changes around the lock and nonce. Perhaps I'm not familiar enough with that area. Can you describe the issue and how this change resolves it?

dgarske:
@elms I pulled out the no RNG code so this is QNX plus minor build fixes only. Can you please review? Thanks

configure.ac
dgarske: With parameter encryption enabled the max sessions needs to be the default (3). If it is not enabled we can reduce the max sessions to 1.

IDE/QNX/README.md
dgarske: Sounds nice, but there isn't a specific location for this user supplied file.

IDE/QNX/README.md
dgarske: Added

examples/tpm_io.c
dgarske: Correct. This manually clears the CS.

elms: If the call to `spi_xchange` fails here, I think it's possible that the CS is left asserted. Does `spi_close` guarantee deasserting or should we have a common clean block that is always executed (either by goto or falling through to a single `return`)?

dgarske: Fixed

elms (approved):

PR #159, opened 2021-04-21
make: Rename GPIO example source files to avoid conflict

elms:
New GPIO examples `read` conflicts with NVRAM example on `make install`.

Resolves conflict with "./examples/nvram/read.c"


dgarske (changes requested):
Can we rename the more recent example for GPIO instead? I see the conflict on gpio/read.c and nvram/read.c. Suggest gpio_read.c and gpio_set.c.

dgarske (changes requested):
Thank you, that's great. Can you also update the examples/README.md with the new names?

dgarske (approved):

tomoveu:
Thank you @elms for catching this! 🏅

PR #163, opened 2021-05-06
Improvements to the Windows TBS documentation

dgarske:
Also fix for link the `examples/README.md`.

### DIFF
--- a/docs/WindowTBS.md
+++ b/docs/WindowTBS.md
@@ -2,6 +2,14 @@
 
 wolfTPM can be built to use Windows native TBS (TPM Base Services)
 
+When using the Windows TBS interface the NV access is blocked by default. TPM NV storage space is very limited and when filled can cause undefined behaviors, such as failures loading key handles. These are not managed by TBS.
+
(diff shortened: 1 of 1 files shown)


docs/WindowTBS.md
```diff
@@ -2,6 +2,14 @@
 
 wolfTPM can be built to use Windows native TBS (TPM Base Services)
 
+When using the Windows TBS interface the NV access is not permitted. This is either because the TPM is started under a locality that does not allow it or the TBS blocks it. We suspect the reason is because the TPM NV storage space is very limited and when filled caused undefined behaviors, such as failures loading key handles.
```
elms, line 5: I think NV commands are blocked by default but can be enabled using registry keys. This [stackoverflow](https://stackoverflow.com/a/61698844) mentions it. But I do think we should dissuade people from using it.

The only mention in mircosoft docs I really found was:

> Limited TPM resources that are not virtualized by the TBS (such as counters and NV store) must be cooperatively shared among software stacks.
[link](https://docs.microsoft.com/en-us/windows/win32/tbs/resource-virtualization)

Maybe something like
```suggestion
When using the Windows TBS interface the NV access is blocked by default. TPM NV storage space is very limited and when filled can cause undefined behaviors, such as failures loading key handles. These are not managed by TBS.
```

elms (changes requested):
Good catch on the link to examples. I think we can refine why NV is blocked. Good additions!

elms (approved):

PR #165, opened 2021-05-12
Quickfix of invalid pcrIndex number

tomoveu:
Typical TPM 2.0 implementations have only 24 PCR registers.

Before, wolfTPM2_ReadPCR would fail with TPM_RC_VALUE if invalid value is given.

Per a user request, we modify wolfTPM2_ReadPCR, so it does not accept invalid values.

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -2462,9 +2462,9 @@ int wolfTPM2_ReadPCR(WOLFTPM2_DEV* dev, int pcrIndex, int hashAlg, byte* digest,
     int rc;
     PCR_Read_In  pcrReadIn;
     PCR_Read_Out pcrReadOut;
-    int digestLen;
+    int digestLen = 0;
 
-    if (dev == NULL)
+    if (dev == NULL || pcrIndex < PCR_FIRST || pcrIndex > PCR_LAST)
         return BAD_FUNC_ARG;
 
     /* set session auth to blank */


tomoveu:
Hi @elms, Could you please review and merge if good?

I saw that we are both on the same wolfSSL forum thread where this originated.

src/tpm2_wrap.c
```diff
@@ -2462,9 +2462,9 @@ int wolfTPM2_ReadPCR(WOLFTPM2_DEV* dev, int pcrIndex, int hashAlg, byte* digest,
     int rc;
     PCR_Read_In  pcrReadIn;
     PCR_Read_Out pcrReadOut;
-    int digestLen;
+    int digestLen = 0;
```
dgarske, line 2465: Don't understand the change to init this... it is always set before use below.

dgarske (approved):

elms (approved):

src/tpm2_wrap.c
tomoveu: In the wolfSSL forum, Hadrien mentioned that for him, the TPM chip on his computer does not return TPM_RC_VALUE when pcrIndex is above 23. The code continues execution and the following XMEMCPY is executed with uninitialized digestLen leading to a crash. Therefore, I thought it is a good thing to initialize digestLen to zero, to eliminate potential crashes for the user.

PR #166, opened 2021-05-13
Add support for `--with-wolfcrypt=/dir/` and fix minor build error/warn

dgarske:
* Fix for build error with `--disable-wolfcrypt`.
* Fix for unused `auth` in store example.
* ZD 12204


tomoveu (commented):
Having `--with-wolfcrypt=/dir/` is so valuable. Especially, when developing under macOS and I have homebrew wolfcrypt, pip3 wolfcrypt and my own wolfcrypt build. 

I have not tested the new configure script, otherwise code changes look good.

tomoveu (changes requested):
Hi David,

Please add information in the configure output about the wolfcrypt path being used. Right now there is only "wolfcrypt: yes".

I have one corner case, let me know if this new option is intended to address this as well:

> Configure ignores incorrect --with-wolfcrypt if there is wolfSSL installed on the system

I tested under MacOS like this:

> 
> Negative test
> - I specify a wrong wolfcrypt path
> - Configure succeeds regardless. 
> - The configure script does not inform me of the used wolfcrypt path.
> 

> 
> Positive test 
> - I specify a valid wolfcrypt path 
> - The configure script seems to use it as expected
> - I did not get a confirmation of the wolfcrypt path used
> 

It seems the configure script uses the wolfSSL/wolfcrypt I have installed on my MacOS and ignores the supplied -with-wolfcrypt=path

dgarske:
@tomoveu , yes if an invalid path is specified it uses the default (/usr/local). I've added the used path as configure output log entry.

tomoveu (approved):
Tested. Working. 

Few whitespaces in the new code.

Good to go @dgarske :)

PR #171, opened 2021-06-03
Refactor examples/tpm_io into separate platform source files

tomoveu:
This is the crude work, now remains the details in automake and testing

* tpm_io_atmel.c contains only ASF code
* tpm_io_barebox.c contains only code for Barebox
* tpm_io_st.c to contain only STM32 specific code
* tpm_io_linux.c to contain only Linux specific code
* tpm_io_qnx includes the new QNX callback
* tpm_io_xilinx.c to contain only Xilinx code

- Noted that native Linux /dev/tpmX does not require an IO callback
- TPM Simulator and Windows TBS also does not require an IO callback


dgarske (changes requested):
Awesome work. A couple things:
1) The new files need added to examples/include.am
2) I would prefer a single `examples/tpm_io.h` for backwards compatibility. So all the functions should be in the single header. Each platform can have its own .c.

Also good to note somewhere the tpm_io.c and the tpm_io_[platform].c are both required.

dgarske:
@tomoveu also looks like you will need to add all tpm_io_[platform].c files to all the examples _SOURCES.

dgarske:
@tomoveu . Also with --enable-swtpm get issue with `#define TPM2_IoCb NULL` and function in tpm_io.c conflicting.

```
examples/tpm_io.c:170:5: error: expected ')'
int TPM2_IoCb(TPM2_CTX* ctx, const byte* txBuf, byte* rxBuf,
```

tomoveu:
@dgarske Oh, thanks! Good points.

 I thought automake could include the correct tpm_io_platform.c at compile time?

Thus, include.am would require only tpm_io.c/.h Hm, I may have overestimated how versatile automake is.

tomoveu:
- Kept single tpm_io.h for backward compatibility
- Fixed missing ifdefs for TPM Simulator case
- TPM Simulator and Native Linux are working.

@elms  could you please quickly run native tests for the Windows port?

examples/pcr/include.am
```diff
@@ -12,18 +12,37 @@ noinst_HEADERS  += examples/pcr/quote.h \
...
 examples_pcr_extend_DEPENDENCIES = src/libwolftpm.la
 
 examples_pcr_reset_SOURCES      = examples/pcr/reset.c \
                                   examples/tpm_io.c \
+                                  examples/tpm_io_atmel.c \
+                                  examples/tpm_io_barebox.c \
+                                  examples/tpm_io_linux.c \
+                                  examples/tpm_io_st.c \
+                                  examples/tpm_io_qnx.c \
+                                  examples/tpm_io_xilinx.c
```
dgarske, line 44: Automake bug here.

dgarske (changes requested):
I would prefer to see the include.am changes reverted and have the tpm_io.c include the other IO .c files directly. Then use a bit of macro magic. I've pushed this improvement.

dgarske (approved):

elms:
@tomoveu Testing on Windows looks like it operates the same as master.

PR #172, opened 2021-06-04
Add GPIO example for NPCT7xx

tomoveu:
UPDATE: WolfTPM Supports GPIO for NPCT75x

Draft PR - Added vendor commands to configure GPIO

It seems like NPCT7xx support only GPIO Output mode and configuration happens in a completely different way than we have right now in gpio/gpio_config.c Therefore, I am thinking of adding newgpio/gpio_npct7xx.c example. 

First, I will confirm my suspicion before moving forward.

### DIFF
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -63,7 +63,9 @@ static TPM_RC TPM2_AcquireLock(TPM2_CTX* ctx)
 
     if (!ctx->hwLockInit) {
         if (wc_InitMutex(&ctx->hwLock) != 0) {
-            WOLFSSL_MSG("TPM Mutex Init failed");
+        #ifdef DEBUG_WOLFTPM
+            printf("TPM Mutex Init failed\n");
+        #endif
             return TPM_RC_FAILURE;
         }
         ctx->hwLockInit = 1;
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -5225,6 +5225,61 @@ int TPM2_GPIO_Config(GpioConfig_In* in)
     }
     return rc;
 }
+#elif defined(WOLFTPM_NUVOTON)
+int TPM2_NTC2_PreConfig(NTC2_PreConfig_In* in)
+{
+    int rc;
+    TPM2_CTX* ctx = TPM2_GetActiveCtx();
+
+    if (in == NULL || ctx == NULL)
+        return BAD_FUNC_ARG;
+
+    rc = TPM2_AcquireLock(ctx);
+    if (rc == TPM_RC_SUCCESS) {
+        CmdInfo_t info = {
+            .inHandleCnt = 1,
+        };
+        TPM2_Packet packet;
+        TPM2_Packet_Init(ctx, &packet);
+        /* Process the auth handle for GPIO configuration */
+        TPM2_Packet_AppendU32(&packet, in->authHandle);
+        info.authCnt = TPM2_Packet_AppendAuth(&packet, ctx);
+        /* Process the NPCT7xx configuration */
+        TPM2_Packet_AppendBytes(&packet, (byte*)&in->preConfig, sizeof(in->preConfig));
+        TPM2_Packet_Finalize(&packet, TPM_ST_SESSIONS, TPM_CC_NTC2_PreConfig);
+
+        /* Send the new NPCT7xx configuration */
+        rc = TPM2_SendCommandAuth(ctx, &packet, &info);
+
+        TPM2_ReleaseLock(ctx);
+    }
+    return rc;
+}
+
+int TPM2_NTC2_GetConfig(NTC2_GetConfig_Out* out)
+{
+    int rc;
+    TPM2_CTX* ctx = TPM2_GetActiveCtx();
+
+    if (out == NULL || ctx == NULL)
+        return BAD_FUNC_ARG;
+
+    rc = TPM2_AcquireLock(ctx);
+    if (rc == TPM_RC_SUCCESS) {
+        TPM2_Packet packet;
+        TPM2_Packet_Init(ctx, &packet);
+        TPM2_Packet_Finalize(&packet, TPM_ST_NO_SESSIONS, TPM_CC_NTC2_GetConfig);
+
+        /* Request the current NPCT7xx configuration */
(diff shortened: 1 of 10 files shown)


src/tpm2.c
```diff
@@ -5225,6 +5225,61 @@ int TPM2_GPIO_Config(GpioConfig_In* in)
     }
     return rc;
 }
+#elif defined(WOLFTPM_NPCT7xx)
+WOLFTPM_API int TPM2_NTC2_PreConfig(NTC2_PreConfig_In* in)
```
dgarske, line 5229: The .c files do not need `WOLFTPM_API`

dgarske (changes requested):
GPIO examples don't work with `--enable-nuvoton`. The GPIO examples also need to check for `WOLFTPM_NPCT7xx`. Also we should update configure.ac to somehow add `NPCT7xx`. Example `--enable-nuvoton=7xx`?

```
$ ./configure --enable-nuvoton --enable-debug --disable-shared && make
$ ./examples/gpio/gpio_config -?
Extra GPIO is supported only on ST33 TPM 2.0 modules.
```

dgarske (changes requested):
Also in the GPIO examples there is a check for the vendor which needs fixed.

```
    /* Enable TPM2_GPIO_Config command */
    if (caps.mfg != TPM_MFG_STM) {
        printf("Extra GPIO is supported only on ST33 TPM 2.0 modules\n");
        goto exit;
    }
```

Also consider not using `WOLFTPM_NPCT7xx` and instead using `WOLFTPM_NUVOTON` and then run-time trapping the vendor command returning TPM_RC_COMMAND_CODE

tomoveu:
Feedback taken in. Testing it tomorrow on the real hardware.

wolftpm/tpm2.h
```diff
@@ -2754,6 +2757,26 @@ WOLFTPM_API TPM_RC TPM2_NV_Certify(NV_Certify_In* in, NV_Certify_Out* out);
 
 /* Vendor Specific API's */
 #if defined(WOLFTPM_ST33) || defined(WOLFTPM_AUTODETECT)
+
```
dgarske, line 2760: Why did this section move up and not near `TPMI_GPIO_NAME`?

dgarske (changes requested):
@tomoveu , I would prefer a single set of GPIO examples. If there are vendor differences they should be handled in the tpm2_wrap.c. Please work with Nuvoton to resolve any issues if you are only getting output to work.

wolftpm/tpm2.h
tomoveu: I moved it up, so both vendor specific sections have the same coding style.  Please see the Nuvoton part below. I could modify this, of course.

tomoveu:
@dgarske I've already contacted Nuvoton, waiting on reply to confirm it is a design limitation. I suspect only output modes are supported. 

Our example:
- ST use separate configuration structures
- Nuvoton use a single structure for every setting and bitfields

I am worried that if I have added the Nuvoton code in the existing gpio_config example, the code will not be very readable. Because there is no matching of values and defines.

dgarske (approved):
PR looks good, but I would like to hold off merging until Nuvoton helps confirm the NPCT7xx only supports GPIO outputs.

tomoveu:
@dgarske yes, please hold. 

I am in communication with Nuvoton about GPIO modes and I also need to confirm some NV attributes, as well.

tomoveu:
Changes pushed based on vendor feedback: 
* TPM_RH_PLATFORM should be used , because the NV Index is Platform create

Result:
* Can create NV Index after GPIO configuration
* Can set high GPIO level 
* Can read back GPIO level

Failure:
* Can not delete NV Index (maybe because it is PLATFORM create?)
* Can not set low GPIO level

Progress made. Must remain on hold @dgarske until we fix the two remaining issues.

tomoveu:
@dgarske Ready for merge 👍 

Details:
* I found a typo in my code for reading and setting the GPIO.
* Reconfiguring GPIO on NPCT75x does not require NV Index deletion.
* Updated examples README.md 

Test output from the real hardware:
```
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/gpio_nuvoton 4 1
Example for GPIO configuration of a NPTC7xx TPM 2.0 module
GPIO number: 4
GPIO mode: 1
wolfTPM2_Init: success
First, the current NPCT7xx config will be read
then modified with the new GPIO configuration
Successfully read the current NPCT7xx configuration
NTC2_PreConfig success
NV Index for GPIO access created
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/gpio_set 4 -low
GPIO4 set to low level
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/gpio_read 4
GPIO4 is Low.
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/gpio_set 4 -high
GPIO4 set to high level
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/gpio_read 4
GPIO4 is High.
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/gpio_nuvoton 4 2
Example for GPIO configuration of a NPTC7xx TPM 2.0 module
GPIO number: 4
GPIO mode: 2
wolfTPM2_Init: success
First, the current NPCT7xx config will be read
then modified with the new GPIO configuration
Successfully read the current NPCT7xx configuration
NTC2_PreConfig success
NV Index for GPIO access created
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/gpio_read 4
GPIO4 is Low.
pi@raspberrypi:~/wolftpm $ sudo ./examples/gpio/gpio_nuvoton -h
Expected usage:
./examples/gpio/gpio_config [num] [mode]
* num is a GPIO number between 3 and 4 (default 3)
* mode is either push-pull, open-drain or open-drain with pull-up
	1. pushpull  - output in push pull configuration
	2. opendrain - output in open drain configuration
	3. pullup - output in open drain with pull-up enabled
	4. unconfig - delete NV index for GPIO access
Example usage, without parameters, configures GPIO3 as push-pull output.
pi@raspberrypi:~/wolftpm $ 
```

dgarske (approved):

PR #173, opened 2021-06-04
Do not use `WOLFSSL_MSG`

dgarske:
Fixes:
```
/usr/bin/ld: src/.libs/libwolftpm.a(libwolftpm_la-tpm2.o): in function `TPM2_AcquireLock':
/home/pi/wolftpm/src/tpm2.c:66: undefined reference to `WOLFSSL_MSG'
```


embhorn (approved):
Looks fine. Thanks, David.

PR #175, opened 2021-06-18
Added option to create TPM keys with OAEP scheme for SSH protocol

tomoveu:
@dgarske please take a look at this addition

OAEP with SHA1 is used by SSH and supported by TPM 2.0 (as it was by TPM 1.2 too)

For some reason, the TPM does not like the Key attributes. I tried with and without fixedTPM.

Note: the new helper Template_SSH sets SHA256 for TPM key Name computation and SHA1 for the signature scheme. 

```
Failure 0x2c2: TPM_RC_ATTRIBUTES: Inconsistent attributes
```

### DIFF
--- a/examples/keygen/keygen.c
+++ b/examples/keygen/keygen.c
@@ -52,6 +52,7 @@ static void usage(void)
     printf("* -sym: Use Symmetric Cypher for key generation\n");
     printf("\tDefault Symmetric Cypher is AES CTR with 256 bits\n");
     printf("* -t: Use default template (otherwise AIK)\n");
+    printf("* -ssh: Use template for SSH key (only RSA algorithm)\n");
     printf("* -aes/xor: Use Parameter Encryption\n");
     printf("Example usage:\n");
     printf("\t* RSA, default template\n");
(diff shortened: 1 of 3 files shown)


tomoveu:
I think I found the source of the issue, but have to verify

TPM 2 spec, Part 1 , page 230

> B.4 RSAES_OAEP
> This encryption scheme is defined in IETF RFC 3447. It is the only scheme used with an RSA-restricted
> decryption key. The algorithm identifier for this scheme is TPM_ALG_OAEP.
> For RSA keys protecting a secret value (such as, an encryption key or a session secret), the L parameter
> is a null-terminated string indicating the intended use of the encrypted value. A command that accepts or
> creates an RSA-encrypted secret indicates the value of the string to use for L. The RSA key’s Name
> algorithm is used to compute lhash := H(L), and the null termination octet is included in the digest.
> MGF1 (as defined in IEEE Std 1363TM-2000) computes dbMask and seedMask. The mask-generation
> function uses the Name algorithm of the RSA key as the hash algorithm
>

tomoveu:
Ready for review @dgarske 

1. Spec is right - OAEP key can be only decrypt , not signing
2. TPMA_OBJECT_SENSITIVE_ORIGIN was missing too

fixed and working.

tomoveu:
@dgarske please hold on merging until i confirm SSH works with these keys

There is something about the padding and singing with TPM OAEP key that I am figuring out on TPM.dev

dgarske (approved):
Looks good. Holding merge until wolfSSH testing.

dgarske:
Closing to open new PR.

PR #176, opened 2021-06-23
Refactor the assignment of structs to use memcpy

dgarske:
This resolves strict pointer alignment issues with some platforms/compilers.
Fix cast warning with g++.
Line length cleanups.

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -2512,7 +2512,7 @@ int wolfTPM2_ReadPCR(WOLFTPM2_DEV* dev, int pcrIndex, int hashAlg, byte* digest,
     PCR_Read_Out pcrReadOut;
     int digestLen = 0;
 
-    if (dev == NULL || pcrIndex < PCR_FIRST || pcrIndex > PCR_LAST)
+    if (dev == NULL || pcrIndex < (int)PCR_FIRST || pcrIndex > (int)PCR_LAST)
         return BAD_FUNC_ARG;
 
     /* set session auth to blank */
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -37,6 +37,8 @@ static void wolfTPM2_CopyPubT(TPMT_PUBLIC* out, const TPMT_PUBLIC* in);
 static void wolfTPM2_CopyPub(TPM2B_PUBLIC* out, const TPM2B_PUBLIC* in);
 static void wolfTPM2_CopyPriv(TPM2B_PRIVATE* out, const TPM2B_PRIVATE* in);
 static void wolfTPM2_CopyEccParam(TPM2B_ECC_PARAMETER* out, const TPM2B_ECC_PARAMETER* in);
+static void wolfTPM2_CopyKeyFromBlob(WOLFTPM2_KEY* key, const WOLFTPM2_KEYBLOB* keyBlob);
+static void wolfTPM2_CopyNvPublic(TPMS_NV_PUBLIC* out, const TPMS_NV_PUBLIC* in);
 
 /******************************************************************************/
 /* --- BEGIN Wrapper Device Functions -- */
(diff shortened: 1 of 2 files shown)


tomoveu (commented):
Code looks okay. One concern

Types like TPMT_SYM_DEF and others, have unions inside. So, copying the whole structure guarantees byte alignment and values in the right place. At the same time, it will also lead to copying empty bytes, because sizeof reports the maximum size of the largest member.

In most cases, the TPM will consume this and not complain. In some cases, where it expect value X of size Y, could lead to issues, as we saw when trying to trim down TPM2B_PUBLIC.

Just a note.

I thinking doing XMEMCPY is more safe than having explicit assignments.

tomoveu (approved):
Code looks okay. One concern

Types like TPMT_SYM_DEF and others, have unions inside. So, copying the whole structure guarantees byte alignment and values in the right place. At the same time, it will also lead to copying empty bytes, because sizeof reports the maximum size of the largest member.

In most cases, the TPM will consume this and not complain. In some cases, where it expect value X of size Y, could lead to issues, as we saw when trying to trim down TPM2B_PUBLIC.

Just a note.

I thinking doing XMEMCPY is more safe than having explicit assignments.

dgarske:
Good feedback. I will write a couple utility functions for copying these structures. There are only a few of them.

src/tpm2_wrap.c
```diff
@@ -2873,7 +2919,21 @@ int wolfTPM2_NVReadPublic(WOLFTPM2_DEV* dev, word32 nvIndex,
...
+            if (nvPublic->authPolicy.size > 
+                    (UINT16)sizeof(nvPublic->authPolicy.buffer)) {
+                nvPublic->authPolicy.size = 
+                    (UINT16)sizeof(nvPublic->authPolicy.buffer);
+            }
+            XMEMCPY(nvPublic->authPolicy.buffer,
+                    out.nvPublic.nvPublic.authPolicy.buffer,
+                    nvPublic->authPolicy.size);
+        }
+        nvPublic->dataSize = out.nvPublic.nvPublic.dataSize;
```
tomoveu, line 2934: Should this whole if be another helper function?

src/tpm2_wrap.c
```diff
@@ -1578,11 +1614,16 @@ int wolfTPM2_LoadEccPrivateKey(WOLFTPM2_DEV* dev, const WOLFTPM2_KEY* parentKey,
...
+            (WOLFTPM2_HANDLE*)&parentKey->handle);
     }
 
     /* return loaded key */
-    XMEMCPY(key, &keyBlob, sizeof(WOLFTPM2_KEY));
+    key->handle.hndl = keyBlob.handle.hndl;
+    wolfTPM2_CopyAuth(&key->handle.auth, &keyBlob.handle.auth);
+    wolfTPM2_CopyName(&key->handle.name, &keyBlob.handle.name);
+    wolfTPM2_CopySymmetric(&key->handle.symmetric, &keyBlob.handle.symmetric);
+    wolfTPM2_CopyPub(&key->pub, &keyBlob.pub);
```
tomoveu, line 1626: In favor of readability, does it make sense to have a helper that combines these for copying WOLFTPM2_KEY , KEYBLOB and HANDLE structures?

This way it will again be a single line action and perhaps easier to follow

src/tpm2_wrap.c
```diff
@@ -4666,7 +4917,7 @@ int wolfTPM2_SetCryptoDevCb(WOLFTPM2_DEV* dev, CryptoDevCallbackFunc cb,
         devId = rc;
         tpmCtx->dev = dev;
 
-        rc = wc_CryptoDev_RegisterDevice(devId, cb, tpmCtx);
+        rc = wc_CryptoCb_RegisterDevice(devId, cb, tpmCtx);
```
tomoveu, line 4920: is this required due to wolfCrypt recent library change?

tomoveu (approved):
Code looks good. Very similar and difficult to eyeball a typo. Otherwise, looks solid.

If native and wrapper unit tests work as before, this is a great improvement.

src/tpm2_wrap.c
dgarske: That is the old name that is mapped via macro in wolfCrypt. I am switching to the new name.

src/tpm2_wrap.c
dgarske: I thought about it, but currently only used in one place. I will do anyways.

src/tpm2_wrap.c
dgarske: Sure will cleanup.

tomoveu (approved):
Looks good. Please only re-run the native and wrapper unit tests.

I support you on the crypto callback decision.
There is no need to update wolfCrypt right now.

PR #178, opened 2021-07-02
Fix default prefix

elms:


elms:
@dgarske looks like I missed fully testing the default case. I'm pretty sure the intended behavior was to fail configure unless wolfssl is found or `--disable-wolfcrypt` is specified.

tomoveu:
I think you are correct @elms 

AFAIK, --enable-wolfcrypt is enabled by default, therefore if no wolfssl is found, indeed the configure step fails. Then, the user has the option to use --disable-wolfcrypt, exactly as you say

dgarske (approved):

PR #181, opened 2021-07-21
Fixes for Windows builds

elms:
* Fixes build errors with some tools 
* Fixes build `--disable-wolfcrypt`

### DIFF
--- a/examples/keygen/keygen.c
+++ b/examples/keygen/keygen.c
@@ -369,6 +369,8 @@ int TPM2_Keygen_Example(void* userCtx, int argc, char *argv[])
     #endif
     }
 #else
+    (void)pemFiles;
+    (void)pubFilename;
     printf("Unable to store EK pub as PEM file. Lack of file support\n");
 #endif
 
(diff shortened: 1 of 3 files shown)


dgarske (approved):
Thanks Jeff. Fixes the build errors.

dgarske (changes requested):
Tested with `./configure --enable-winapi --disable-wolfcrypt` and got this warning:

```
examples/keygen/keygen.c: In function ‘TPM2_Keygen_Example’:
examples/keygen/keygen.c:134:17: error: variable ‘pubFilename’ set but not used [-Werror=unused-but-set-variable]
  134 |     const char *pubFilename = NULL;
      |                 ^~~~~~~~~~~
examples/keygen/keygen.c:128:9: error: variable ‘pemFiles’ set but not used [-Werror=unused-but-set-variable]
  128 |     int pemFiles = 0;
      |         ^~~~~~~~
```

Can you patch that one in this PR as well?

Thanks

elms:
Interesting one version I have installed didn't generate those, but cross-compiling on Linux did. Good catch.

dgarske (approved):

PR #187, opened 2021-09-16
RSA_Encrypt: Fix for RSAES padding

elms:
Added test and excluding hash for `RSAES` as well as `NULL`

Ref Forum: https://www.wolfssl.com/forums/topic1810-cryptographic-operations-with-wolftpm.html


dgarske (approved):
Awesome work Jeff. I did some digging and can confirm `TPM_ALG_RSAES` should not have the `hashAlg` added.

Good example for this in the Microsoft Urchin stack here: https://github.com/LordOfDorks/Urchin/blob/master/Lib/Marshal.c#L4017

Screen Shot 2021-09-16 at 9 33 31 AM

PR #191, opened 2021-10-29
Fix for big endian in TIS layer.

dgarske:


elms (approved):
Wasn't able to test on hardware locally due to unrelated issues. Changes look good and confirmed working elsewhere so approving and merging.

PR #192, opened 2021-11-08
SSH support for public key based authentication

dgarske:
Experimental SSH support for public key based authentication using a TPM.


dgarske:
Closing. This PR is not required for SSH.

PR #196, opened 2021-11-18
Fix `make install` by renaming pcr example `read.c`

elms:
example files are installed into the same directory and must have unique names.


dgarske (changes requested):
Can you also add the new `read_pcr` to the .gitignore?

elms:
@dgarske This was the quickest solution, but would like to look at keeping the examples in subdirectories when installed. Any thoughts on that?

dgarske:
The make install for me currently generates:

```
./configure --prefix="`pwd`/myinstall" && make install
% ls -la myinstall/bin
total 528
drwxr-xr-x  8 davidgarske  staff    256 Nov 18 10:24 .
drwxr-xr-x  6 davidgarske  staff    192 Nov 18 10:24 ..
-rwxr-xr-x  1 davidgarske  staff  52840 Nov 18 10:24 keygen
-rwxr-xr-x  1 davidgarske  staff  52120 Nov 18 10:24 keyimport
-rwxr-xr-x  1 davidgarske  staff  51928 Nov 18 10:24 keyload
-rwxr-xr-x  1 davidgarske  staff  51864 Nov 18 10:24 read
-rwxr-xr-x  1 davidgarske  staff  51968 Nov 18 10:24 store
-rwxr-xr-x@ 1 davidgarske  staff   4017 Nov 18 10:24 wolftpm-config
```

I don't think any apps should be installed with `make install`. Can you remove those too?

dgarske (approved):
Awesome! Thanks @elms

PR #200, opened 2022-01-24
Add support for ST33 vendor specific command `TPM_CC_GetRandom2`

dgarske:
Allows getting DRBG data up to `TPM2B_MAX_BUFFER` in size. ZD13545

### DIFF
--- a/examples/native/native_test.c
+++ b/examples/native/native_test.c
@@ -330,22 +330,35 @@ int TPM2_Native_TestArgs(void* userCtx, int argc, char *argv[])
     if (TPM2_GetVendorID() == TPM_VENDOR_STM) {
         XMEMSET(&cmdIn.getRand, 0, sizeof(cmdIn.getRand));
         i = (int)sizeof(cmdOut.getRand2.randomBytes);
+        if (i > (MAX_RESPONSE_SIZE-(int)sizeof(UINT16))) {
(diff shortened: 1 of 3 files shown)


elms (commented):
Looks good. 

I don't see a way to exercise this and test it.

dgarske:
Customer is happy with this. @elms over to you for final. Thanks

dgarske:
I will add a test case to call it for ST33 cases.

examples/native/native_test.c
```diff
@@ -323,15 +326,27 @@ int TPM2_Native_TestArgs(void* userCtx, int argc, char *argv[])
 
 
     /* Random */
-    XMEMSET(&cmdIn.getRand, 0, sizeof(cmdIn.getRand));
-    cmdIn.getRand.bytesRequested = MAX_RNG_REQ_SIZE;
-    rc = TPM2_GetRandom(&cmdIn.getRand, &cmdOut.getRand);
+#if defined(WOLFTPM_ST33) || defined(WOLFTPM_AUTODETECT)
```
elms, line 329: It appears not all ST supports this. For that case if we see `TPM_RC_COMMAND_CODE` I think it should fall back to `TPM2_GetRandom` so this test can still pass on that hardware.

elms (changes requested):

elms (approved):

PR #215, opened 2022-06-28
Add create primary key example

dgarske:
* Add create primary key example.
* Add support for using a unique template with create and create primary.

Confirmed the "unique" string is used for the primary key derivation from seed.
The primary key must be recreated on Windows TBS as it cannot be persisted to NV. The public template (including unique value) is used for the KDF based on hierarchy seed.

### DIFF
--- a/examples/keygen/create_primary.c
+++ b/examples/keygen/create_primary.c
@@ -1,0 +1,248 @@
+/* create_primary.c
+ *
+ * Copyright (C) 2006-2022 wolfSSL Inc.
+ *
+ * This file is part of wolfTPM.
+ *
+ * wolfTPM is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfTPM is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/* Tool and example for creating, storing and loading keys using TPM2.0 */
+
+#include <wolftpm/tpm2_wrap.h>
+
+#include <examples/keygen/keygen.h>
+#include <examples/tpm_io.h>
+#include <examples/tpm_test.h>
+#include <examples/tpm_test_keys.h>
+
+#include <stdio.h>
+#include <stdlib.h> /* atoi */
+
+#ifndef WOLFTPM2_NO_WRAPPER
+
+/******************************************************************************/
+/* --- BEGIN TPM Create Primary Key Example -- */
+/******************************************************************************/
+static void usage(void)
+{
+    printf("Expected usage:\n");
+    printf("./examples/keygen/create_primary [-ecc/-rsa] [-oh/eh/ph/] [-unique=] [-auth=] [-aes/xor] [-store=]\n");
+    printf("Primary Key Type:\n");
+    printf("\t-rsa: Use RSA for asymmetric key generation (DEFAULT)\n");
+    printf("\t-ecc: Use ECC for asymmetric key generation \n");
+    printf("Hierarchy:\n");
+    printf("\t-oh: Create keys under the Owner Hierarchy (DEFAULT)\n");
+    printf("\t-eh: Create keys under the Endorsement Hierarchy\n");
+    printf("\t-ph: Create keys under the Platform Hierarchy\n");
+    printf("Unique Template:\n");
+    printf("\t-unique=[value]\n");
+    printf("\t\tOptional unique value for the KDF of the create\n");
+    printf("Authentication:\n");
+    printf("\t-auth=[value]\n");
+    printf("\t\tOptional authentication string for primary\n");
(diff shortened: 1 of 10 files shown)


examples/keygen/create_primary.c
```diff
@@ -0,0 +1,248 @@
...
+
+#ifndef WOLFTPM2_NO_WRAPPER
+
+/******************************************************************************/
+/* --- BEGIN TPM Create Primary Key Example -- */
+/******************************************************************************/
+static void usage(void)
+{
+    printf("Expected usage:\n");
+    printf("./examples/keygen/create_primary [-ecc/-rsa] [-oh/eh/ph/] [-unique=] [-auth=] [-aes/xor] [-store=]\n");
```
anhu, line 42: `/` after ph should be removed.  

`-` before rsa should be removed for consistency with others.

examples/keygen/create_primary.c
```diff
@@ -0,0 +1,248 @@
...
+    printf("\t\tcreate_primary -rsa -oh -auth=ThisIsMyStorageKeyAuth -store=0x81000200\n");
+}
+
+int TPM2_CreatePrimaryKey_Example(void* userCtx, int argc, char *argv[])
+{
+    int rc;
+    WOLFTPM2_DEV dev;
+    WOLFTPM2_KEY primary;
+    TPMT_PUBLIC publicTemplate;
+    TPMI_ALG_PUBLIC alg = TPM_ALG_RSA; /* default, see usage() for options */
```
anhu, line 74: Comment not needed, but if you want to keep it, add the same thing to `hierarchy` below for consistency.

examples/keygen/create_primary.c
```diff
@@ -0,0 +1,248 @@
...
+    if (argc >= 2) {
+        if (XSTRNCMP(argv[1], "-?", 2) == 0 ||
+            XSTRNCMP(argv[1], "-h", 2) == 0 ||
+            XSTRNCMP(argv[1], "--help", 6) == 0) {
+            usage();
+            return 0;
+        }
+    }
+    while (argc > 1) {
+        if (XSTRNCMP(argv[argc-1], "-rsa", 4) == 0) {
```
anhu, line 91: suggestion: Change use of numbers to strlen() 

ie 4 ---> strlen("-rsa")

examples/keygen/create_primary.c
```diff
@@ -0,0 +1,248 @@
...
+        printf("Storing Primary key to handle 0x%08x\n", persistHandle);
+        rc = wolfTPM2_NVStoreKey(&dev, hierarchy, &primary,
+            persistHandle);
+        if (rc != TPM_RC_SUCCESS) goto exit;
+    #else
+        printf("Windows TBS does not allow persisting handles to NV\n");
+    #endif
+    }
+
+exit:
```
anhu, line 212: if `wolfTPM2_Init()` fails then `dev` , `primary` and `tpmSession`  are uninitialized. Is that okay?

wolftpm/tpm2_wrap.h
```diff
@@ -2160,6 +2160,21 @@ WOLFTPM_API int wolfTPM2_GetKeyTemplate_RSA_AIK(TPMT_PUBLIC* publicTemplate);
...
 
+/*!
+    \ingroup wolfTPM2_Wrappers
+    \brief Sets the unique area of a public template used by Create or CreatePrimary.
+    \return TPM_RC_SUCCESS: successful
+    \return BAD_FUNC_ARG: check the provided arguments
+
+    \param publicTemplate pointer to an empty structure of TPMT_PUBLIC type, to store the new template
+    \param unique optional pointer to buffer to populate unique area of public template. If NULL used the buffer will be zeroized.
+    \param uniqueSz size to fill the unique field. If zero the key size is used.
```
anhu, line 2171: >  If zero the key size is used.

That seems odd.

anhu (changes requested):
Just a few comments. Looks great though!!

examples/keygen/create_primary.c
```diff
@@ -0,0 +1,248 @@
...
+
+    if (persistHandle > 0) {
+    #ifndef WOLFTPM_WINAPI
+        /* Move storage key into persistent NV */
+        printf("Storing Primary key to handle 0x%08x\n", persistHandle);
+        rc = wolfTPM2_NVStoreKey(&dev, hierarchy, &primary,
+            persistHandle);
+        if (rc != TPM_RC_SUCCESS) goto exit;
+    #else
+        printf("Windows TBS does not allow persisting handles to NV\n");
```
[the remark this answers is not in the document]
anhu: *sigh* I was joking. :laughing:

anhu (approved):
Excellent!!! Will merge momentarily.

examples/keygen/create_primary.c
```diff
@@ -0,0 +1,248 @@
...
+#ifdef DEBUG_WOLFTPM
+    printf("Primary Key Public (%d bytes)\n", primary.pub.size);
+    TPM2_PrintBin((const byte*)&primary.pub.publicArea, primary.pub.size);
+#endif
+
+    if (persistHandle > 0) {
+    #ifndef WOLFTPM_WINAPI
+        /* Move storage key into persistent NV */
+        printf("Storing Primary key to handle 0x%08x\n", persistHandle);
+        rc = wolfTPM2_NVStoreKey(&dev, hierarchy, &primary,
```
JasonMac219, line 207: Could the primary key be stored as persistent?

dgarske: > Could the primary key be stored as persistent?

Hi @JasonMac219 , on TPM's where the internal NV is available it is possible. On platforms like Windows TBS there is no way to store the primary key. You would need to recreate it using the same template and unique at startup which will create the same primary key, since it uses the same seed. Once it is created a temporary (memory based) handle is returned, which you can use until next reboot.

Issue #221, opened 2022-07-06
wolftpm C# wrapper needs to implement IDisposable instead of using Finalizer.


zhichao-h:
In the wolfTPM.cs, when cleaning up resources, the C# Finalizer are used instead of following the IDisposable. This can be dangerous as the resources are only cleaned up when the finalizers are run by the GC (hopefully). In many cases, the finalizer may never run as such resource leak will ensure.

The following classes should implement IDisposable:

- KeyBlob
- Key
- Template
- Session
- Device.

dgarske:
Hi @zhichao-h , great suggestion! I will get a PR put up tomorrow fixing the way the classes are finalized.

dgarske:
Fixed in #225. Marking closed.

PR #225, opened 2022-07-12
Implement `IDisposable` for CSharp classes for cleanup of unmanaged resources

dgarske:
Implement `IDisposable` for CSharp classes that need cleanup of unmanaged resources (see #221)


wrapper/CSharp/wolfTPM.cs
```diff
@@ -147,7 +147,7 @@ public enum X509_Format : int
         DER = 2,
     }
```
JasonMac219, line 149: A terrific change, does the WoltTPM2Exception implement this interface?

PR #226, opened 2022-07-15
Improve cleanup logic for create primary

dgarske:


dgarske:
This PR does nothing useful. The key handle will not be unloaded unless `UnloadHandle` is called. Closing.

PR #231, opened 2022-07-20
Improve the byte swapping logic for GCC

dgarske:
Fixes ZD 14539


douzzer (approved):
LGTM.  built locally, and experimentally confirmed the new gating for gcc `__builtin`s is working.

anhu:
Thanks douzzer!!

PR #239, opened 2022-08-25
Fix for CSharp `SetCustomExtension`

anhu:
Tested with the CSharp wrapper tests and inspected the CSR.

```
# wolfSSL
./configure --enable-wolftpm --enable-certgen --enable-asn=template \
    CFLAGS="-DWOLFSSL_CUSTOM_OID -DHAVE_OID_ENCODING"
make && sudo make install

# wolfTPM
./configure --enable-swtpm --enable-debug
make && sudo make install
cd wrapper/CSharp
mcs wolfTPM.cs wolfTPM-tests.cs -r:/usr/lib/cli/nunit.framework-2.6.3/nunit.framework.dll -t:library
LD_LIBRARY_PATH=../../src/.libs/ nunit-console wolfTPM.dll
openssl req -inform pem -in csr_customExt.pem -text
```

Fixes ZD14713


wolftpm/tpm2_wrap.h
```diff
@@ -83,10 +83,25 @@ typedef struct WOLFTPM2_HMAC {
 } WOLFTPM2_HMAC;
 
 #ifdef WOLFTPM2_CERT_GEN
+
 typedef struct WOLFTPM2_CSR {
+#if defined(WOLFSSL_ASN_TEMPLATE) && defined(WOLFSSL_CUSTOM_OID) && \
+    defined(HAVE_OID_ENCODING)
+
+    #ifndef WOLFTPM2_NUM_CUSTOM_EXT
+    #define WOLFTPM2_NUM_CUSTOM_EXT 16
```
dgarske, line 92: Can’t we get the defaults from wolfCrypt?

wolftpm/tpm2_wrap.h
```diff
@@ -83,10 +83,25 @@ typedef struct WOLFTPM2_HMAC {
 } WOLFTPM2_HMAC;
 
 #ifdef WOLFTPM2_CERT_GEN
+
 typedef struct WOLFTPM2_CSR {
+#if defined(WOLFSSL_ASN_TEMPLATE) && defined(WOLFSSL_CUSTOM_OID) && \
```
dgarske, line 88: I prefer these defines to be outside the typedef although I realize it works this way too.

wrapper/CSharp/wolfTPM-tests.cs
```diff
@@ -540,6 +540,8 @@ public void TryGenerateCSRCustomOID()
             Assert.That(rc, Is.GreaterThan(0));
 
             Console.WriteLine("CSR PEM {0} bytes", rc.ToString());
+            var writer = new BinaryWriter(File.OpenWrite("csr.pem"));
```
dgarske, line 543: This could be handy. Any issue running it over and over?

dgarske (changes requested):

wrapper/CSharp/wolfTPM-tests.cs
anhu: Glad you like it.!

If you used over and over then it would probably write all the things to the same file?  I would think the writer needs to fall out of scope and then the garbage collector would then close the file?

I can break this out to its own function.  Then it could fall out of scope and it would make things clearer.

anhu:
Please do not merge.  Still waiting for customer feedback.

dgarske (approved):

anhu:
I have tested @dgarske 's lastest change.  It works as expected and I approve.

dgarske:
Going to go ahead and merge. I believe this fix solves things for the customer.

PR #240, opened 2022-09-01
wolfTPM v2.6.0 Release

dgarske:
* Fix CMakeList.txt to fix include order.
* Remove execute bit on code files.


ChangeLog.md
```diff
@@ -1,5 +1,23 @@
...
 
+## wolfTPM Release 2.6 (09/01/2022)
+
+**Summary**
+
+Fix for CSharp wrapper when setting a custom OID for a CSR. Added CSharp wrapper documentation and improved a few others. Added CSharp function to set key password for blob.
+
+**Detail**
+
+* Fix for CSharp `SetCustomExtension` to use allocated buffer (PR #239)
```
anhu, line 11: The change is actually to use byte instead of string. So this might be better:

 Fix for CSharp `SetCustomExtension` to use `byte` instead of `string` for `oid`

anhu (changes requested):

anhu (approved):

PR #241, opened 2022-09-07
Save OIDs to a persistent buffer. Gated

anhu:
Fixes ZD14745


dgarske (approved):
Looks good. I'll test on customer hardware today. The reason I suggested using `WOLFTPM2_NO_HEAP` is because the CSharp wrapper requires it not defined.

anhu:
I'm good with the Close()

anhu:
Closed.  The bug this was supposed to fix had a different root cause.

PR #243, opened 2022-09-16
Added example for NV counter increment and Fixes for NV auth handling

dgarske:
Re: https://developers.tpm.dev/chats/1689646


haydenroche5:
> Re: [developers.tpm.dev/chats/1689646](https://developers.tpm.dev/chats/1689646)

This link just takes me to the TPM.dev homepage.

examples/nvram/counter.c
```diff
@@ -0,0 +1,200 @@
...
+    int paramEncAlg = TPM_ALG_NULL;
+    word32 nvIndex = TPM2_DEMO_NV_COUNTER_INDEX;
+
+    XMEMSET(&tpmSession, 0, sizeof(tpmSession));
+    XMEMSET(&parent, 0, sizeof(parent));
+    XMEMSET(&nv, 0, sizeof(nv));
+    XMEMSET(&storage, 0, sizeof(storage));
+
+    if (argc >= 2) {
+        if (XSTRNCMP(argv[1], "-?", 2) == 0 ||
```
haydenroche5, line 70: XSTRNCMP not necessary. Just use XSTRCMP. argv guaranteed to be null-terminated strings. Same applies below.

examples/nvram/counter.c
```diff
@@ -0,0 +1,200 @@
...
+            XSTRNCMP(argv[1], "--help", 6) == 0) {
+            usage();
+            return 0;
+        }
+    }
+    while (argc) {
+        if (XSTRNCMP(argv[argc-1], "-aes", 4) == 0) {
+            paramEncAlg = TPM_ALG_CFB;
+        }
+        if (XSTRNCMP(argv[argc-1], "-xor", 4) == 0) {
```
haydenroche5, line 81: if -> else if

examples/nvram/counter.c
```diff
@@ -0,0 +1,200 @@
...
+            paramEncAlg = TPM_ALG_XOR;
+        }
+
+        if (XSTRNCMP(argv[argc-1], "-nvindex=", 8) == 0) {
```
haydenroche5, line 85: if -> else if. Would also add an else to warn about unrecognized options. Or, you can leave as-is (i.e. ignore unrecognized options).

examples/nvram/counter.c
```diff
@@ -0,0 +1,200 @@
...
+            nvIndex = (word32)strtol(argv[argc-1] + 8, NULL, 0);
```
haydenroche5, line 86: Convert arg to signed long then back to unsigned int -- can we just use XATOI? Also, what's the + 8 doing?

examples/nvram/counter.c
```diff
@@ -0,0 +1,200 @@
...
+            if (nvIndex > TPM_20_OWNER_NV_SPACE &&
+                nvIndex < TPM_20_OWNER_NV_SPACE + 0x003FFFFF)
+            {
```
haydenroche5, line 89: Bracket on same line as if.

examples/nvram/counter.c
```diff
@@ -0,0 +1,200 @@
...
+            paramEncAlg = TPM_ALG_CFB;
+        }
+        if (XSTRNCMP(argv[argc-1], "-xor", 4) == 0) {
+            paramEncAlg = TPM_ALG_XOR;
+        }
+
+        if (XSTRNCMP(argv[argc-1], "-nvindex=", 8) == 0) {
+            nvIndex = (word32)strtol(argv[argc-1] + 8, NULL, 0);
+            if (nvIndex > TPM_20_OWNER_NV_SPACE &&
+                nvIndex < TPM_20_OWNER_NV_SPACE + 0x003FFFFF)
```
haydenroche5, line 88: What is the meaning of 0x003FFFFF? Prefer named constant to magic number.

dgarske:
> > Re: [developers.tpm.dev/chats/1689646](https://developers.tpm.dev/chats/1689646)
> 
> This link just takes me to the TPM.dev homepage.

You'd probably need to register. This was in response to some issues with NV increment:

Hi everyone. I am struggling with hybrid NV counters.
We are trying to use a hybrid NV counter to ensure the freshness of frequently changing data on the disk.
We were able to create the hybrid NV counter using `TPM2_NV_DefineSpace` with the following NV index attributes:

```
TPMA_NV_AUTHWRITE |
TPMA_NV_AUTHREAD |
TPMA_NV_NO\_DA |
TPMA_NV_ORDERLY |
TPM_NT_COUNTER
```

yielding a total value of `0x06040014`.

We are supplying a constant password for that NV index.

We are then able to increment the counter and read it using the constant password until the next TPM reset.
After the TPM reset (power-off and power-on), we can still read the counter but no longer increment it using that password. Calling `TPM2_NV_Increment` then results in `TPM_RC_ATTRIBUTES (0xA82)`.
Has anyone a clue why we cannot increment the counter after a reset of the TPM?

> if you do a nv_readpublic on the NV area after the TPM reset, what does the attributes say?
The attributes of the NV index are 0x26040014 after TPM reset. So nothing changed except TPMA_NV_WRITTEN is now set as expected.

The wolftpm library we are using sets TPM2B_AUTH auth field of the TPM2_NV_DefineSpace command to set the password for the NV index.

Thanks for engaging with our problem.

We are using the wrapper `wolfTPM2_NVCreateAuth` to create the NV counter index, but unfortunately, there is no wrapper to increment the counter, so we are using `TPM2_NV_Increment` directly. As a quick hack, we copied and adapted the implementation of  `wolfTPM2_NVWriteAuth`:

```c++
WOLFTPM2_HANDLE handle;
memset(&handle, 0, sizeof(handle));
handle.hndl = TPM2_NV_INDEX_TRUST_ANCHOR_HASH_COUNTER;
memcpy(handle.auth.buffer, "test", 4);
handle.auth.size = 4;

// TODO this check probably is not necessary:
if (_dev.ctx.session) {
    wolfTPM2_SetAuthHandle(&_dev, 0, &handle);
}

TPMS_NV_PUBLIC nv_public;
int result = wolfTPM2_NVReadPublic(&_dev, handle.hndl, &nv_public);
_assert_tpm_result("TPM NV read public:", result);

result = TPM2_HashNvPublic(&nv_public, (byte*)&handle.name.name, &handle.name.size);
_assert_tpm_result("TPM hash NV public:", result);

result = wolfTPM2_SetAuthHandleName(&_dev, 0, &handle);
_assert_tpm_result("TPM set auth handle name:", result);

result = wolfTPM2_SetAuthHandleName(&_dev, 1, &handle);
_assert_tpm_result("TPM set auth handle name:", result);

NV_Increment_In in;
memset(&in, 0, sizeof(in));
in.authHandle = handle.hndl;
in.nvIndex = TPM2_NV_INDEX_TRUST_ANCHOR_HASH_COUNTER;

result = TPM2_NV_Increment(&in);
_assert_tpm_result("TPM increment counter:", result); // This fails
```

src/tpm2_wrap.c
```diff
@@ -3168,22 +3173,14 @@ int wolfTPM2_NVCreateAuth(WOLFTPM2_DEV* dev, WOLFTPM2_HANDLE* parent,
     #endif
         return rc;
     }
+    if (rc == TPM_RC_SUCCESS && alreadyExists)
```
haydenroche5, line 3176: These if's without brackets always scare me a bit. Worry about someone adding a second line to the conditional and not adding brackets. Prefer to add brackets always, but up to you. Same comment applies elsewhere in this PR.

src/tpm2_wrap.c
```diff
@@ -3218,40 +3215,21 @@ int wolfTPM2_NVWriteAuth(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv,
...
-
-    /* Compute NV Index name in case of parameter encryption */
-#ifndef WOLFTPM2_NO_WOLFCRYPT
-    rc = TPM2_HashNvPublic(&nvPublic, (byte*)&nv->handle.name.name,
-                            &nv->handle.name.size);
-    if (rc != TPM_RC_SUCCESS) {
-        return rc;
+    /* make sure the name is computed for the handle */
+    if (!nv->handle.nameLoaded) {
+        wolfTPM2_NVOpen(dev, nv, nvIndex, NULL, 0);
```
haydenroche5, line 3224: Do we need to check the return value?

src/tpm2_wrap.c
```diff
@@ -3401,6 +3360,47 @@ int wolfTPM2_NVRead(WOLFTPM2_DEV* dev, TPM_HANDLE authHandle,
     return wolfTPM2_NVReadAuth(dev, &nv, nvIndex, dataBuf, pDataSz, offset);
 }
 
+int wolfTPM2_NVOpen(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv, word32 nvIndex,
+    const byte* auth, word32 authSz)
+{
+    int rc = TPM_RC_SUCCESS;
+    TPMS_NV_PUBLIC nvPublic;
+
+    if (dev == NULL || nv == NULL || authSz > sizeof(nv->handle.auth.buffer))
```
haydenroche5, line 3369: Need to check if auth is NULL, too, I think.

wolftpm/tpm2_wrap.h
```diff
@@ -1560,6 +1561,13 @@ WOLFTPM_API int wolfTPM2_NVWriteAuth(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv,
 WOLFTPM_API int wolfTPM2_NVReadAuth(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv,
     word32 nvIndex, byte* dataBuf, word32* pDataSz, word32 offset);
 
+
+WOLFTPM_API int wolfTPM2_NVIncrement(WOLFTPM2_DEV* dev, WOLFTPM2_NV* nv);
```
haydenroche5, line 1565: Should we add documentation for these new API functions?

wolftpm/tpm2_wrap.h
```diff
@@ -35,6 +35,7 @@ typedef struct WOLFTPM2_HANDLE {
     TPMT_SYM_DEF    symmetric;
     TPM2B_NAME      name;
     int             policyAuth; /* Handle requires Policy, not password Auth */
+    int             nameLoaded; /* flag to indicate if "name" was loaded and computed */
```
haydenroche5, line 38: Could be a bitfield, but not strictly necessary.

src/tpm2_wrap.c
```diff
@@ -3437,7 +3437,51 @@ int wolfTPM2_NVReadPublic(WOLFTPM2_DEV* dev, word32 nvIndex,
...
+{
+    int rc = TPM_RC_SUCCESS;
+    NV_Increment_In in;
+
+    if (dev == NULL || nv == NULL)
+        return BAD_FUNC_ARG;
+
+    /* make sure the name is computed for the handle */
+    if (!nv->handle.nameLoaded) {
+        wolfTPM2_NVOpen(dev, nv, nv->handle.hndl, NULL, 0);
```
haydenroche5, line 3454: Check return value?

src/tpm2_wrap.c
```diff
@@ -3401,6 +3360,47 @@ int wolfTPM2_NVRead(WOLFTPM2_DEV* dev, TPM_HANDLE authHandle,
...
+        XMEMCPY(nv->handle.auth.buffer, auth, authSz);
+    }
+
+    /* Read the NV Index publicArea to have up to date NV Index Name */
+    rc = wolfTPM2_NVReadPublic(dev, nv->handle.hndl, &nvPublic);
+    if (rc != TPM_RC_SUCCESS) {
+    #ifdef DEBUG_WOLFTPM
+        printf("Failed to open (read) NV\n");
+    #endif
+        return TPM_RC_FAILURE;
```
haydenroche5, line 3386: Should we return the rc instead of converting to TPM_RC_FAILURE? May help with debug to know the actual rc.

src/tpm2_wrap.c
```diff
@@ -3437,7 +3437,51 @@ int wolfTPM2_NVReadPublic(WOLFTPM2_DEV* dev, word32 nvIndex,
...
+
+    /* make sure auth not set */
+    wolfTPM2_UnsetAuth(dev, 1);
+
+    rc  = wolfTPM2_SetAuthHandleName(dev, 0, &nv->handle);
+    if (rc != TPM_RC_SUCCESS) {
+    #ifdef DEBUG_WOLFTPM
+        printf("Setting NV index name failed\n");
+    #endif
+        return TPM_RC_FAILURE;
```
haydenroche5, line 3465: Same comment as earlier: should we return rc?

haydenroche5 (changes requested):

examples/nvram/counter.c
dgarske: strtol can accept base16 hex strings and atoi cannot. Converted to portable macro XSTRTOL()

haydenroche5 (approved):

Issue #109, opened 2020-07-24
Add TPM 2.0 example for seal and unseal (using PCR)


dgarske:
TODO: Add example for sealing / unsealing data with TPM based on PCR value.

HarryR:
The implementation in #244 uses `TPM2_PolicyAuthorize` with `TPM2_PolicyPCR`, however for my use case I needed a direct PCR policy for Unseal.

It would be good to get the PolicyAuthorize technique merged though as iirc that's what BitLocker uses, and WolfTPM seemed to be the sanest and easiest to embed library (for EFI) vs the N-levels of indirection in `tpm2-tss` with lots of dependencies and the scary spaghetti that is the IBM libraries.

Couldn't figure out how to get your branch of WolfTPM to do the unseal the way I needed, but hopefully this information will be useful to you if you choose to implement it.

I ended up just pushing the bytes to/from the TPM by hand as I didn't need flexibility or integration headaches, although I got stuck on how to calculate the HMAC for Unseal with a Policy session. Reading the TCG specs sent me down a week-long rabbit hole, and eventually found it easier to enable `TSS2_LOG=all+TRACE` rather than relying on the spec or the TSS code.

Attached is a sample TPM conversation performing an Unseal of a PCR policy locked key, the calls to `PCR_Read` and `GetCapability` are unnecessary, iirc it should also be unnecessary to pass `pcrDigest` via `PolicyPCR`. [log.unseal2.txt](https://github.com/wolfSSL/wolfTPM/files/10932795/log.unseal2.txt)

Snippet to calculate the HMAC (see the log.unseal2.txt file for where the hashes come from):

```python
from hashlib import sha256
import hmac

# Constant
CC_Unseal = bytes.fromhex('0000015e')

# Returned from TPM_CC.ReadPublic
obj_name = bytes.fromhex('000bafb549e7ce3f1c929e4446f529f6999104eaceb66adedddd0ac8528bd815621d')

context = sha256(CC_Unseal + obj_name).digest()

# Submitted with Unseal request as authorizationArea[0].nonce.buffer
unseal_nonce = bytes.fromhex('0f38a8ce6b7397b81db627a8616c89521d2feb2d1b65f8f546338307d1309b6e')

# Returned from StartAuthSession as .parameters.nonceTPM.buffer
nonceTPM = bytes.fromhex('9ba1856fa01a2559b55ff813a9063f2e56d934a3131af768f869940406
bfab73')

# Constant
[...]

PR #244, opened 2022-09-28
Support for PCR with policy

dgarske:
Issue #109

### DIFF
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -187,8 +187,9 @@ static int TPM2_CommandProcess(TPM2_CTX* ctx, TPM2_Packet* packet,
         /* Note: Copy between TPM2_AUTH_SESSION and TPMS_AUTH_COMMAND is allowed */
         XMEMCPY(&authCmd, session, sizeof(TPMS_AUTH_COMMAND));
 
-        /* Skip Policy session, because Enhanced Authorization is not yet implemented */
-        if (TPM2_IS_HMAC_SESSION(session->sessionHandle)) {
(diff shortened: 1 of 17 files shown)


src/tpm2.c
```diff
@@ -5540,10 +5541,20 @@ int TPM2_GetName(TPM2_CTX* ctx, UINT32 handleValue, int handleCnt, int idx, TPM2
...
 {
     if (pcr && pcrIndex >= (int)PCR_FIRST && pcrIndex <= (int)PCR_LAST) {
-        pcr->count = 1;
-        pcr->pcrSelections[0].hash = alg;
-        pcr->pcrSelections[0].sizeofSelect = PCR_SELECT_MIN;
-        pcr->pcrSelections[0].pcrSelect[pcrIndex >> 3] = (1 << (pcrIndex & 0x7));
+        pcr->pcrSelections[pcr->count].hash = alg;
+        pcr->pcrSelections[pcr->count].sizeofSelect = PCR_SELECT_MIN;
+        pcr->pcrSelections[pcr->count].pcrSelect[pcrIndex >> 3] = (1 << (pcrIndex & 0x7));
+        pcr->count++;
```
HarryR, line 5547: It's not necessary to increase the PCR selection count for every PCR selected.

However, one `pcrSelection` is necessary per hash algorithm.

Suggest only increasing `pcr->count` if there is not an existing `pcrSelection` with the same `.hash == alg`.

dgarske:
Replaced by #261

PR #206, opened 2022-05-06
Fixes for c++ compiler

elms:
* remove use of reserved c++ keyword `template`
 * remove use of struct designator initialization (sort of supported in c++20, but easier to just remove)

PR #246, opened 2022-09-28
Fix for variable declarations to be at top of functions

dgarske:


src/tpm2.c
```diff
@@ -2177,9 +2196,10 @@ TPM_RC TPM2_Hash(Hash_In* in, Hash_Out* out)
 
     rc = TPM2_AcquireLock(ctx);
     if (rc == TPM_RC_SUCCESS) {
-        CmdInfo_t info = {0,0,0,0,};
-        info.flags = (CMD_FLAG_ENC2 | CMD_FLAG_DEC2);
         TPM2_Packet packet;
+        CmdInfo_t info = {0,0,0,0};
+        info.flags = (CMD_FLAG_ENC2 | CMD_FLAG_DEC2);
```
haydenroche5, line 2201: Not critical, but here and elsewhere, why don't we just set the fields in the init of info? If we're going to assign them right after init to 0, might as well not init to 0 in the first place.

haydenroche5 (approved):
Feel free to address my comment if you want, but I'm ok with merging as-is. Just let me know.

src/tpm2.c
dgarske: This was changed to resolve a compiler issue: https://github.com/wolfSSL/wolfTPM/pull/206
Reverting my last commit.

haydenroche5 (approved):

PR #252, opened 2023-01-11
add rsa key import methods to handle pem and der encoding directly

jpbland1:
ZD 14996


examples/keygen/keyimport.c
```diff
@@ -33,16 +33,47 @@
 
 #ifndef WOLFTPM2_NO_WRAPPER
 
+static const char* kRsaKeyPrivPem =
```
dgarske, line 36: Please add a comment about where this RSA private key came from. Like `./certs/server-key.pem`?

examples/keygen/keyimport.c
```diff
@@ -55,6 +86,8 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
     TPM_ALG_ID paramEncAlg = TPM_ALG_NULL;
     WOLFTPM2_SESSION tpmSession;
     const char* outputFile = "keyblob.bin";
+    uint8_t derEncode = 0;
```
dgarske, line 89: I like stdint.h types, but please use `byte`. Thanks

examples/keygen/keyimport.c
```diff
@@ -121,12 +163,24 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
...
-            kRsaKeyPrivQ,      (word32)sizeof(kRsaKeyPrivQ),
-            TPM_ALG_NULL, TPM_ALG_NULL);
+        if (derEncode == 1) {
+            rc = wolfTPM2_RsaPrivateKeyImportDer(&dev, &storage, &impKey,
+                kRsaKeyPrivDer, sizeof(kRsaKeyPrivDer), TPM_ALG_NULL,
+                TPM_ALG_NULL);
+        }
+        else if (pemEncode == 1) {
+            rc = wolfTPM2_RsaPrivateKeyImportPem(&dev, &storage, &impKey,
+                kRsaKeyPrivPem, strlen(kRsaKeyPrivPem), NULL,
```
dgarske, line 173: Please use `(word32)XSTRLEN(kRsaKeyPrivPem)`.

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,76 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+    byte d[RSA_MAX_SIZE / 8];
+    byte p[RSA_MAX_SIZE / 8];
+    byte q[RSA_MAX_SIZE / 8];
+    word32  eSz = sizeof(e);
+    word32  nSz = sizeof(n);
+    word32  dSz = sizeof(d);
+    word32  pSz = sizeof(p);
+    word32  qSz = sizeof(q);
+
+    if (dev && parentKey && keyBlob && input && inSz != 0)
```
dgarske, line 2081: Confusing logic. Would prefer init `rc = 0` and set `rc = BAD_FUNC_ARG;` on failure.

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,76 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+{
+    int rc = BAD_FUNC_ARG;
+    RsaKey key[1];
+    word32 idx = 0;
+    word32  e;
+    byte n[RSA_MAX_SIZE / 8];
+    byte d[RSA_MAX_SIZE / 8];
+    byte p[RSA_MAX_SIZE / 8];
+    byte q[RSA_MAX_SIZE / 8];
+    word32  eSz = sizeof(e);
```
dgarske, line 2075: Must cast these: `word32 eSz = (word32)sizeof(e);`

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,76 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+    const char* input, word32 inSz, char* pass,
+    TPMI_ALG_RSA_SCHEME scheme, TPMI_ALG_HASH hashAlg)
+{
+    int rc = 0;
+    byte* derBuf;
+    word32 derSz;
+
+    if (dev == NULL || parentKey == NULL || keyBlob == NULL || input == NULL ||
+            inSz == 0) {
+        rc = 0;
```
dgarske, line 2109: Should be `return BAD_FUNC_ARG`.

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,76 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+            &qSz);
+
+    if (rc == 0)
+        rc = wolfTPM2_ImportRsaPrivateKey(dev, parentKey, keyBlob, n, nSz, e, q,
+            qSz, scheme, hashAlg);
+
+    return rc;
+}
+
+int wolfTPM2_RsaPrivateKeyImportPem(WOLFTPM2_DEV* dev,
```
dgarske, line 2098: Please wrap with `#ifndef WOLFTPM2_NO_HEAP`

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,76 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+
+    if (dev == NULL || parentKey == NULL || keyBlob == NULL || input == NULL ||
+            inSz == 0) {
+        rc = 0;
+    }
+
+    /* der size is base 64 decode length */
+    derSz = inSz * 3 / 4 + 1;
+
+    derBuf = XMALLOC(derSz, NULL, DYNAMIC_TYPE_TMP_BUFFER);
```
dgarske, line 2115: `derBuf = (byte*)XMALLOC`

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,76 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+
+    if (derBuf == NULL)
+        return MEMORY_E;
+
+    if (rc == 0)
+        rc = wc_KeyPemToDer((byte*)input, inSz, derBuf, derSz, pass);
+
+    /* returns the number of bytes*/
```
dgarske, line 2123: Nit: space between `bytes*/`

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,76 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+    if (rc > 0)
```
dgarske, line 2124: What about rc == 0 case?

examples/keygen/keyimport.c
```diff
@@ -77,6 +110,15 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
         else if (XSTRCMP(argv[argc-1], "-xor") == 0) {
             paramEncAlg = TPM_ALG_XOR;
         }
+        else if (XSTRCMP(argv[argc-1], "-pem") == 0) {
```
dgarske, line 113: `-pem` is in here twice. Looks like copy/paste issue. This one should be deleted.

dgarske (changes requested):

src/tpm2_wrap.c
jpbland1: saw it in some wolfssl function and thought it was concise, if it's not preferred I'll change it

src/tpm2_wrap.c
jpbland1: yes copy paste error, meant to return before the malloc happened

src/tpm2_wrap.c
jpbland1: can you have a der key with 0 byte length?I thought that would return bad arg error, I looked at the underlying function `PemToDer` and if it doesn't find an asn header it returns `ASN_NO_PEM_HEADER` which is less than 0

src/tpm2_wrap.c
jpbland1: do you want me to use macros to make a a stack only version? I didn't want to do this because the pem can vary in length and it would have to be a huge buffer by default

examples/keygen/keyimport.c
jpbland1: so is it only wolfBoot that uses uint types?

examples/keygen/keyimport.c
jpbland1: I moved it to a file instead, I was just copying the way the der key was included raw

certs/example-rsa-key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
dgarske, line 1: Please add to `certs/include.am`

examples/keygen/keyimport.c
```diff
@@ -86,8 +56,10 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
     TPM_ALG_ID paramEncAlg = TPM_ALG_NULL;
     WOLFTPM2_SESSION tpmSession;
     const char* outputFile = "keyblob.bin";
-    uint8_t derEncode = 0;
-    uint8_t pemEncode = 0;
+    byte derEncode = 0;
+    byte pemEncode = 0;
+    FILE* pemFile = NULL;
+    char pemBuf[2048];
```
dgarske, line 62: Just use `WOLFTPM2_MAX_BUFFER`

examples/keygen/keyimport.c
```diff
@@ -169,9 +138,16 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
                 TPM_ALG_NULL);
         }
         else if (pemEncode == 1) {
-            rc = wolfTPM2_RsaPrivateKeyImportPem(&dev, &storage, &impKey,
-                kRsaKeyPrivPem, strlen(kRsaKeyPrivPem), NULL,
-                TPM_ALG_NULL, TPM_ALG_NULL);
+            pemFile = fopen("./certs/example-rsa-key.pem", "r");
```
dgarske, line 141: Please use `XFOPEN`

examples/keygen/keyimport.c
```diff
@@ -169,9 +138,16 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
...
+
+            rc = fread(pemBuf, 1, 2048, pemFile);
```
dgarske, line 143: Please use `XFREAD` and use `sizeof(pemBuf)`

examples/keygen/keyimport.c
```diff
@@ -169,9 +138,16 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
...
+
+            if (rc > 0) {
+                rc = wolfTPM2_RsaPrivateKeyImportPem(&dev, &storage, &impKey,
+                    pemBuf, rc, NULL, TPM_ALG_NULL, TPM_ALG_NULL);
+            }
+
+            fclose(pemFile);
```
dgarske, line 150: `XFCLOSE`

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,83 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+
+    /* der size is base 64 decode length */
+    derSz = inSz * 3 / 4 + 1;
+
+    derBuf = (byte*)XMALLOC(derSz, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+
+    if (derBuf == NULL)
+        return MEMORY_E;
+
+    if (rc == 0)
```
dgarske, line 2124: This `rc == 0` is not needed. Remove please

examples/keygen/keyimport.c
```diff
@@ -55,6 +56,10 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
     TPM_ALG_ID paramEncAlg = TPM_ALG_NULL;
     WOLFTPM2_SESSION tpmSession;
     const char* outputFile = "keyblob.bin";
+    byte derEncode = 0;
```
dgarske, line 59: Wrap all this new PEM logic with `#if !defined(NO_FILESYSTEM) && !defined(NO_WRITE_TEMP_FILES)`

examples/keygen/keyimport.c
dgarske: Add support to take the key filename as argument. Default to this example.

dgarske (changes requested):

examples/keygen/keyimport.c
```diff
@@ -67,20 +79,38 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
...
+#if !defined(NO_FILESYSTEM) && !defined(NO_WRITE_TEMP_FILES)
+        else if (XSTRCMP(argv[i], "-pem") == 0) {
+            pemEncode = 1;
+        }
+        else if (XSTRCMP(argv[i], "-pem-file") == 0) {
+            if (i + 1 >= argc) {
+                printf("Warning: No pem file specified, using default: %s\n", pemName);
+            }
+            else {
+                XMEMCPY(pemName, argv[i + 1], XSTRLEN(argv[i + 1]) + 1);
```
dgarske, line 105: Please print the name of the file used in the console output.

examples/keygen/keyimport.c
```diff
@@ -67,20 +79,38 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
...
             paramEncAlg = TPM_ALG_CFB;
         }
-        else if (XSTRCMP(argv[argc-1], "-xor") == 0) {
+        else if (XSTRCMP(argv[i], "-xor") == 0) {
             paramEncAlg = TPM_ALG_XOR;
         }
+        else if (XSTRCMP(argv[i], "-der") == 0) {
+            derEncode = 1;
+        }
+#if !defined(NO_FILESYSTEM) && !defined(NO_WRITE_TEMP_FILES)
```
dgarske, line 96: The `-pem` option should still be available with no filesystem.

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,82 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
 
 #ifndef WOLFTPM2_NO_WOLFCRYPT
 #ifndef NO_RSA
+int wolfTPM2_RsaPrivateKeyImportDer(WOLFTPM2_DEV* dev,
+    const WOLFTPM2_KEY* parentKey, WOLFTPM2_KEYBLOB* keyBlob, const byte* input,
+    word32 inSz, TPMI_ALG_RSA_SCHEME scheme, TPMI_ALG_HASH hashAlg)
+{
+    int rc = 0;
+    RsaKey key[1];
```
dgarske, line 2068: You are missing the RSA init and free.

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,82 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+    if (dev == NULL || parentKey == NULL || keyBlob == NULL || input == NULL ||
+            inSz == 0) {
+        return BAD_FUNC_ARG;
+    }
+
+    /* der size is base 64 decode length */
+    derSz = inSz * 3 / 4 + 1;
+
+    derBuf = (byte*)XMALLOC(derSz, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+
```
dgarske, line 2120: Remove the extra line here.

dgarske (changes requested):

examples/keygen/keyimport.c
```diff
@@ -121,12 +151,37 @@ int TPM2_Keyimport_Example(void* userCtx, int argc, char *argv[])
...
+        else if (pemEncode == 1) {
+            pemFile = XFOPEN(pemName, "r");
+
+            if (pemFile == NULL)
+                printf("Failed to read pem file %s\n", pemName);
+
+            if (rc == 0)
+                rc = XFREAD(pemBuf, 1, sizeof(pemBuf), pemFile);
```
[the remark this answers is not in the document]
jpbland1: how do I test this, I need to set CC to clang?

src/tpm2_wrap.c
jpbland1: I thought we didn't need that since we're not doing any operations with the key, we're basically just using the `mp_int` members of the key and those are all static

jpbland1: It's more complicated, I don't need to init because decode will call init on the the `mp_int` members but it's it's too complicated to tell if the `mp_int` members alloc or not so I'll free just to be safe

jpbland1: nevermind, the free call has problems if init wasn't called so I'll just do it

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,87 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+    byte d[RSA_MAX_SIZE / 8];
+    byte p[RSA_MAX_SIZE / 8];
+    byte q[RSA_MAX_SIZE / 8];
+    word32  eSz = (word32)sizeof(e);
+    word32  nSz = (word32)sizeof(n);
+    word32  dSz = (word32)sizeof(d);
+    word32  pSz = (word32)sizeof(p);
+    word32  qSz = (word32)sizeof(q);
+
+    wc_InitRsaKey(key, NULL);
```
dgarske, line 2081: Please check the return code here. Only call free if `rc == 0`.

src/tpm2_wrap.c
```diff
@@ -2059,6 +2060,87 @@ int wolfTPM2_ReadPublicKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+    derSz = inSz * 3 / 4 + 1;
+
+    derBuf = (byte*)XMALLOC(derSz, NULL, DYNAMIC_TYPE_TMP_BUFFER);
+    if (derBuf == NULL)
+        return MEMORY_E;
+
+    rc = wc_KeyPemToDer((byte*)input, inSz, derBuf, derSz, pass);
+
+    /* returns the number of bytes */
+    if (rc > 0) {
```
dgarske, line 2132: Perhaps the `rc == 0` case needs to be handled too?

dgarske (changes requested):
Very close! Looks great, just a couple comments.

dgarske (approved):

PR #254, opened 2023-01-17
Fix for using type not available with wolfCrypt only

dgarske:
Resolves error with `WOLFSSL_FILETYPE_PEM` not being available with `WOLFCRYPT_ONLY`.


jpbland1 (approved):
looks good, didn't find any missing uses of `WOLFSSL_FILETYPE_`

PR #255, opened 2023-01-27
fix casting and unused variable problems on windows

jpbland1:
ZD 14996

### DIFF
--- a/examples/tpm_test.h
+++ b/examples/tpm_test.h
@@ -103,8 +103,9 @@ static const char pemFileKey[] = "key.pem";
         extern double current_time(int reset);
         return current_time(reset);
     #elif defined(_WIN32)
+        unsigned long long ticks = GetTickCount64();
         (void)reset;
-        return (0.0 + GetTickCount64())/1000.0;
+        return ((double)ticks)/1000.0;
     #else
         struct timeval tv;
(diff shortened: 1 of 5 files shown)


examples/tpm_test.h
```diff
@@ -103,7 +103,8 @@ static const char pemFileKey[] = "key.pem";
         extern double current_time(int reset);
         return current_time(reset);
     #elif defined(_WIN32)
-        return ((double)GetTickCount64())/1000.0;
+        (void)reset;
+        return (0.0 + GetTickCount64())/1000.0;
```
dgarske, line 107: What was the original error/warning? I don't like the 0.0 workaround. Can you try using the following? `return ((double)GetTickCount64())/1000.0F;`

dgarske (changes requested):

examples/tpm_test.h
jpbland1: it doesn't like that we're casting the return value of a function, the error was -Werror=bad-function-cast, let me see if it complains if I put it in a variable first

jpbland1: yeah compiler is fine with it as long as it's stored in a variable fist, should be good now

dgarske (approved):

PR #257, opened 2023-02-10
fix the TPM_E_COMMAND_BLOCKED macro to have the correct value

jpbland1:
according to the microsoft documentation

ZD 15643

### DIFF
--- a/wolftpm/tpm2.h
+++ b/wolftpm/tpm2.h
@@ -1667,7 +1667,7 @@ struct wolfTPM_winContext {
 };
 /* may be needed with msys */
 #ifndef TPM_E_COMMAND_BLOCKED
-#define TPM_E_COMMAND_BLOCKED (0x80284000)
+#define TPM_E_COMMAND_BLOCKED (0x80280400)
 #endif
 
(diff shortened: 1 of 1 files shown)


dgarske (approved):

wolftpm/tpm2.h
```diff
@@ -1667,7 +1667,7 @@ struct wolfTPM_winContext {
 };
 /* may be needed with msys */
 #ifndef TPM_E_COMMAND_BLOCKED
-#define TPM_E_COMMAND_BLOCKED (0x80284000)
+#define TPM_E_COMMAND_BLOCKED (0x80280400)
```
dgarske, line 1670: If you search the code there are two other places in tpm2_wrap.c the "wrong" value is used. `0x80280400`

dgarske (changes requested):

wolftpm/tpm2.h
jpbland1: I don't understand, when I grep for the old value nothing comes up
```
john@john-XPS-9320:~/Documents/wolfTPM$ grep -R "0x80284000"
john@john-XPS-9320:~/Documents/wolfTPM$ 
```
`0x80280400` is the right value according to this https://learn.microsoft.com/en-us/windows/win32/com/com-error-codes-6

dgarske (approved):

PR #264, opened 2023-04-12
Fix for building with wolfTPM without wolfCrypt `--disable-wolfcrypt`

dgarske:
Added test.


jpbland1 (approved):
Tested locally looks good

PR #267, opened 2023-04-17
add external nvram policy sealing example, fix wolfTPM2_SealWithAuthS…

jpbland1:
…igNV

wolfTPM2_SealWithAuthSigNV needs to have PolicyPCR called as a part of its logic since it uses wolfTPM2_SetAuthPassword, which interferes with the policy digest

### DIFF
--- a/examples/nvram/seal_policy_auth_nv_external.c
+++ b/examples/nvram/seal_policy_auth_nv_external.c
@@ -1,4 +1,4 @@
-/* seal_policy_auth_nv.c
+/* seal_policy_auth_nv_external.c
  *
  * Copyright (C) 2006-2023 wolfSSL Inc.
  *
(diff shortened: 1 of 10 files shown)


examples/nvram/seal_policy_auth_nv_external.c
```diff
@@ -0,0 +1,340 @@
...
+    printf("Expected usage:\n");
+    printf("./examples/pcr/policy [-aes/xor] [pcr]\n");
+    printf("* pcr: PCR index between 0-23 (default %d)\n", 16);
+    printf("* -aes/xor: Use Parameter Encryption\n");
+}
+
+static const word32 sealNvIndex = TPM2_DEMO_NV_TEST_INDEX;
+static const word32 policyDigestNvIndex = TPM2_DEMO_NV_TEST_INDEX + 1;
+
+int TPM2_PCR_Seal_With_Policy_Auth_NV_External_Test(void* userCtx, int argc, char *argv[])
```
dgarske, line 58: Add note this is hard-coded only for SECP256R1 (ECC 256-bit).

dgarske (changes requested):

dgarske (approved):

PR #268, opened 2023-04-18
Corrected Linux usage of cs_change, added config overrides.

billphipps:
The Linux IO subsystem's usage of cs_change within the SPI transfer ioctl() call has been corrected.  Since all transactions are single messages (last transactions), the cs_change must be set to 1 when a sequence of transactions must be issued while CS is asserted.  Only the last transaction should have the cs_change set back to 0 to ensure the cs line is desasserted.  Specifically, the fsl-dspi Linux kernel driver does not have a method to dessert CS under error conditions.  Hence, code was added to always complete all SPI transaction sequences with an additional byte transfer with cs_change=0 to ensure the kernel driver correctly reasserts cs.  This code will only be triggered in error handling cases where the SPI transactions have either failed to be issued or timeout for wait states is exceeded.

This code was tested on a Raspberry Pi with a Nuvoton SPI TPM and an LS1028A connected to an Optiga SLB9672.  Both configurations were run with and without auto detection as well as with and without wait state checking.  

Additionally, logic was added to allow the spidev path and cs to be overridden at configuration time by setting the TPM2_SPI_DEV_PATH and TPM_SPI_DEV_CS macros to valid strings such as:
```./configure CPPFLAGS="-DTPM2_SPI_DEV_PATH=\"/dev/spidev0.\" -DTPM2_SPI_DEV_CS=\"0\" "```

[pi_novuton_wraptest.txt](https://github.com/wolfSSL/wolfTPM/files/11266193/pi_novuton_wraptest.txt)
[ls1028a_optiga_wraptest.txt](https://github.com/wolfSSL/wolfTPM/files/11266194/ls1028a_optiga_wraptest.txt)

### DIFF
--- a/hal/tpm_io_linux.c
+++ b/hal/tpm_io_linux.c
@@ -80,7 +80,7 @@
                 #define TPM2_SPI_DEV_CS "0"
             #else
                 /* OPTIGA SLB9670/SLB9762 and LetsTrust TPM use CE1 */
-                #define TPM2_SPI_DEV_CS "0"
+                #define TPM2_SPI_DEV_CS "1"
             #endif
         #endif
 
@@ -91,7 +91,7 @@
         #ifdef WOLFTPM_AUTODETECT
             #undef TPM2_SPI_DEV
             /* this will try incrementing spidev chip selects */
-            static char TPM2_SPI_DEV[] = TPM2_SPI_DEV_PATH TPM2_SPI_DEV_CS;
+            static char TPM2_SPI_DEV[] = TPM2_SPI_DEV_PATH "0";
             #define MAX_SPI_DEV_CS '4'
             static int foundSpiDev = 0;
         #else
--- a/hal/tpm_io_linux.c
+++ b/hal/tpm_io_linux.c
@@ -68,28 +68,34 @@
         #define TPM2_I2C_HZ   400000 /* 400kHz */
     #else
         /* SPI */
-        #ifdef WOLFTPM_MICROCHIP
-            /* Microchip ATTPM20 uses CE0 */
-            #define TPM2_SPI_DEV_CS "0"
-        #elif defined(WOLFTPM_ST33)
-            /* STM ST33HTPH SPI uses CE0 */
-            #define TPM2_SPI_DEV_CS "0"
-        #elif defined(WOLFTPM_NUVOTON)
-            /* Nuvoton NPCT75x uses CE0 */
-            #define TPM2_SPI_DEV_CS "0"
-        #else
-            /* OPTIGA SLB9670/SLB9762 and LetsTrust TPM use CE1 */
-            #define TPM2_SPI_DEV_CS "1"
+        #ifndef TPM2_SPI_DEV_CS
+            #ifdef WOLFTPM_MICROCHIP
+                /* Microchip ATTPM20 uses CE0 */
+                #define TPM2_SPI_DEV_CS "0"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: hal/tpm_io_linux.c [hunk 1]
```diff
@@ -68,28 +68,34 @@
...
+
+        #ifndef TPM2_SPI_DEV_PATH
+            #define TPM2_SPI_DEV_PATH "/dev/spidev0."
         #endif
 
         #ifdef WOLFTPM_AUTODETECT
             #undef TPM2_SPI_DEV
             /* this will try incrementing spidev chip selects */
-            static char TPM2_SPI_DEV[] = "/dev/spidev0.0";
+            static char TPM2_SPI_DEV[] = TPM2_SPI_DEV_PATH TPM2_SPI_DEV_CS;
```
dgarske, line 94: This auto-detect logic should always be 0.0 to 0.4.

hal/tpm_io_linux.c
```diff
@@ -68,28 +68,34 @@
...
+                #define TPM2_SPI_DEV_CS "0"
+            #elif defined(WOLFTPM_ST33)
+                /* STM ST33HTPH SPI uses CE0 */
+                #define TPM2_SPI_DEV_CS "0"
+            #elif defined(WOLFTPM_NUVOTON)
+                /* Nuvoton NPCT75x uses CE0 */
+                #define TPM2_SPI_DEV_CS "0"
+            #else
+                /* OPTIGA SLB9670/SLB9762 and LetsTrust TPM use CE1 */
+                #define TPM2_SPI_DEV_CS "0"
```
dgarske, line 83: Why change this? The comment still says CE1, but it uses CE0. This case is when we know it is an Infineon, this is not the auto-detect case.

dgarske (changes requested):

hal/tpm_io_linux.c
billphipps: Got it.  I thought I was being smart by using the previously provided CS.  Much better to force this back to 0 to 4.  I'll update.

hal/tpm_io_linux.c
billphipps: Since the autodetect will be updated to always be 0 to 4, there's no need for this change either.  I'll revert.

hal/tpm_io_linux.c
```diff
@@ -226,73 +236,88 @@
...
+        else {
+            /* Failed to open device */
+            ret = TPM_RC_FAILURE;
+        }
 
     #ifdef WOLFTPM_AUTODETECT
         /* if response is not 0xFF then we "found" something */
         if (!foundSpiDev) {
-            if (ret == TPM_RC_SUCCESS && rxBuf[0] != 0xFF) {
+            if ((ret == TPM_RC_SUCCESS) && (rxBuf[TPM_TIS_HEADER_SZ-1] != 0xFF)) {
```
jpbland1, line 313: > 80 characters, also bonus parenthesis clutter things up, I think your only supposed to add them when you're doing bit wise operations

dgarske (approved):

hal/tpm_io_linux.c
billphipps: Oops,  was gonna put that on 2 lines, so I added parenthesis to make it clearer.  Easier to do what you recommend.

jpbland1 (approved):

dgarske (approved):

PR #270, opened 2023-05-23
add wolfSSL_Init to csr example, required for fips ecc

jpbland1:
ZD 16126

### DIFF
--- a/examples/csr/csr.c
+++ b/examples/csr/csr.c
@@ -166,6 +166,13 @@ int TPM2_CSR_ExampleArgs(void* userCtx, int argc, char *argv[])
         argc--;
     }
 
+    /* init wolfssl, required for fips */
+    rc = wolfSSL_Init();
+    if (rc != SSL_SUCCESS) {
+        printf("wolfSSL_Init failed %d\n", rc);
+        return rc;
+    }
+
     /* Init the TPM2 device */
     rc = wolfTPM2_Init(&dev, TPM2_IoCb, userCtx);
     if (rc != 0) {
--- a/examples/csr/csr.c
+++ b/examples/csr/csr.c
@@ -167,16 +167,29 @@ int TPM2_CSR_ExampleArgs(void* userCtx, int argc, char *argv[])
     }
 
     /* init wolfssl, required for fips */
+#ifdef WOLFCRYPT_ONLY
+    rc = wolfCrypt_Init();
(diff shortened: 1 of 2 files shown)


examples/csr/csr.c
```diff
@@ -166,6 +166,13 @@ int TPM2_CSR_ExampleArgs(void* userCtx, int argc, char *argv[])
         argc--;
     }
 
+    /* init wolfssl, required for fips */
+    rc = wolfSSL_Init();
```
dgarske, line 170: This is great, but a couple things.
1) In WOLFCRYPT_ONLY this needs to be `wolfCrypt_Init()`.
2) Use the wolfSSL values `WOLFSSL_SUCCESS`.
3) If you are going to call init you also need to call `_Cleanup()`...

dgarske (changes requested):

examples/csr/csr.c
```diff
@@ -166,9 +166,30 @@ int TPM2_CSR_ExampleArgs(void* userCtx, int argc, char *argv[])
         argc--;
     }
 
+    /* init wolfssl, required for fips */
+#ifdef WOLFCRYPT_ONLY
```
dgarske, line 170: This logic should be in `wolfTPM2_Init` and `wolfTPM2_Cleanup`? I am pretty sure these init/cleanups are already called. Perhaps the issue is I only call the wolfCrypt_Init version?

dgarske (changes requested):

examples/csr/csr.c
dgarske: Please swap out the logic you added in csr.c, but put into tpm2.c and add the `#ifdef WOLFCRYPT_ONLY` logic. Call wolfSSL_Init by default unless WOLFCRYPT_ONLY is defined.

jpbland1: https://github.com/wolfSSL/wolfssl/blob/master/src/ssl.c#L6421
I think this is why, I'm going to call `wolfSSL_Init` when not `WOLFCRYPT_ONLY` and call `wolfCrypt_Init` and `wc_SetSeed_Cb` when it is

jpbland1:
So you don't need to call wolfSSL_Init, I just needed to add the `wc_SetSeed_Cb` call if `WC_RNG_SEED_CB` is defined

dgarske (approved):

PR #275, opened 2023-07-05
Fix for TPM2 create with decrypt or restricted flag set

dgarske:
Fix for TPM2 create with decrypt or restricted flag set (must use symmetric algorithm).
ZD 16249

Reproducer was hitting failure in simulator here: https://github.com/kgoldman/ibmswtpm2/blob/master/src/Object_spt.c#L530


jpbland1 (approved):
Looks good

PR #276, opened 2023-07-06
Example for Secure Boot solution to store root of trust in NV

dgarske:
* TPM based toot of trust for secure boot. Provides authentication and tamper protection.
* Fixed uses of `arg=` in examples 
* Add support for encrypting secret using ECC key. Allows using ECC for parameter encryption and importing ECC keys with custom seed. Requires https://github.com/wolfSSL/wolfssl/pull/6683
* Add support for NV lock.
* Cleanup wrapper function order/groups.
* Added `wolfTPM2_ChangePlatformAuth` wrapper to help set the platform auth. This is useful from the bootloader to make sure no one can use the platform hierarchy from application.
* Sanitize the IO TX/RX buffers (make sure they are zero initialized).
* Fixes for building with `NO_HMAC`.
* Update wolfSSL test certs.
* Added TPM_TIS_MAX_WAIT.
* Fix build with WOLFTPM_DEBUG_VERBOSE only.


jpbland1 (approved):
Reviewed and tested, confirmed that the index is locked by running a second time to see the write fail to overwrite the locked index

PR #277, opened 2023-07-07
Make sure the file IO XFEOF and XREWIND are available

dgarske:
These are not defined in wolfSSL


night1rider:
Was able to apply patch to allow wolftpm v2.7.0 to work with wolfssl v5.6.3

night1rider (approved):

PR #281, opened 2023-07-19
Fixes for TPM private key import with custom seed

dgarske:
* Fix for import of private key with seed. 
* Added new API `wolfTPM2_ImportPrivateKeyBuffer` for importing private key ECC/RSA in either PEM or DER(ASN.1). 

ZD 16249

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -1051,14 +1051,13 @@ int wolfTPM2_EncryptSalt(WOLFTPM2_DEV* dev, const WOLFTPM2_KEY* tpmKey,
             rc = NOT_COMPILED_IN;
             break;
     }
-#else
-    (void)salt;
-    (void)encSalt;
-    (void)label;
     rc = NOT_COMPILED_IN;
 #endif /* !WOLFTPM2_NO_WOLFCRYPT */
 
     (void)dev;
+    (void)salt;
+    (void)encSalt;
+    (void)label;
 
     return rc;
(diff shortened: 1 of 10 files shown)


src/tpm2_wrap.c
```diff
@@ -1052,23 +1044,20 @@ int wolfTPM2_EncryptSalt(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* tpmKey,
...
     #endif
         default:
             rc = NOT_COMPILED_IN;
             break;
     }
-#else
-    (void)dev;
-    (void)in;
-    (void)salt;
     rc = NOT_COMPILED_IN;
```
jpbland1, line 1054: `wolfTPM2_EncryptSalt` is always returning `NOT_COMPILED_IN`, I think this `rc = NOT_COMPILED_IN` is from the old code and you just missed deleting this line

jpbland1 (changes requested):
Check my comment on `wolfTPM2_EncryptSalt`

src/tpm2_wrap.c
jpbland1: actually now that I look at it why was that else macro removed? if `WOLFTPM2_NO_WOLFCRYPT` is defined rc will be undefined before return, you could probably change it to `int rc = NOT_COMPILED_IN;` at the start of the function

jpbland1 (approved):
Compiled and ran

PR #286, opened 2023-07-28
Added TPM simulator support for UART transport

dgarske:
Enabled with `--enable-swtpm=uart`.


dgarske:
Replaced with https://github.com/wolfSSL/wolfTPM/pull/301

PR #289, opened 2023-08-10
Secure Boot (Root of Trust) - Part 2

dgarske:
* Fix for wolfTPM2_EncryptSecret to not loose error code.
* Avoid using heap (malloc/free) for ecc_point.

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -1044,6 +1044,9 @@ exit:
     return ret;
 }
 
+#ifdef ALT_ECC_SIZE
+#error use of ecc_point below does not support ALT_ECC_SIZE
+#endif
 /* returns both the plaintext and encrypted value */
 /* ECC: data = derived symmetric key
  *      secret = exported public point */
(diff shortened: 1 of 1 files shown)


src/tpm2_wrap.c
```diff
@@ -1063,6 +1068,9 @@ static int wolfTPM2_EncryptSecret_ECC(WOLFTPM2_DEV* dev, const WOLFTPM2_KEY* tpm
     XMEMSET(&eccKeyPriv, 0, sizeof(eccKeyPriv));
     XMEMSET(&pubPoint, 0, sizeof(pubPoint));
     XMEMSET(&secretPoint, 0, sizeof(secretPoint));
+    XMEMSET(&r, 0, sizeof(r));
```
jpbland1, line 1071: `r` is already pointer type, does de-referencing here point XMEMSET to the buffer or a pointer to the buffer?

jpbland1 (changes requested):

src/tpm2_wrap.c
dgarske: Good catch (this was r directly, then I changed to avoid altering code from r. to r->. I think the behavior is the same. Point to start of array vs pointer to first element in array (same). Pushed fix.

jpbland1 (approved):
Reviewed and re-ran the rot example

PR #290, opened 2023-08-15
Support for importing and loading public ECC/RSA keys formatted as PEM or DER

dgarske:


jpbland1 (changes requested):
This test requires CFLAGS='-DWOLFSSL_PUBLIC_MP' for wolfSSL right? should add
```
#if !defined(WOLFTPM2_NO_WRAPPER) && !defined(WOLFTPM2_NO_WOLFCRYPT) && defined(WOLFSSL_PUBLIC_MP)
```
to main, couldn't tag it in the review

dgarske:
> This test requires CFLAGS='-DWOLFSSL_PUBLIC_MP' for wolfSSL right? should add
> 
> ```
> #if !defined(WOLFTPM2_NO_WRAPPER) && !defined(WOLFTPM2_NO_WOLFCRYPT) && defined(WOLFSSL_PUBLIC_MP)
> ```
> 
> to main, couldn't tag it in the review

Yes and that change has been upstreamed into wolfSSL master with --enable-wolftpm. The `WOLFSSL_PUBLIC_MP` is required to expose some of the math functions for the ECC encrypt. Am I missing some documentation?

jpbland1:
No I'm saying to give the user the not compiled in error if their wolfSSL is compiled wrong, like this:
```
int main(int argc, char *argv[])
{
    int rc = NOT_COMPILED_IN;

#if !defined(WOLFTPM2_NO_WRAPPER) && !defined(WOLFTPM2_NO_WOLFCRYPT)
    rc = TPM2_Keyimport_Example(NULL, argc, argv);
#else
    printf("KeyImport code not compiled in\n");
    (void)argc;
    (void)argv;
#endif

    return rc;
}
```
but add the `WOLFSSL_PUBLIC_MP` to the macros

jpbland1 (approved):
tested with known bad wolfssl build, rsa woked correctly and ecc correctly printed the not compiled in error

PR #293, opened 2023-08-18
Handle unknown `TPM2_GetCapability` type and fix bad printf

dgarske:
Fixed bad printf. Don't error on unknown `TPM2_GetCapability` type.


jpbland1 (approved):
Reviewed and Tested

PR #294, opened 2023-08-25
wolfTPM Support for sealing/unsealing based on a PCR that is signed externally

dgarske:
Use an external key to sign a PCR digest. Allows a new signed policy to be sent with updates to continue allowing a sealed secret to be unsealed when PCR's change. This resolves the issue with PCR brittleness.


src/tpm2_wrap.c
```diff
@@ -7029,837 +7029,323 @@ int wolfTPM2_GetPolicyDigest(TPM_HANDLE sessionHandle, byte* policyDigest,
...
-
-    /* get the digest */
-    if (rc == 0)
-        rc = wolfTPM2_HashFinish(dev, hash, checkDigest, &checkDigestSz);
-
-    /* sign the hashed policyDigest so we can authorize it */
-    if (rc == 0) {
-        if (authKey->pub.publicArea.type == TPM_ALG_ECC) {
-            sigAlg = authKey->pub.publicArea.parameters.eccDetail.scheme.scheme;
+    rc = TPM2_GetHashType(pcrAlg);
```
jpbland1, line 7107: why set rc if you're not going to check it before overwriting?

src/tpm2_wrap.c
```diff
@@ -7029,837 +7029,323 @@ int wolfTPM2_GetPolicyDigest(TPM_HANDLE sessionHandle, byte* policyDigest,
...
-        rc = TPM2_PolicySigned(policySignedIn, policySignedOut);
-    }
-
-    /* get the policyDigest */
-    if (rc == 0) {
-        rc = wolfTPM2_GetPolicyDigest(session->handle.hndl, policyDigest,
-            &policyDigestSz);
-    }
+    inSz = *digestSz; /* capture input digest size (for policyDigestOld) */
+    rc = TPM2_GetHashType(pcrAlg);
```
jpbland1, line 7164: why set rc if you're going to overwrite it before checking?

src/tpm2_wrap.c
```diff
@@ -7029,837 +7029,323 @@ int wolfTPM2_GetPolicyDigest(TPM_HANDLE sessionHandle, byte* policyDigest,
...
+        rc = wc_HashUpdate(&hash_ctx, hashType, digest, inSz);
     }
-
-    /* write the policy digest to the nv space */
-    if (rc == 0) {
-        rc = wolfTPM2_NVWriteAuth(dev, nv, policyDigestNvIndex, policyDigest,
-            policyDigestSz, sizeof(TPM_ALG_ID));
+    /* policyRef */
+    if (rc == 0 && policyRefSz > 0) {
+        rc = wc_HashUpdate(&hash_ctx, hashType, digest, inSz);
```
jpbland1, line 7255: should hash policyRef, not digest

src/tpm2_wrap.c
```diff
@@ -7029,837 +7029,323 @@ int wolfTPM2_GetPolicyDigest(TPM_HANDLE sessionHandle, byte* policyDigest,
...
-        policyAuthNVIn->policySession = session->handle.hndl;
-
-        /* set nvIndex */
-        policyAuthNVIn->nvIndex = policyDigestNvIndex;
-
-        /* set authHandle */
-        policyAuthNVIn->authHandle = parent->hndl;
-
-        rc = TPM2_PolicyAuthorizeNV(policyAuthNVIn);
+        rc = wc_HashFinal(&hash_ctx, hashType, digest);
```
jpbland1, line 7258: this will overwrite the contents of digest, might not be wise if the caller needs the buffer's contents

src/tpm2_wrap.c
```diff
@@ -7029,837 +7029,323 @@ int wolfTPM2_GetPolicyDigest(TPM_HANDLE sessionHandle, byte* policyDigest,
...
-        TPM2_SetSessionAuth(dev->session);
-
-        /* set nvIndex */
-        policyAuthNVIn->nvIndex = policyDigestNvIndex;
-
-        /* set authHandle */
-        policyAuthNVIn->authHandle = parent->hndl;
-
-        rc = TPM2_PolicyAuthorizeNV(policyAuthNVIn);
+        rc = wc_HashFinal(&hash_ctx, hashType, digest);
```
jpbland1, line 7325: this also overwrites digest, this might be okay since it's the new policyDigest anyway right?

jpbland1 (changes requested):
general review, found a few issues

examples/boot/secret_seal.c
```diff
@@ -0,0 +1,310 @@
...
+    }
+    printf("wolfTPM2_Init: success\n");
+
+    /* get SRK */
+    rc = getPrimaryStoragekey(&dev, &storage, srkAlg);
+    if (rc != 0) goto exit;
+
+    /* Start an authenticated session (salted / unbound) */
+    rc = wolfTPM2_StartSession(&dev, &tpmSession, &storage, NULL,
+        TPM_SE_HMAC, paramEncAlg);
```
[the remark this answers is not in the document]
dgarske: Creating a sealed object does not require a policy session. We have a signed PCR policy digest we write when creating the keyed hashed (seal) object. No policy session is needed.

examples/boot/secret_seal.c
```diff
@@ -0,0 +1,310 @@
...
+    if (rc != 0) goto exit;
+    printf("Session Handle 0x%x\n", (word32)tpmSession.handle.hndl);
+    printf("Parameter Encryption: %s\n", TPM2_GetAlgName(paramEncAlg));
+
+    rc = wolfTPM2_SetAuthSession(&dev, 1, &tpmSession,
```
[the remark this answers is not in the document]
dgarske: It will not enable parameter encryption if index 0 is used.

examples/boot/secret_unseal.c
```diff
@@ -0,0 +1,372 @@
...
+    printf("* -ecc/-rsa: Public key is RSA or ECC (default is RSA)\n");
+    printf("* -publickey=file: Public key file (PEM or DER) for the policy signing key used\n");
+    printf("Examples:\n");
+    printf("./examples/boot/secret_seal -policy=policyauth.bin -out=sealblob.bin\n");
+
+}
+
+/* Load Key Public Info */
+#if !defined(NO_FILESYSTEM)
+static int LoadAuthKeyInfo(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* authKey,
```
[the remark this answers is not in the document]
dgarske: All the code for this is already in wolfTPM2_ImportPublicKeyBuffer. This is just to load a file, which I don't want in the library proper.

src/tpm2_wrap.c
dgarske: I designed it this way. The "digest" are in both input and output (as documented). This is because some policies take in an "old" digest to get a new one (chaining them).

src/tpm2_wrap.c
dgarske: Haven't found a better way to solve this. Same pattern is used elsewhere. Resolves a compiler warning about int to enum.

src/tpm2_wrap.c
jpbland1: just realized you cast it to hashType, whoops

src/tpm2_wrap.c
jpbland1: sorry I didn't see hashType was being set to it, which is fine. I thought it was being overwritten by 7163

jpbland1 (approved):
Looks good, CI running all the examples is awesome

PR #299, opened 2023-09-15
Improvements/fixes to NV auth and session auth set/unset

dgarske:
* Fix bug with NV name after first write (only appears when using HMAC session).
* Add new API `wolfTPM2_UnsetAuthSession` to unset auth index for a session and save off the nonce from the TPM. This allows auth to be unset/set again with the same session.
* Cleanup in the NV API's for unsetting of the auth to be handled by caller, not in API.

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -752,6 +752,26 @@ int wolfTPM2_UnsetAuth(WOLFTPM2_DEV* dev, int index)
     return TPM2_SetSessionAuth(dev->session);
 }
 
+int wolfTPM2_UnsetAuthSession(WOLFTPM2_DEV* dev, int index,
+    WOLFTPM2_SESSION* tpmSession)
+{
+    TPM2_AUTH_SESSION* devSession;
+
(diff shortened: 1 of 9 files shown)


src/tpm2_wrap.c
```diff
@@ -752,6 +752,26 @@ int wolfTPM2_UnsetAuth(WOLFTPM2_DEV* dev, int index)
...
+
+    if (dev == NULL || tpmSession == NULL ||
+            index >= MAX_SESSION_NUM || index < 0) {
+        return BAD_FUNC_ARG;
+    }
+
+    devSession = &dev->session[index];
+
+    /* save off nonce from TPM to support continued use of session */
+    memcpy(&tpmSession->nonceTPM, &devSession->nonceTPM, sizeof(TPM2B_NONCE));
```
jpbland1, line 768: XMEMCPY?

jpbland1 (changes requested):

jpbland1 (approved):
Looks good

PR #300, opened 2023-09-22
Various build fixes

dgarske:
* Fix for building without public MP.


jpbland1 (approved):
Tested and it looks good

PR #301, opened 2023-10-04
addition of uart and uartns550 to swtpm

JacobBarthelmeh:


dgarske:
For there UART transport it seems odd to still use the swtpm feature. Is there a reason you didn't set it up as its own transport like SPI/I2C as I did in this PR https://github.com/wolfSSL/wolfTPM/pull/286 ?

JacobBarthelmeh:
Not sure I fully understand, is this pulling the uartns550 additions out of swtpm and placing it in its own location. Possibly something under hal/tpm_io_uartns550 or the like? Which re-looking at it may be a better fit for it. No reason for it to be in the swtpm code other than familiarity with that section of code when doing the work, happy to move it to a better spot.

dgarske:
> Not sure I fully understand, is this pulling the uartns550 additions out of swtpm and placing it in its own location. Possibly something under hal/tpm_io_uartns550 or the like? Which re-looking at it may be a better fit for it. No reason for it to be in the swtpm code other than familiarity with that section of code when doing the work, happy to move it to a better spot.

I see... you are still using the SWPTM packet framing (locality, TX size, packet, RX size, etc..) but you are adding UART transport instead of socket.
For some reason I was environing this solution looking more like a physical TPM, not SWTPM.
Keep this as-is for now.

src/tpm2.c
```diff
@@ -34,7 +34,13 @@
 /* --- Local Variables -- */
 /******************************************************************************/
 
-static THREAD_LS_T TPM2_CTX* gActiveTPM;
+#if defined(WOLFTPM_SWTPM_UARTNS550)
```
dgarske, line 37: Would prefer a different build option to make this a non-static non-thread local. Maybe `WOLFTPM2_PUBLIC_CTX`?

dgarske (changes requested):
Looks great and tested it out. Just one minor suggestion.

gojimmypi:
Hi @dgarske and @JacobBarthelmeh - just wondering what's the status of this PR?

I see there's a useful [DEBUG_PRINTF](https://github.com/JacobBarthelmeh/wolfTPM/blob/fc5ff50a799c8ac5154303748694e5c471ae4449/src/tpm2.c#L78) proposed. I almost when down a similar path, but ended up changing my mind.

I have a similar change in the works, but using the [ERROR_OUT(rc, exit)](https://github.com/gojimmypi/wolfTPM/blob/48a2effe5e0283a779f5db7356f647e47defe366/examples/native/native_test.c#L213) pattern found in the [wolfCrypt test](https://github.com/wolfSSL/wolfssl/blob/cb689104d1aaf274639ff45b7dd69d6160b82d50/wolfcrypt/test/test.c#L766).

This is all in relation to my https://github.com/wolfSSL/wolfTPM/pull/351 adding Espressif TPM support, and in particular the ability to use the colorization of messages, such as the red `ESP_LOGE()` messages.

Any preference here?

JacobBarthelmeh:
@gojimmypi it's been blocked on me. I went back and rebased it but after the rebase it was not working like I expected. Have not resolved that yet.

Yeah the abstraction of printf's saved a large amount of data when trying to debug in a resource constrained device that had a light version of printf.

My preference is not having ERROR_OUT and goto's.....that a preference though.

gojimmypi:
@JacobBarthelmeh I hear ya on the `goto`'s. I too, have mixed feeling about using them. It can be acceptable if used properly.

@dgarske has quite a few instances of the [goto exit](https://github.com/wolfSSL/wolfTPM/blob/e5da5560a744646c262d44a080492966099b0f7f/examples/native/native_test.c#L215) pattern, which of course lend themselves to easily adapting to `ERROR_OUT`.

In any case, I'll hold off on that for now.

dgarske:
The --enable-swtpm=uart support is added in https://github.com/wolfSSL/wolfTPM/pull/474

PR #303, opened 2023-10-04
Fixes to avoid struct assignment and C++ build fixes

dgarske:
Which can result is unaligned access (use memcpy). ZD 16788
Also fixes for building with C++.


jpbland1 (approved):
Looks good

PR #304, opened 2023-10-04
Relocate crypto callback code to its own code file

dgarske:


jpbland1 (approved):
Tested working with tls client, might be worth making a CI test for that

PR #307, opened 2023-10-18
Fixes and improvements for wolfTPM CSR wrappers

dgarske:
* Fix for trying to use a custom wolfTPM CSR `sigType`. The `csr->req.sigType` was being initialized to CTC_SHA256wRSA, and not allowing override.
* Add wrapper support for setting key usage (not just extended key usage).
* Fix support for ECC 384-bit only support. Tested with: `./configure --enable-wolftpm CFLAGS="-DECC_USER_CURVES -DNO_ECC256 -DHAVE_ECC384" --disable-examples --disable-crypttests && make`
* Add CI tests for CSharp wrappers.

### DIFF
--- a/.github/workflows/make-test-swtpm.yml
+++ b/.github/workflows/make-test-swtpm.yml
@@ -52,6 +52,25 @@ jobs:
       run: |
         make check
         WOLFSSL_PATH=./wolfssl ./examples/run_examples.sh
+    - name: make install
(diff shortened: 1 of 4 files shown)


dgarske:
I am going to add a test case for this.

jpbland1 (approved):
Looks good

PR #311, opened 2023-11-13
Add TPM crypto callback support for RSA key generation

dgarske:
Required for https://github.com/wolfSSL/wolfPKCS11/pull/23

* Add TPM crypto callback support for RSA key generation
* Allow import of wolf ECC marked as private only (`ECC_PRIVATEKEY_ONLY`).
* Properly translate a TPM ECC signature verify error for compatibility.
* Fixes for building wolfCrypt without PEM to DER support.
* Improve the ECC key import scheme for signing.
* Improve logic for finding TPM curve in ECC key generation. A call to wc_ecc_make_key can use curve_id 0 (to detect), but we can get it from the "dp".
* Fix to make sure leading ECC sign leading zeros are removed when not required.
* Fix leading zero issue on verify.
* Fix for ECC encrypt secret integrity check failed due to zero pad issue.
* Fix for policy_sign issue when r or s is less than key size (needs zero padding).
* Support ECC KeyGen for signing or derive based on callback context eccKey or ecdhKey population.
* Cleanup KDF function return code checking to avoid scan-build warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ Mfg NTC (0), Vendor NPCT75x"!!4rls, Fw 7.2 (131072), FIPS 140-2 1, CC-EAL4 0
 git clone https://github.com/wolfSSL/wolfssl.git
 cd wolfssl
 ./autogen.sh
-./configure --enable-certgen --enable-certreq --enable-certext --enable-pkcs7 --enable-cryptocb --enable-aescfb
+./configure --enable-wolftpm
 make
 sudo make install
 sudo ldconfig
--- a/examples/pcr/policy_sign.c
+++ b/examples/pcr/policy_sign.c
@@ -109,7 +109,7 @@ static int PolicySign(TPM_ALG_ID alg, const char* keyFile, const char* password,
     if (rc == 0) {
         /* handle PEM conversion to DER */
         if (encType == ENCODING_TYPE_PEM) {
-        #if !defined(WOLFTPM2_NO_HEAP) && defined(WOLFSSL_PEM_TO_DER)
+        #ifdef WOLFTPM2_PEM_DECODE
             /* der size is base 64 decode length */
             word32 derSz = (word32)bufSz * 3 / 4 + 1;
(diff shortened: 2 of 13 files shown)


dgarske:
> It should be documented that `WOLFSSL_PUBLIC_MP` is required for all the examples in `examples/run_examples.sh` to succeed. Any reason not to add it to the recommended wolfSSL config in the readme?
> 
> Also seeing a failure in the csr test:
> 
> ```
> :~/git/wolfTPM $ ./examples/csr/csr -cert
> TPM2 CSR Example
> Trying TPM @ /dev/spidev0.0 (33 MHz)
> Trying TPM @ /dev/spidev0.1 (33 MHz)
> Found TPM @ /dev/spidev0.1
> TPM2: Caps 0x30000697, Did 0x001b, Vid 0x15d1, Rid 0x16
> TPM2_Startup pass
> TPM2_SelfTest pass
> TPM2_ReadPublic Handle 0x81000200: pub 282, name 34, qualifiedName 34
> Loading SRK: Storage 0x81000200 (282 bytes)
> Reading 506 bytes from rsa_test_blob.raw
> Public Area (size 278):
>   Type: RSA (0x1), name: SHA256 (0xB), objAttr: 0x60460, authPolicy sz: 0
>   RSA: sym algorithm: NULL (0x10), sym keyBits: 0, sym mode: Unknown (0x0)
>        scheme: NULL (0x10), scheme hash: Unknown (0x0)
>        keyBits: 2048, exponent: 0x10001, unique size 256
> Reading the private part of the key
> TPM2_Load Key Handle 0x80000001
> Loaded key to 0x80000001
> TPM2_FlushContext: Closed handle 0x80000001
> Failure 0xffffff39: Random Number Generator failed
> ```

That is correct. The ECC encrypt scheme requires direct mp_ math functions. It has already been added to wolfSSL's `--enable-wolftpm`. See PR https://github.com/wolfSSL/wolfssl/pull/6683

embhorn (approved):
Tested with RPi / Infineon TPM

PR #316, opened 2023-12-06
Added Windows Visual Studio project for wolfTPM

dgarske:


embhorn (approved):
I like the idea of building wolfssl in the TPM solution!

PR #322, opened 2024-01-17
Added PKCS7 ECC support to example

dgarske:
* Added PKCS7 ECC support to example.
* Added wrapper function to export TPM public key as DER/ASN.1 or PEM.
* Fix for crypto callback ECC sign to handle getting keySz for unknown cases (like PKCS7 without privateKey set).
* Fixed keygen and csr to use ECC SRK.

Fixes ZD 16675

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -110,10 +110,14 @@ The script creates the following X.509 files (also in .pem format):
 
 Example signs and verifies data with PKCS #7 using a TPM based key.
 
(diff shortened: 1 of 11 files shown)


JacobBarthelmeh:
Tested the PKCS7 example out on an Infineon TPM + PI2. Code looks good, thanks David.

JacobBarthelmeh (approved):

PR #323, opened 2024-01-19
Fix build error with missing `wc_RsaKeyToPublicDer_ex`

dgarske:
Broken in #322


JacobBarthelmeh:
Please also double check the macro guards around the ECC export above. I believe it needs

`#if defined(HAVE_ECC) && defined(HAVE_ECC_KEY_EXPORT) `

JacobBarthelmeh (approved):

PR #341, opened 2024-04-05
Fix TLS examples to not use openssl compatibility macros

night1rider:
Fixes issues building wolfSSL with provider or engine enabled.

### DIFF
--- a/examples/tls/tls_client.c
+++ b/examples/tls/tls_client.c
@@ -594,7 +594,7 @@ exit:
     }
 
     /* Bidirectional shutdown */
-    while (wolfSSL_shutdown(ssl) == SSL_SHUTDOWN_NOT_DONE) {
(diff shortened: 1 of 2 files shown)


examples/tls/tls_client.c
```diff
@@ -48,6 +48,11 @@
     #include <wolfssl/certs_test.h>
 #endif
 
+#ifdef OPENSSL_COEXIST
+    #undef SSL_SHUTDOWN_NOT_DONE
+    #define SSL_SHUTDOWN_NOT_DONE WOLFSSL_SHUTDOWN_NOT_DONE
```
dgarske, line 53: No, just switch to using `WOLFSSL_SHUTDOWN_NOT_DONE`. None of this code is needed.

dgarske (changes requested):
For both

dgarske (approved):
Thank you

PR #366, opened 2024-07-31
Various improvements for `/dev/tpm0`

### LINKED ISSUE #358: Support of tpmrmX interface
Hi,

Are there any plans to support tpm resource manager integration(tpmrm0)?

Ildar

### LINKED ISSUE #365: tpm2_linux open() file descriptor 0 should be allowed
Hi,
I'm using older version of WolfTPM but see this part of code is same in newest version.

What my program does is intense usage of TPM2_GetRandom from few threads in parallel, for couple of minutes.

Observation is that in some cases /dev/tpm0 cannot be opened with `errno=Success` that caused then cascade of `errno=Device or resource busy`.

As from [Linux man](https://man7.org/linux/man-pages/man2/open.2.html) for open(0 can be seen that open() valid return value is nonnegative integer, thus 0 is valid value.
```
On success, open(), openat(), and creat() return the new file
       descriptor (a nonnegative integer).  On error, -1 is returned and
       [errno](https://man7.org/linux/man-pages/man3/errno.3.html) is set to indicate the error.
```

Suggest to change a condition fro `>` to `>=` to:
```c
    fd = open(TPM2_LINUX_DEV, O_RDWR | O_NONBLOCK);
    if (fd >= 0) {
        /* Send the TPM command */
```

After fix, I see no more issues with my code using wolfTPM library.

dgarske:
* Fix for /dev/tpm0 file descriptor check (fixes #365). 
* Added documentation for /dev/tpm0 permissions (fixes #358). 
* Various spelling and documentation cleanups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,8 @@ cd wolfTPM
 make
 ```
 
+The default is SLB9672/SLB9673 (if I2C). To specify SLB9670 use `--enable-infineon=slb9670`.
+
 ### Building ST ST33
 
 Build wolfTPM:
@@ -268,9 +270,9 @@ idf.py build
 
 ### Building for "/dev/tpmX"
 
-This build option allows you to talk to any TPM vendor supported by the Linux TIS kernel driver
+The `--enable-devtpm` or `WOLFTPM_LINUX_DEV` build option allows you to use the Linux supplied TPM (TIS) driver.
 
-Build wolfTPM:
+To specify a different `/dev/tpmX` device use `CFLAGS="-DTPM2_LINUX_DEV=/dev/tpm1"`
 
 ```bash
 ./autogen.sh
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -268,21 +270,40 @@ idf.py build
...
 ```bash
-sudo adduser yourusername tss
+sudo addgroup wolftpm
+sudo adduser [username] wolftpm
+sudo chgrp wolftpm /dev/tpm0
+```
+
+2) Create new rule file: `sudo vim /etc/udev/rules.d/wolftpm-udev.rules`
+
+3) Add the following replacing "yourusername" with actual user or group.
```
embhorn, line 299: I do not see "yourusername" in the command below

embhorn (approved):

PR #389, opened 2024-12-10
Fix issue with session binding / add NV extend example

dgarske:
* Fixed issue with session binding.
* Fixed issue with `wolfTPM2_PolicyHash` where input digest could be too large.
* Added example for NV extend based on the TCG "TCG_-CPU_-TPM_Bus_Protection_Guidance_Active_Attack_Mitigations-V1-R30_PUB-1.pdf".
ZD 19076


embhorn (approved):
Tested examples successfully on RPi with Infineon TPM

PR #390, opened 2024-12-13
Support for building wolfTPM against older wolfCrypt (like v4.7.0)

dgarske:


embhorn (approved):
Did not run test, but LGTM

PR #391, opened 2024-12-13
Create separate tool for performing the TPM2_Clear

dgarske:
Create separate tool for performing the TPM2_Clear (don't use args to wrap_test).


embhorn (approved):
Tested successfully with RPi and Infineon TPM

PR #394, opened 2025-01-17
EK Cert Verification (no wolfCrypt)

dgarske:
New verify_ek_cert is tested with ST33KTPM2X RSA EK Certificate only.


tmael (approved):
Thanks @dgarske! Looks good. I will consider cleaning/reusing cert_asn.c in subsequent pull requests.

PR #397, opened 2025-01-24
zephyr.yml CI github test for zephyr wolftpm

aidangarske:
test the wolftpm_wrap_caps and wolftpm_wrap_test based off the wolfssl workflow.


aidangarske:
need to add to wolftpm zephyr pr for it to actually run and test that it works...

PR #406, opened 2025-03-06
Self Signed CA always set to version 1

night1rider:
Intial work on the bug, I don't think this is the right fix for the ` wolfTPM2_NewCSR(void)` function as `csr->req.selfSigned` is always set to `1` via the init call


dgarske (changes requested):
Is there a version macro in wolfCrypt asn.h that we can use? In tpm2_types.h we have logic to add macros when no wolfCrypt is defined.

dgarske (approved):

PR #407, opened 2025-03-06
Various: Cleanup for public area copy and added cmake provisioning option

dgarske:
Cleanup to replace memcpy with public area copy function. 
Add CMake provisioning option (`WOLFTPM_PROVISIONING`) to match configure.


embhorn (approved):
Looks good to me

PR #408, opened 2025-03-27
correct ASN.1 cert parsing

tmael:
This PR fixes an issue in ASN.1 certificate parsing that was introduced during a refactor with the assistance of Davin AI.

### DIFF
--- a/src/tpm2_asn.c
+++ b/src/tpm2_asn.c
@@ -157,7 +157,7 @@ int TPM2_ASN_DecodeX509Cert(uint8_t* input, int inputSz,
     }
 
     /* Store certificate location */
-    if (rc == 0) {
+    if (rc >= 0) {
         x509->certBegin = idx;
         x509->cert = &input[idx];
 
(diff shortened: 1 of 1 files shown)


src/tpm2_asn.c
```diff
@@ -157,7 +157,7 @@ int TPM2_ASN_DecodeX509Cert(uint8_t* input, int inputSz,
     }
 
     /* Store certificate location */
-    if (rc == 0) {
+    if (rc >= 0) {
```
dgarske, line 160: Why not just change `TPM2_ASN_DecodeTag`, `TPM2_ASN_GetHeader` and `TPM2_ASN_GetLength_ex` to return 0 on success. The length is already returned in arguments...? I would much prefer fixing the return code of these functions then refactor things for >= and not even using rc.

dgarske (changes requested):

src/tpm2_asn.c
dgarske: Tesfa let me know its because we are reusing the code from asn.c as much as possible. Allowing this.

dgarske (approved):

PR #409, opened 2025-03-28
Keygen Optional Authentication Password -auth=<yourpassword>

aidangarske:
Added keygen optional password authentication. 
Added keygen testing for:
- custom passwor
- AIK custom password

### DIFF
--- a/examples/keygen/keygen.c
+++ b/examples/keygen/keygen.c
@@ -43,7 +43,7 @@
 static void usage(void)
 {
     printf("Expected usage:\n");
-    printf("./examples/keygen/keygen [keyblob.bin] [-ecc/-rsa/-sym] [-t] [-aes/xor] [-eh] [-pem]\n");
+    printf("./examples/keygen/keygen [keyblob.bin] [-ecc/-rsa/-sym] [-t] [-aes/xor] [-eh] [-pem] [-auth=pass]\n");
     printf("* -pem: Store the primary and child public keys as PEM formatted files\n");
     printf("\t child public key filename: ak.pem or key.pem\n");
     printf("\t primary public key filename: ek.pem or srk.pem\n");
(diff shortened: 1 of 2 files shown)


examples/keygen/keygen.c
```diff
@@ -57,6 +57,8 @@ static void usage(void)
     printf("* -aes/xor: Use Parameter Encryption\n");
     printf("* -unique=[value]\n");
     printf("\t* Used for the KDF of the create\n");
+    printf("* -auth=pass: Use custom password for key authentication\n");
+    printf("\t* If not specified, no password is used\n");
```
dgarske, line 61: This comment is wrong. It uses the default testing passwords from tpm_test_keys.h.

dgarske (changes requested):

examples/keygen/keygen.c
aidangarske: Thank you, fixed.

examples/run_examples.sh
```diff
@@ -251,6 +251,14 @@ if [ $WOLFCRYPT_ENABLE -eq 1 ]; then
     ./examples/keygen/keyload ecckeyblobeh.bin -ecc -eh >> $TPMPWD/run.out 2>&1
     RESULT=$?
     [ $RESULT -ne 0 ] && echo -e "keyload endorsement ecc failed! $RESULT" && exit 1
+
+    # Test KeyGen with custom password
+    ./examples/keygen/keygen rsakeyblobeh.bin -rsa -eh -t -auth=custompass >> $TPMPWD/run.out 2>&1
+    RESULT=$?
+    [ $RESULT -ne 0 ] && echo -e "keygen endorsement rsa failed! $RESULT" && exit 1
+    ./examples/keygen/keygen rsakeyblobeh.bin -rsa -eh -auth=custompass >> $TPMPWD/run.out 2>&1
```
dgarske, line 259: Looks like the password isn't needed to load the key and get a handle. It is only needed when you use it later to sign with it. So you can also add tests like: `./examples/keygen/keyload rsakeyblobeh.bin -eh`

dgarske: Honestly not much value in adding this test until it actually uses it. Please add a TODO note.

dgarske (changes requested):

dgarske (approved):

PR #411, opened 2025-05-01
examples: bench: Fix handle leak in bench_sym_hash

iorlov-dev:
If one of the TPM2 functions called from bench_sym_hash returns an error, the exit path won't unload the hash handle leaving it dangling. This could eventually lead to OOM TPM issues (like TPM2_RC_OBJECT_MEMORY).

Unload the handle on the exit path if rc is non-zero.


dgarske:
Thank you @iorlov-dev for this report! As this is a small fix, I've treated this as a bug report and put up a fix here: https://github.com/wolfSSL/wolfTPM/pull/412
Can you tell us more about your project and use case for wolfTPM?
Thanks, David Garske, wolfSSL

PR #412, opened 2025-05-02
Fix issue with possible handle leak in bench example

dgarske:
Fix issue with possible handle leak in bench example. Reported in #411.


embhorn (approved):
Approved

PR #427, opened 2025-07-17
Fix possible buffer overrun issues

dgarske:
Fix possible buffer overrun issues. Specifically issues with exporting RSA keys where the wolfCrypt max key size doesn't match what TPM supports in `wolfTPM2_RsaKey_TpmToWolf`. ZD 20237.

Fix checks around `TPM2_GetHashDigestSize`.
Fix for HASH_COUNT (make sure its at least 2 by default). 
Fix for `TPM2_GetWolfRng` to ensure NULL is set on RNG init error.
Fix some of the configurable limit defaults.

Implement an address sanitizer test.

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -915,7 +915,8 @@ int wolfTPM2_SetAuth(WOLFTPM2_DEV* dev, int index,
   #ifdef WOLFTPM_DEBUG_VERBOSE
     printf("Session %d: Edit\n", index);
     printf("\tHandle 0x%x -> 0x%x\n", session->sessionHandle, sessionHandle);
-    printf("\tAttributes 0x%x -> 0x%x\n", session->sessionAttributes, sessionAttributes);
+    printf("\tAttributes 0x%x -> 0x%x\n", session->sessionAttributes,
+        sessionAttributes);
     if (auth) {
         printf("\tAuth Sz %d -> %d\n", session->auth.size, auth->size);
         TPM2_PrintBin(session->auth.buffer, session->auth.size);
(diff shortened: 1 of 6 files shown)


src/tpm2_wrap.c
```diff
@@ -980,8 +985,8 @@ int wolfTPM2_SetAuthHandle(WOLFTPM2_DEV* dev, int index,
             TPM2_PrintBin(handle->name.name, handle->name.size);
         #endif
             session->policyAuth = handle->policyAuth;
-            if ((word32)handle->auth.size + authDigestSz >
-                    sizeof(session->auth.buffer)) {
+            if (authDigestSz <= 0 ||
+                (handle->auth.size + authDigestSz) > (int)sizeof(session->auth.buffer)) {
```
embhorn: Past 80 char width here in several places where the condition was added

src/tpm2_wrap.c
```diff
@@ -1014,25 +1019,38 @@ int wolfTPM2_SetAuthHandleName(WOLFTPM2_DEV* dev, int index,
         if (session->sessionHandle == TPM_RS_PW) {
             /* password based authentication */
             session->auth.size = handle->auth.size;
-            XMEMCPY(session->auth.buffer, handle->auth.buffer, handle->auth.size);
+            if (session->auth.size > sizeof(session->auth.buffer)) {
+                session->auth.size = sizeof(session->auth.buffer); /* truncate */
```
embhorn, line 1025: In all these places where you truncate, should it actually be an error? Will the function return "success" but send back invalid data? If so, it would it be better to just return an error?

dgarske: The auth can be a password or hmac. Truncate seems like the right solution. If it is the wrong value then the device will return an error.

src/tpm2_wrap.c
```diff
@@ -5273,6 +5412,10 @@ static int wolfTPM2_ComputeSymmetricUnique(WOLFTPM2_DEV* dev, int hashAlg,
         }
         if (rc == 0) {
             word32 uniqueSz = TPM2_GetHashDigestSize(hashAlg);
+            if (uniqueSz == 0 || uniqueSz > sizeof(unique->buffer)) {
+                rc = BUFFER_E;
+            }
+        if (rc == 0) {
```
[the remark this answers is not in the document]
dgarske: Thank you! This found a bug with my changes. I fixed and added a test.

embhorn (approved):

PR #450, opened 2026-01-08
Add TPM support for wc_SignCert_cb callback API

jackctj117:
Prereq PR: https://github.com/wolfSSL/wolfssl/pull/9632
This pull request introduces a new callback-based approach for TPM-based certificate and CSR signing, which is recommended for FIPS compliance and simplifies the signing process. The changes add a dedicated internal signing callback and supporting context structure, update the main CSR signing function to select between the new and legacy approaches based on the device ID, and provide thorough documentation for both methods.

**Callback-based TPM signing enhancements:**

* Added `wolfTPM2_SignCertCb`, an internal callback that implements the `wc_SignCertCb` interface, enabling direct TPM-based signing for both RSA and ECC keys without relying on wolfCrypt crypto callbacks. This supports FIPS compliance and simplifies the code path.
* Introduced `CSR_MakeAndSign_Cb`, a new internal function that generates and signs CSRs or self-signed certificates using the callback-based approach, leveraging the new signing callback for direct TPM operations.
* Defined the `TpmSignCbCtx` structure in `wolftpm/tpm2_wrap.h` to hold TPM device and key references for use with the signing callback.

**Main CSR signing function update:**

* Updated `wolfTPM2_CSR_MakeAndSign_ex` to select the callback-based signing approach when `devId` is `INVALID_DEVID`, and fall back to the legacy crypto callback method for backward compatibility. This includes logic to determine key type and set signature type appropriately. [[1]](diffhunk://#diff-a30be5d9e35466ee621c1960901aeca99dfba85d6e33e37e9ff2d1ba22cebffeL7579-R7797) [[2]](diffhunk://#diff-a30be5d9e35466ee621c1960901aeca99dfba85d6e33e37e9ff2d1ba22cebffeR7806-R7820)

**Documentation improvements:**

* Added detailed comments in `csr.c` describing both the new callback-based and legacy crypto callback approaches for TPM certificate signing, including usage recommendations and instructions.

Dependent on https://github.com/wolfSSL/wolfssl/pull/9632

Testing
Setup:
TPM simulator: swtpm running on port 2321
Built wolfSSL with: --enable-certgen --enable-certreq --enable-certext --enable-cryptocb
Built wolfTPM with: --enable-swtpm --enable-certgen --enable-debug
Tests Run:
Generated RSA and ECC test keys in TPM
Created CSRs using ./examples/csr/csr
Validated CSRs with openssl req -text -noout
Results:
wc_SignCert_cb compiled into wolfSSL
wolfTPM2_SignCertCb and CSR_MakeAndSign_Cb compiled into wolfTPM
Generated valid RSA (1228 bytes) and ECC (696 bytes) CSRs
CSRs verified successfully with OpenSSL

### DIFF
--- a/.github/workflows/make-test-swtpm.yml
+++ b/.github/workflows/make-test-swtpm.yml
@@ -67,7 +67,7 @@ jobs:
           - name: certsigncb
             wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-certsigncb
             wolftpm_config: --enable-swtpm --enable-certgen
-            test_command: "make check && WOLFSSL_PATH=./wolfssl ./examples/run_examples.sh"
+            test_command: "make check && WOLFSSL_PATH=./wolfssl ./examples/run_examples.sh && ./examples/csr/csr -signcb && ./examples/csr/csr -signcb -cert"
 
           # STMicro ST33KTPM2
           - name: st33ktpm2
--- a/examples/csr/csr.c
+++ b/examples/csr/csr.c
@@ -86,9 +86,9 @@ static const char* gClientCertEccFile = ECC_CERT_PEM;
  *    - Maintains backward compatibility
  *    - Requires crypto callback setup
  * 
- * This example demonstrates the crypto callback approach. To use the new
- * callback-based approach, pass INVALID_DEVID instead of tpmDevId when calling
- * TPM2_CSR_Generate() or wolfTPM2_CSR_MakeAndSign_ex().
+ * This example demonstrates both approaches. By default it uses the crypto
+ * callback approach. Use the -signcb option to use the new callback-based
+ * approach, which passes INVALID_DEVID to wolfTPM2_CSR_MakeAndSign_ex().
  */
 
 static int TPM2_CSR_Generate(WOLFTPM2_DEV* dev, int keyType, WOLFTPM2_KEY* key,
@@ -173,9 +173,11 @@ static int TPM2_CSR_Generate(WOLFTPM2_DEV* dev, int keyType, WOLFTPM2_KEY* key,
 static void usage(void)
 {
     printf("Expected usage:\n");
-    printf("./examples/csr/csr [-cert]\n");
+    printf("./examples/csr/csr [-cert] [-signcb]\n");
     printf("\t-cert: Make self signed certificate instead of "
                     "default CSR (Certificate Signing Request)\n");
+    printf("\t-signcb: Use wc_SignCert_cb callback-based signing "
+                    "(FIPS compliant, requires WOLFSSL_CERT_SIGN_CB)\n");
 }
 
 int TPM2_CSR_Example(void* userCtx)
@@ -192,6 +194,7 @@ int TPM2_CSR_ExampleArgs(void* userCtx, int argc, char *argv[])
     int tpmDevId;
     TPMT_PUBLIC publicTemplate;
     int makeSelfSignedCert = 0;
+    int useSignCb = 0;
 
     printf("TPM2 CSR Example\n");
 
@@ -206,6 +209,9 @@ int TPM2_CSR_ExampleArgs(void* userCtx, int argc, char *argv[])
     while (argc > 1) {
         if (XSTRCMP(argv[argc-1], "-cert") == 0) {
             makeSelfSignedCert = 1;
+        }
+        else if (XSTRCMP(argv[argc-1], "-signcb") == 0) {
+            useSignCb = 1;
         }
         else {
             printf("Warning: Unrecognized option: %s\n", argv[argc-1]);
@@ -245,7 +251,8 @@ int TPM2_CSR_ExampleArgs(void* userCtx, int argc, char *argv[])
         if (rc == 0) {
             rc = TPM2_CSR_Generate(&dev, RSA_TYPE, &key,
                 makeSelfSignedCert ? gClientCertRsaFile : gClientCsrRsaFile,
-                makeSelfSignedCert, tpmDevId, CTC_SHA256wRSA);
+                makeSelfSignedCert,
+                useSignCb ? INVALID_DEVID : tpmDevId, CTC_SHA256wRSA);
         }
         wolfTPM2_UnloadHandle(&dev, &key.handle);
         wolfTPM2_UnloadHandle(&dev, &storageKey.handle);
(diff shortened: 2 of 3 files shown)


dgarske (dismissed):
Looks good thanks!

src/tpm2_wrap.c
```diff
@@ -7332,6 +7332,103 @@ typedef struct CSRKey {
...
+            byte *r, *s;
+            word32 rLen, sLen;
+
+            /* Split R and S from concatenated signature */
+            rLen = sLen = rsLen / 2;
+            r = &sigRS[0];
+            s = &sigRS[rLen];
+
+            /* Encode as DER ECDSA signature */
+            rc = wc_ecc_rs_raw_to_sig(r, rLen, s, sLen, out, outLen);
```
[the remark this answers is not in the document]
aidangarske: agreed, please update because there is missing validation that rsLen is even before splitting. If rsLen is odd (shouldn't happen with valid ECC signatures) R and S 
  could be incorrectly split.

src/tpm2_wrap.c
```diff
@@ -7382,6 +7479,149 @@ static int CSR_MakeAndSign(WOLFTPM2_DEV* dev, WOLFTPM2_CSR* csr, CSRKey* key,
...
+    else {
+        rc = BAD_FUNC_ARG;
+    }
+
+    /* Setup signing context */
+    if (rc == 0) {
+        signCtx.dev = dev;
+        signCtx.key = key;
+    }
+
```
aidangarske, line 7576: The legacy path in `CSR_KeySetup()` adds the SKID extension via `wc_SetSubjectKeyIdFromPublicKey_ex()` (see existing code around like    
  line 7717-7722). This new callback function does not set SKID so CSRs generated via the callback path will be missing this extension compared to   
  the legacy path. Consider adding SKID from the public key here

src/tpm2_wrap.c
```diff
@@ -7382,6 +7479,149 @@ static int CSR_MakeAndSign(WOLFTPM2_DEV* dev, WOLFTPM2_CSR* csr, CSRKey* key,
...
+ *   BAD_FUNC_ARG: invalid parameters
+ *   BUFFER_E: output buffer too small
+ *   Other negative: error code from wolfSSL or TPM
+ */
+static int CSR_MakeAndSign_Cb(WOLFTPM2_DEV* dev, WOLFTPM2_CSR* csr,
+    WOLFTPM2_KEY* key, int keyType, int outFormat, byte* out, int outSz,
+    int selfSignCert)
+{
+    int rc = 0;
+    TpmSignCbCtx signCtx;
```
aidangarske, line 7525: Should this be initialized?

src/tpm2_wrap.c
```diff
@@ -7382,6 +7479,149 @@ static int CSR_MakeAndSign(WOLFTPM2_DEV* dev, WOLFTPM2_CSR* csr, CSRKey* key,
...
+    if (keyType == ECC_TYPE) {
+    #ifdef HAVE_ECC
+        wc_ecc_free(&wolfKey.ecc);
+    #endif
+    }
+    else if (keyType == RSA_TYPE) {
+    #ifndef NO_RSA
+        wc_FreeRsaKey(&wolfKey.rsa);
+    #endif
+    }
```
aidangarske, line 7637: Should we track initialization state here? Maybe add `int keyInited = 0;` and only calling free if init succeeded?

src/tpm2_wrap.c
```diff
@@ -7382,6 +7479,149 @@ static int CSR_MakeAndSign(WOLFTPM2_DEV* dev, WOLFTPM2_CSR* csr, CSRKey* key,
...
+    if (rc >= 0) {
+        rc = wc_SignCert_cb(csr->req.bodySz, csr->req.sigType, out,
+            (word32)outSz, keyType, wolfTPM2_SignCertCb, &signCtx,
+            wolfTPM2_GetRng(dev));
+    }
+
+    /* Optionally convert to PEM */
+    if (rc >= 0 && outFormat == ENCODING_TYPE_PEM) {
+#ifdef WOLFSSL_DER_TO_PEM
+        byte tmp[MAX_CONTEXT_SIZE];
```
aidangarske, line 7607: `MAX_CONTEXT_SIZE` can be 4096+ bytes on the stack. This could be problematic for embedded systems using `--enable-smallstack`. Consider noting this         
  limitation or using like dynamic allocation when `WOLFTPM_SMALL_STACK` is defined which might be nice in this case..

examples/csr/csr.c
```diff
@@ -67,6 +67,30 @@ static const char* gClientCertEccFile = ECC_CERT_PEM;
...
+ * 2. CRYPTO CALLBACK APPROACH (Legacy/Backward Compatible):
+ *    When devId is set via wolfTPM2_SetCryptoDevCb(), the legacy crypto
+ *    callback infrastructure is used. This approach:
+ *    - Uses wc_SignCert_ex() which offloads crypto operations to TPM
+ *    - Maintains backward compatibility
+ *    - Requires crypto callback setup
+ * 
+ * This example demonstrates the crypto callback approach. To use the new
+ * callback-based approach, pass INVALID_DEVID instead of tpmDevId when calling
+ * TPM2_CSR_Generate() or wolfTPM2_CSR_MakeAndSign_ex().
```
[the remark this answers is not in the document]
aidangarske: Please fix this as well

.github/workflows/make-test-swtpm.yml
```diff
@@ -63,6 +63,12 @@ jobs:
           # Infineon SLB9673
           - name: slb9673
             wolftpm_config: --enable-infineon=slb9673 --enable-i2c
+          # Cert sign callback (wc_SignCert_cb)
+          - name: certsigncb
+            wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-certsigncb
+            wolftpm_config: --enable-swtpm --enable-certgen
+            test_command: "make check && WOLFSSL_PATH=./wolfssl ./examples/run_examples.sh"
```
aidangarske: Your new logic isnt actually hitting the `wc_SignCert_cb` code path I think. The CSR example always calls `wolfTPM2_SetCryptoDevCb()` to get a `tpmDevId`, then 
  passes that to `TPM2_CSR_Generate()`. The new callback path is only used when `devId == INVALID_DEVID` To actually test the new code path, the CSR example needs a command-line option (e.g., -signcb) that passes `INVALID_DEVID` instead of `tpmDevId`. Then      
  update the test_command to run `./examples/csr/csr -signcb`. This needs updated in documentation as well please

aidangarske (changes requested):
Nice job with this I have some nits and changes. 

I verified this run and test with RSA and ECC and CSR generation and OpenSSL validation passes. The CI currently tests compilation but not the actual new callback code path. The missing SKID extension could cause issues for users who rely on that extension being present in their CSRs.

Also Another thing to consider is updating `wolfTPM2_CSR_Generate_ex()` to also support the callback-based approach for consistency, or document that only `wolfTPM2_CSR_MakeAndSign_ex()` supports the new path.

src/tpm2_wrap.c
jackctj117: The same pattern exists in the legacy CSR_MakeAndSign(), do you still want me to change it?

aidangarske: I might update it in both places but use your best discretion

jackctj117: Once this is merged I can open up a new PR for both

PR #455, opened 2026-02-17
Improve get_ek_certs to handle indices

dgarske:


examples/endorsement/get_ek_certs.c
```diff
@@ -237,7 +237,34 @@ int TPM2_EndorsementCert_Example(void* userCtx, int argc, char *argv[])
         /* Get Endorsement Public Key template using NV index */
         rc = wolfTPM2_GetKeyTemplate_EKIndex(nvIndex, &publicTemplate);
         if (rc != 0) {
-            printf("EK Index 0x%08x not valid\n", nvIndex);
+            printf("NV Index 0x%08x is not an EK certificate index\n", nvIndex);
+
+            /* Try to read the NV public info to show what it contains */
+            rc = wolfTPM2_NVReadPublic(&dev, nvIndex, &nvPublic);
```
dgarske, line 325: Let's actually mention "indices"

examples/endorsement/get_ek_certs.c
```diff
@@ -237,7 +267,112 @@ int TPM2_EndorsementCert_Example(void* userCtx, int argc, char *argv[])
...
+                indexType = "EK Certificate Chain";
+            }
+            else if (offset >= 0x7F01 && offset <= 0x7F04) {
+                indexType = "EK Policy Index";
+                if (offset == 0x7F01) indexType = "EK Policy Index (SHA256)";
+                else if (offset == 0x7F02) indexType = "EK Policy Index (SHA384)";
+                else if (offset == 0x7F03) indexType = "EK Policy Index (SHA512)";
+                else if (offset == 0x7F04) indexType = "EK Policy Index (SM3_256)";
+            }
+            else if (nvIndex > TPM_20_TCG_NV_SPACE + 0x7FFF) {
```
aidangarske: The range expansion from 0x1FF to 0x7FFF is a 32x increase with no explanation 
for why 0x7FFF is the correct upper bound. The policy indices at 0x7F01-0x7F04 
are clearly the motivation, but there's now a large unvalidated gap between 
0x1FF and 0x7F01 that this function will silently accept and likely produce a 
garbage template for. 

Consider adding a guard inside the function body to reject the dead zone kinda like this:
```
    if (offset > 0x1FF && offset < 0x7F01) {
        return BAD_FUNC_ARG;
    }
```

Also consider updating the comment to explain all four stages this gate is now 
covering

examples/endorsement/get_ek_certs.c
```diff
@@ -237,7 +267,112 @@ int TPM2_EndorsementCert_Example(void* userCtx, int argc, char *argv[])
         /* Get Endorsement Public Key template using NV index */
         rc = wolfTPM2_GetKeyTemplate_EKIndex(nvIndex, &publicTemplate);
         if (rc != 0) {
-            printf("EK Index 0x%08x not valid\n", nvIndex);
+            const char* indexType = "Unknown";
+            word32 offset = nvIndex - TPM_20_TCG_NV_SPACE;
```
aidangarske: `offset` is computed here unconditionally but the 
check `if (nvIndex < TPM_20_TCG_NV_SPACE)` doesn't happen until a few lines 
later. `word32` is unsigned if nvIndex is below `TPM_20_TCG_NV_SPACE` 
this silently wraps to a huge value before the guard ever fires.

Consider maybe moving the offset calculation inside the else branch after the bounds check

wolftpm/tpm2.h
```diff
@@ -1708,6 +1708,12 @@ typedef struct TPM2_AUTH_SESSION {
 /* EK Certificate Chains (0x100 - 0x1FF) - Not common */
 #define TPM2_NV_EK_CHAIN               (TPM_20_TCG_NV_SPACE + 0x100)
 
+/* EK Policy Indices for PolicyAuthorizeNV (0x7F01 - 0x7F04) */
+#define TPM2_NV_EK_POLICY_SHA256       (TPM_20_TCG_NV_SPACE + 0x7F01)
+#define TPM2_NV_EK_POLICY_SHA384       (TPM_20_TCG_NV_SPACE + 0x7F02)
```
aidangarske, line 1713: These four defines are added here but never actually referenced in the PR can you use these for the cosntants

examples/endorsement/get_ek_certs.c
```diff
@@ -237,7 +267,112 @@ int TPM2_EndorsementCert_Example(void* userCtx, int argc, char *argv[])
...
+                }
+
+                /* Read and display data if appropriate */
+                if (showData && nvPublic.dataSize > 0) {
+                    certSz = nvPublic.dataSize;
+                    if (certSz > sizeof(certBuf)) {
+                        certSz = sizeof(certBuf);
+                    }
+                    
+                    rc = wolfTPM2_NVReadAuth(&dev, &nv, nvIndex, certBuf, &certSz, 0);
```
aidangarske, line 370: Should we open a new nv handle here if its not correctly initialized for nvIndex?

aidangarske (approved):

PR #492, opened 2026-04-22
Enhance the TPM 2 native_test to include additional ECC curves (P384)

dgarske:


aidangarske (commented):
## Skoll Code Review

**Scan type:** `review-security`
**Findings:** 4 total — 3 posted, 1 skipped

### Posted findings
- [Medium] **TPM_RC_VALUE treated as 'curve unsupported' for every TPM call masks genuine regressions** — `examples/native/native_test.c:68-79`
- [Low] **Refactor weakens Create/Load failure detection: previously-hard-failed error codes are now silent skips** — `examples/native/native_test.c:144-178`
- [Low] **goto done after rc=-1 still triggers caller's `if (rc != 0) goto exit` but prints a confusing skip message path** — `examples/native/native_test.c:229-238`

### Skipped findings
- [Info] ~~P-384 key created with SHA-256 nameAlg and scheme hash may be rejected by strict TPMs~~

examples/native/native_test.c
```diff
@@ -60,6 +60,251 @@ typedef struct tmpHandle {
 } TpmHandle;
 
 
+/* Returns 1 if the TPM response code indicates the requested curve (or the
+ * command itself) is not supported by this TPM — these are graceful skips,
+ * not test failures. Covers TPM_RC_CURVE (wrapped with parameter/handle/
+ * session flags), TPM_RC_VALUE (some TPMs surface unsupported curves this
+ * way on Create), and TPM_RC_COMMAND_CODE for missing commands. */
+static int is_curve_or_cmd_unsupported(TPM_RC rc)
```
aidangarske: **🟠 [Medium] TPM_RC_VALUE treated as 'curve unsupported' for every TPM call masks genuine regressions** · _Logic_

The new helper returns 1 for any `rc` whose FMT1-masked value equals `TPM_RC_VALUE`, and the caller uses that result to convert the error into a silent skip (`rc = 0; goto done;`) for every step in the sequence: `TPM2_Create`, `TPM2_Load`, `TPM2_ECDH_KeyGen`, `TPM2_ECDH_ZGen`…

**Fix:** Either (a) restrict the TPM_RC_VALUE branch to the `TPM2_Create` call site (which is where the justification actually applies) by handling it inline rather…

examples/native/native_test.c
```diff
@@ -60,6 +60,251 @@ typedef struct tmpHandle {
...
+    cmdIn.create.inPublic.publicArea.parameters.eccDetail.symmetric.algorithm
+        = TPM_ALG_NULL;
+    cmdIn.create.inPublic.publicArea.parameters.eccDetail.scheme.scheme
+        = TPM_ALG_ECDH;
+    cmdIn.create.inPublic.publicArea.parameters.eccDetail.scheme.details.ecdsa.hashAlg
+        = TPM_ALG_SHA256;
+    cmdIn.create.inPublic.publicArea.parameters.eccDetail.curveID = curveID;
+    cmdIn.create.inPublic.publicArea.parameters.eccDetail.kdf.scheme
+        = TPM_ALG_NULL;
+    rc = TPM2_Create(&cmdIn.create, &cmdOut.create);
```
aidangarske, line 152: **🔵 [Low] Refactor weakens Create/Load failure detection: previously-hard-failed error codes are now silent skips** · _Regression_

BEFORE the PR, a `TPM_RC_CURVE` or `TPM_RC_VALUE` return from `TPM2_Create` or `TPM2_Load` on the P-256 path would `goto exit` with the real error, failing the test (the BEFORE code wrote `if (rc != TPM_RC_SUCCESS) { printf(...); goto exit; }` for those calls). AFTER the PR the…

**Fix:** Gate the "graceful skip on unsupported" branches on `curveID != TPM_ECC_NIST_P256` (or on an explicit "may-skip" flag passed into the helper), so that P-256…

examples/native/native_test.c
```diff
@@ -60,6 +60,251 @@ typedef struct tmpHandle {
...
+    }
+    if (rc != TPM_RC_SUCCESS) {
+        printf("TPM2_ECDH_ZGen failed 0x%x: %s\n", rc,
+            TPM2_GetRCString(rc));
+        goto done;
+    }
+    printf("TPM2_ECDH_ZGen: zPt %d\n",
+        cmdOut.ecdhZ.outPoint.size);
+
+    /* Verify ECDH_KeyGen and ECDH_ZGen produced the same shared secret */
```
aidangarske, line 256: **🔵 [Low] goto done after rc=-1 still triggers caller's `if (rc != 0) goto exit` but prints a confusing skip message path** · _Logic_

When the ECDH/ZGen shared-secret comparison fails, the code sets `rc = -1`, prints "TPM2 ECC Shared Secret Fail", and `goto done`. The `done:` label flushes the ECC key (good) and returns `rc`, which is `-1`. The caller (`for (eccCurveIdx = 0; ...)`) checks `rc != 0` and does…

**Fix:** Return a distinct, non-TPM-RC-colliding error value (e.g., a static `const int TEST_RC_SHARED_SECRET_MISMATCH = -1001;`) and let the helper print the full…

examples/native/native_test.c
```diff
@@ -60,6 +60,265 @@ typedef struct tmpHandle {
...
+
+    printf("--- ECDH/ZGen tests for curve 0x%04x ---\n", curveID);
+
+    /* set session auth for storage key */
+    session[0].auth.size = storagePwdSz;
+    XMEMCPY(session[0].auth.buffer, storagePwd, storagePwdSz);
+
+    /* Create an ECC key for ECDH */
+    XMEMSET(&cmdIn.create, 0, sizeof(cmdIn.create));
+    cmdIn.create.parentHandle = parentHandle;
```
aidangarske, line 134: **🟠 [Medium] P-384 key configured with SHA-256 nameAlg/scheme — deviates from repo convention**

The shared helper is invoked for both `TPM_ECC_NIST_P256` and `TPM_ECC_NIST_P384`, but the `Create_In` is hard-coded with `nameAlg = TPM_ALG_SHA256` and `scheme.details.ecdsa.hashAlg = TPM_ALG_SHA256`. Elsewhere in the repo the convention is to pair the curve with a…

**Fix:** Derive nameAlg and scheme hashAlg from the curve (SHA-256 for P-256, SHA-384 for P-384) to match the existing repo convention and reduce spurious P-384…

examples/native/native_test.c
```diff
@@ -192,6 +451,14 @@ int TPM2_Native_TestArgs(void* userCtx, int argc, char *argv[])
         "\x06\xC1";
 
     int perform_EncryptDecrypt2 = 1;
+    int eccCurveIdx;
+    /* Exercise ECDH / EC_Ephemeral / ZGen_2Phase on both P-256 and P-384 so
+     * intermittent coordinate-padding bugs are caught in CI (P-384 surfaces
+     * leading-zero-strip issues ~2x more often than P-256 in practice). */
+    static const UINT16 eccTestCurves[2] = {
```
aidangarske: **🔵 [Low] Two parallel arrays must stay in sync — one-step-off error waiting to happen**

`eccTestCurves` and `eccTestKeySizes` are declared as two separate fixed-length arrays that must stay index-aligned. If a future contributor adds P-521 to `eccTestCurves` but forgets to extend `eccTestKeySizes`, the loop silently indexes the wrong entry (or reads past the array…

**Fix:** Combine the two parallel arrays into a single struct-of-arrays to keep curve and expected-key-size paired at the declaration site.

examples/native/native_test.c
```diff
@@ -1067,155 +1334,15 @@ int TPM2_Native_TestArgs(void* userCtx, int argc, char *argv[])
...
-            }
-            printf("TPM2 Two-Phase ECDH Pass\n");
-        }
-        else if (WOLFTPM_IS_COMMAND_UNAVAILABLE(rc)) {
-            printf("TPM2_ZGen_2Phase: not supported by TPM\n");
-        }
-        else {
-            printf("TPM2_ZGen_2Phase failed 0x%x: %s\n", rc,
-                TPM2_GetRCString(rc));
+    for (eccCurveIdx = 0; eccCurveIdx < 2; eccCurveIdx++) {
```
aidangarske: **🔵 [Low] Loop bound uses magic literal instead of array length**

`for (eccCurveIdx = 0; eccCurveIdx < 2; eccCurveIdx++)` hard-codes `2` even though the curve count is driven by `eccTestCurves`. Adding a third curve requires updating three places (both array sizes plus the loop bound). Deriving the count from…

**Fix:** Use a `sizeof()`-derived count so the loop automatically picks up new curves added to the table.

examples/native/native_test.c
```diff
@@ -60,6 +60,265 @@ typedef struct tmpHandle {
...
+ * unsupported), non-zero on hard failure. */
+static int test_TPM2_ECDH_ZGen_curve(TPM_HANDLE parentHandle,
+    TPM2_AUTH_SESSION* session,
+    const char* storagePwd, int storagePwdSz,
+    const char* usageAuth, int usageAuthSz,
+    UINT16 curveID, int expectedEccKeySz)
+{
+    int rc;
+    int mayBeUnsupported;
+    TpmEccKey eccKey;
```
aidangarske, line 98: **🔵 [Low] `zCompare` uses `TPM2B_PUBLIC_KEY_RSA` to hold an ECC x-coordinate**

The helper declares `TPM2B_PUBLIC_KEY_RSA zCompare;` solely to hold a copy of the ECDH z-point for later comparison against `ECDH_ZGen`'s output. The buffer size is adequate (`MAX_RSA_KEY_BYTES`), and the surrounding test code historically reused `message` (also…

**Fix:** Rename/retype `zCompare` so its declaration doesn't hint at RSA — e.g. use `TPM2B_DIGEST` or a raw byte buffer sized for the largest supported curve point.

examples/native/native_test.c
```diff
@@ -60,6 +60,265 @@ typedef struct tmpHandle {
...
+    cmdIn.create.inPublic.publicArea.nameAlg = TPM_ALG_SHA256;
+    cmdIn.create.inPublic.publicArea.objectAttributes = (
+        TPMA_OBJECT_sensitiveDataOrigin | TPMA_OBJECT_userWithAuth |
+        TPMA_OBJECT_decrypt | TPMA_OBJECT_noDA);
+    cmdIn.create.inPublic.publicArea.parameters.eccDetail.symmetric.algorithm
+        = TPM_ALG_NULL;
+    cmdIn.create.inPublic.publicArea.parameters.eccDetail.scheme.scheme
+        = TPM_ALG_ECDH;
+    cmdIn.create.inPublic.publicArea.parameters.eccDetail.scheme.details.ecdsa.hashAlg
+        = TPM_ALG_SHA256;
```
aidangarske: **🟠 [Medium] The new P-384 path can pass without ever exercising the added regression check**

This refactor turns every newly added curve-specific path into a best-effort check: on any "unsupported" result, `test_TPM2_ECDH_ZGen_curve()` resets `rc` to `0`, and the caller in `TPM2_Native_TestArgs()` treats that as a passing iteration. That means the PR's advertised P-384…

**Fix:** Make the new P-384 assertion observable in at least one guaranteed test environment; otherwise this PR can merge without actually validating the added curve.

---
> **Note:** Referenced line (`examples/native/native_test.c:145-149,168-172,190-194,215-219,244-248,268-272`) is outside the diff hunk. Comment anchored to nearest changed region.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review`

### Posted findings
- [Medium] **The new P-384 path can pass without ever exercising the added regression check** — `examples/native/native_test.c:145-149,168-172,190-194,215-219,244-248,268-272`
- [Medium] **P-384 key configured with SHA-256 nameAlg/scheme — deviates from repo convention** — `examples/native/native_test.c:131-140`
- [Low] **Two parallel arrays must stay in sync — one-step-off error waiting to happen** — `examples/native/native_test.c:444-447`
- [Low] **Loop bound uses magic literal instead of array length** — `examples/native/native_test.c:1323`
- [Low] **`zCompare` uses `TPM2B_PUBLIC_KEY_RSA` to hold an ECC x-coordinate** — `examples/native/native_test.c:96`

aidangarske (approved):

PR #493, opened 2026-04-24
Fix for FwCmd_StirRandom with custom hardware rand generate

dgarske:
Gated wc_RNG_DRBG_Reseed in FwCmd_StirRandom on ctx->rng.drbg != NULL. Fixes the TPM2_StirRandom → TPM_RC_FAILURE failure with CUSTOM_RAND_GENERATE_BLOCK+HAVE_HASHDRBG; HW-RNG-backed TPMs now treat it as a TCG-compliant no-op.

### DIFF
--- a/src/fwtpm/fwtpm_command.c
+++ b/src/fwtpm/fwtpm_command.c
@@ -901,11 +901,18 @@ static TPM_RC FwCmd_StirRandom(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
         printf("fwTPM: StirRandom(%d bytes)\n", inDataSize);
     #endif
 
-        /* Reseed the DRBG with the caller-provided additional input */
+        /* Reseed the Hash DRBG with the caller-provided additional input.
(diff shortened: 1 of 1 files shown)


dgarske:
Thanks @Copilot.

Re: 1 (struct internals) — fixed. Replaced the runtime `ctx->rng.drbg != NULL` check with a compile-time gate `#if defined(HAVE_HASHDRBG) && !defined(CUSTOM_RAND_GENERATE_BLOCK)`. Slightly broader than `CUSTOM_RAND_GENERATE_BLOCK` alone since the DRBG is also absent under `WC_NO_HASHDRBG`. No struct internals accessed.

Re: 2 (CI/test for HW-RNG path) — out of scope for this PR. The fwTPM unit harness has no hook to swap wolfCrypt's RNG backend, and a dedicated CI variant just for `StirRandom` is disproportionate. Will revisit when other HW-RNG paths need coverage.

aidangarske (approved):

PR #500, opened 2026-05-03
configure.ac: --enable-pqc fails with current wolfSSL master because mlkem.h was renamed wc_mlkem.h

eramusa:
## Summary
`wolfTPM/configure.ac` AC_CHECK_DECL probes for `wc_MlKemKey_Init` by including `<wolfssl/wolfcrypt/mlkem.h>`, but wolfSSL HEAD installs the header under `wc_mlkem.h` (with the `wc_` prefix). The probe fails, and `--enable-pqc` configure aborts.

This PR fixes the header include in `configure.ac` to correctly use `wc_mlkem.h`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -728,7 +728,7 @@ else
             [WOLFTPM_HAVE_MLKEM_FN=yes],
             [WOLFTPM_HAVE_MLKEM_FN=no],
             [[#include <wolfssl/options.h>
-              #include <wolfssl/wolfcrypt/mlkem.h>]])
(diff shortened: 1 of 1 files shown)


dgarske:
Hi @eramusa thank you for the bug report and PR. We require a signed contributor form. As this is a small PR we are going to look at your original issues and opt to close this PR. Can you tell us more about your project and interest in wolfTPM with PQC? Thanks, David Garske, wolfSSL

PR #509, opened 2026-05-20
ML-DSA renaming

Frauschi:
Move to the new `wc_MlDsaKey` API and get rid of all "Dilithium" naming.


Frauschi:
I made the backward compatibility CI test more version and PQC aware to only test PQC with wolfssl master and a future post 5.9.1 stable release (as these contain the new ML-DSA API).

aidangarske (approved):
Looks good thank you!

PR #519, opened 2026-06-03
Add new WOLFTPM2_ECC_DEFAULT_CURVE

dgarske:
Fixes ZD 21780

### DIFF
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -7038,9 +7038,7 @@ TPM_ALG_ID TPM2_GetCurveHashAlg(TPM_ECC_CURVE curveID)
 {
     int curveSz = TPM2_GetCurveSize(curveID);
 
-    /* Pair the message digest strength to the curve size, matching the TCG
-     * recommended curve/hash combinations (P256/SHA256, P384/SHA384,
-     * P521/SHA512). */
+    /* Pair digest strength to curve size per TCG (P256/SHA256, P384/SHA384, P521/SHA512). */
     if (curveSz >= 64) /* P521 (66), BP512 (64) */
         return TPM_ALG_SHA512;
     if (curveSz == 48) /* P384, BP384 */
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -7034,6 +7034,20 @@ int TPM2_GetCurveSize(TPM_ECC_CURVE curveID)
     return 0;
 }
 
+TPM_ALG_ID TPM2_GetCurveHashAlg(TPM_ECC_CURVE curveID)
+{
+    int curveSz = TPM2_GetCurveSize(curveID);
+
+    /* Pair the message digest strength to the curve size, matching the TCG
+     * recommended curve/hash combinations (P256/SHA256, P384/SHA384,
+     * P521/SHA512). */
+    if (curveSz >= 64) /* P521 (66), BP512 (64) */
+        return TPM_ALG_SHA512;
+    if (curveSz == 48) /* P384, BP384 */
+        return TPM_ALG_SHA384;
+    return TPM_ALG_SHA256; /* P256 and smaller */
+}
+
 int TPM2_GetTpmCurve(int curve_id)
 {
     int ret = -1;
--- a/src/tpm2_cryptocb.c
+++ b/src/tpm2_cryptocb.c
@@ -234,15 +234,13 @@ int wolfTPM2_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)
                     TPMT_PUBLIC publicTemplate;
                     TPMI_ALG_HASH hashAlg;
 
-                    if (curve_id == TPM_ECC_NIST_P521)
-                        hashAlg = TPM_ALG_SHA512;
-                    else if (curve_id == TPM_ECC_NIST_P384)
-                        hashAlg = TPM_ALG_SHA384;
-                    else
-                        hashAlg = TPM_ALG_SHA256;
+                    hashAlg = TPM2_GetCurveHashAlg(curve_id);
 
                     XMEMSET(&publicTemplate, 0, sizeof(publicTemplate));
-                    rc = wolfTPM2_GetKeyTemplate_ECC_ex(&publicTemplate, hashAlg,
+                    /* Use the internal helper (not the public _ex wrapper) so
+                     * the curve wolfCrypt negotiated is honored exactly and is
+                     * not remapped by WOLFTPM2_ECC_DEFAULT_CURVE. */
+                    rc = GetKeyTemplateECC(&publicTemplate, hashAlg,
                         TPMA_OBJECT_sensitiveDataOrigin | TPMA_OBJECT_userWithAuth |
                         TPMA_OBJECT_sign | TPMA_OBJECT_noDA,
                         curve_id, TPM_ALG_ECDSA, hashAlg);
--- a/src/tpm2_cryptocb.c
+++ b/src/tpm2_cryptocb.c
@@ -237,10 +237,8 @@ int wolfTPM2_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)
                     hashAlg = TPM2_GetCurveHashAlg(curve_id);
 
                     XMEMSET(&publicTemplate, 0, sizeof(publicTemplate));
-                    /* Use the internal helper (not the public _ex wrapper) so
-                     * the curve wolfCrypt negotiated is honored exactly and is
-                     * not remapped by WOLFTPM2_ECC_DEFAULT_CURVE. */
-                    rc = GetKeyTemplateECC(&publicTemplate, hashAlg,
+                    /* Explicit curve negotiated by wolfCrypt; honored exactly. */
+                    rc = wolfTPM2_GetKeyTemplate_ECC_ex(&publicTemplate, hashAlg,
                         TPMA_OBJECT_sensitiveDataOrigin | TPMA_OBJECT_userWithAuth |
                         TPMA_OBJECT_sign | TPMA_OBJECT_noDA,
                         curve_id, TPM_ALG_ECDSA, hashAlg);
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -8224,19 +8224,31 @@ int GetKeyTemplateECC(TPMT_PUBLIC* publicTemplate,
     TPM_ALG_ID nameAlg, TPMA_OBJECT objectAttributes, TPM_ECC_CURVE curve,
     TPM_ALG_ID sigScheme, TPM_ALG_ID sigHash)
 {
-    int curveSz = TPM2_GetCurveSize(curve);
+    int curveSz;
 
-    if (publicTemplate == NULL || curveSz == 0)
+    if (publicTemplate == NULL)
         return BAD_FUNC_ARG;
 
-#if defined(NO_ECC256) && defined(HAVE_ECC384) && ECC_MIN_KEY_SZ <= 384
-    /* make sure we use a curve that is enabled */
+#ifdef NO_ECC256
+    /* P256 is compiled out of wolfCrypt; substitute the configured (enabled)
+     * default curve so callers requesting P256 still get a usable curve. This
+     * is availability-driven and applies to every caller (including the named
+     * EK/SRK/AIK templates) because the build simply cannot honor P256. The
+     * configurable default-curve preference lives in the public ECC template
+     * wrappers, not here, so explicit curves through this helper are honored. */
     if (curve == TPM_ECC_NIST_P256) {
-        curve = TPM_ECC_NIST_P384;
-        nameAlg = TPM_ALG_SHA384;
-        sigHash = TPM_ALG_SHA384;
+        curve = WOLFTPM2_ECC_DEFAULT_CURVE;
+        nameAlg = TPM2_GetCurveHashAlg(curve);
+        if (sigHash != TPM_ALG_NULL)
+            sigHash = nameAlg;
     }
 #endif
+
+    /* Compute curve size after any substitution so coordinate sizes and the
+     * AES wrap strength below match the curve actually used. */
+    curveSz = TPM2_GetCurveSize(curve);
+    if (curveSz == 0)
+        return BAD_FUNC_ARG;
 
     XMEMSET(publicTemplate, 0, sizeof(TPMT_PUBLIC));
     publicTemplate->type = TPM_ALG_ECC;
@@ -8280,10 +8292,35 @@ int wolfTPM2_GetKeyTemplate_RSA(TPMT_PUBLIC* publicTemplate,
         TPM_ALG_NULL, WOLFTPM2_WRAP_DIGEST);
 }
 
+/* Resolve a P256 "use the library default" request to the build-configured
+ * WOLFTPM2_ECC_DEFAULT_CURVE, pairing the name/sig hash via
+ * TPM2_GetCurveHashAlg. No-op when the default is P256 (the shipped default),
+ * so existing builds are unchanged. This lets a build make P384 (or another
+ * curve) the wrapper default - e.g. wolfTPM2_CreateSRK(TPM_ALG_ECC) - without
+ * editing call sites. The EK index templates (TCG-fixed curve and auth policy),
+ * the explicit-curve IAK/IDevID helpers, and the crypto callback (curve
+ * negotiated by wolfCrypt) intentionally do NOT use this and keep their exact
+ * curve. */
+static void wolfTPM2_ResolveDefaultEccCurve(TPM_ECC_CURVE* curve,
+    TPM_ALG_ID* nameAlg, TPM_ALG_ID* sigHash)
+{
+    if (*curve == TPM_ECC_NIST_P256 &&
+            WOLFTPM2_ECC_DEFAULT_CURVE != TPM_ECC_NIST_P256) {
+        *curve = WOLFTPM2_ECC_DEFAULT_CURVE;
+        if (nameAlg != NULL)
+            *nameAlg = TPM2_GetCurveHashAlg(*curve);
+        if (sigHash != NULL && *sigHash != TPM_ALG_NULL)
+            *sigHash = TPM2_GetCurveHashAlg(*curve);
+    }
+}
+
 int wolfTPM2_GetKeyTemplate_ECC_ex(TPMT_PUBLIC* publicTemplate,
     TPM_ALG_ID nameAlg, TPMA_OBJECT objectAttributes, TPM_ECC_CURVE curve,
     TPM_ALG_ID sigScheme, TPM_ALG_ID sigHash)
 {
+    /* General-purpose ECC template API: a P256 request follows the configured
+     * default curve. */
+    wolfTPM2_ResolveDefaultEccCurve(&curve, &nameAlg, &sigHash);
     return GetKeyTemplateECC(publicTemplate, nameAlg,
         objectAttributes, curve, sigScheme, sigHash);
 }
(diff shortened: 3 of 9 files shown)


src/tpm2_wrap.c
```diff
@@ -8224,19 +8224,29 @@ int GetKeyTemplateECC(TPMT_PUBLIC* publicTemplate,
...
-    if (curve == TPM_ECC_NIST_P256) {
-        curve = TPM_ECC_NIST_P384;
-        nameAlg = TPM_ALG_SHA384;
-        sigHash = TPM_ALG_SHA384;
+    /* Remap a P256 request to the configured default curve. This is a no-op
+     * when the default is P256 (the shipped default), so existing builds are
+     * unchanged. Set WOLFTPM2_ECC_DEFAULT_CURVE to opt in to a stronger
+     * default; it also auto-selects when P256 is compiled out (NO_ECC256). The
+     * name/sig hash is paired to the new curve via TPM2_GetCurveHashAlg. */
+    if (curve == TPM_ECC_NIST_P256 &&
```
aidangarske: **🔴 [High] Default-curve remap overrides explicit P-256 curve requests** · _Logic/API_

The new `WOLFTPM2_ECC_DEFAULT_CURVE` remap lives in the shared `GetKeyTemplateECC` helper, so every caller that explicitly passes `TPM_ECC_NIST_P256` now gets the configured default curve instead. This widens the old `NO_ECC256` fallback into a general configurable remap and changes documented explicit-curve APIs such as `wolfTPM2_GetKeyTemplate_ECC[_ex]`. In a build with `-DWOLFTPM2_ECC_DEFAULT_CURVE=TPM_ECC_NIST_P384` (or P-521), it also affects call paths where P-256 is an explicit semantic choice rather than a library default: `wolfTPM2_CryptoDevCb` converts wolfCrypt's requested curve into `curve_id` and then calls `wolfTPM2_GetKeyTemplate_ECC_ex`, so a P-256 keygen request silently creates a P-384/P-521 TPM key; and EK-index helpers such as `wolfTPM2_GetKeyTemplate_EKIndex(TPM2_NV_EK_ECC_P256)` and the low-range `TPM2_NV_ECC_EK_CERT` case turn a P-256 EK NV index into a P-384/P-521 template, which may no longer match the requested NV certificate/template. Severity differs by mode: the review mode rated this High (api/logic), the review-security mode rated it Medium (HIGH change risk, MEDIUM blast radius); the stricter High is kept.

**Fix:** Do not apply `WOLFTPM2_ECC_DEFAULT_CURVE` inside the generic helper that accepts an explicit `curve` parameter. Apply the macro only at call sites that intentionally use the library default (or add a separate default-template path / explicit sentinel-flag). Keep `wolfTPM2_GetKeyTemplate_ECC[_ex]`, crypto-callback requested curves, and EK-index-derived curves honoring the caller-supplied `curve`. Add tests for `WOLFTPM2_ECC_DEFAULT_CURVE=P384/P521` covering `wolfTPM2_GetKeyTemplate_ECC` and `wolfTPM2_GetKeyTemplate_EKIndex(TPM2_NV_EK_ECC_P256)`.

src/tpm2_wrap.c
```diff
@@ -8224,19 +8224,29 @@ int GetKeyTemplateECC(TPMT_PUBLIC* publicTemplate,
...
+            WOLFTPM2_ECC_DEFAULT_CURVE != TPM_ECC_NIST_P256) {
+        curve = WOLFTPM2_ECC_DEFAULT_CURVE;
+        nameAlg = TPM2_GetCurveHashAlg(curve);
```
aidangarske, line 8237: **🔴 [High] Default-curve remap leaves policy callers using the stale SHA-256 hash** · _bug_

The helper now rewrites `nameAlg` and `sigHash` locally when it remaps P-256, but callers that also build policy data still use their original local hash variables after the helper returns. In a `WOLFTPM2_ECC_DEFAULT_CURVE=TPM_ECC_NIST_P384` build, `GetKeyTemplateECC` can produce a template with `publicTemplate->nameAlg == TPM_ALG_SHA384`, while `wolfTPM2_GetKeyTemplate_EK` still selects the EK auth policy from its unchanged `nameAlg` argument (the `nameAlg == TPM_ALG_SHA256 && !highRange` branch), and the IAK/IDevID helpers still pass the unchanged `hashAlg` into their policy setters. The result is inconsistent templates where the curve/name hash has been upgraded but the auth policy remains SHA-256, so a remapped P-384/P-521 EK template can retain the SHA-256 EK policy. Reported as a bug by the review mode and corroborated by the review-security mode's EK-policy analysis.

**Fix:** If the remap is kept, propagate the effective curve/hash back to callers that derive policy data — e.g. select policy from `publicTemplate->nameAlg` after template creation, return the effective name algorithm to the caller, or remap before calling `GetKeyTemplateECC` so the same effective hash is used everywhere. Ensure EK, IAK, IDevID, and any other policy-building template callers use the effective hash after default-curve selection.

src/tpm2.c
```diff
@@ -7034,6 +7034,20 @@ int TPM2_GetCurveSize(TPM_ECC_CURVE curveID)
     return 0;
 }
 
+TPM_ALG_ID TPM2_GetCurveHashAlg(TPM_ECC_CURVE curveID)
```
aidangarske, line 7037: **🟠 [Medium] No tests cover the new default-curve behavior or hash helper** · _test_

This PR adds a public `TPM2_GetCurveHashAlg` API and a build-time `WOLFTPM2_ECC_DEFAULT_CURVE` behavior change, but no tests were added. Existing tests cover `TPM2_GetCurveSize` and ordinary ECC template creation under the default P-256 configuration, but there is no test for `TPM2_GetCurveHashAlg` and no test build or unit case that defines `WOLFTPM2_ECC_DEFAULT_CURVE` to P-384/P-521 and verifies the effective curve, name hash, signature hash, and policy hash.

**Fix:** Add unit coverage for `TPM2_GetCurveHashAlg` across P-256/P-384/P-521/Brainpool sizes and a template test compiled with `WOLFTPM2_ECC_DEFAULT_CURVE=TPM_ECC_NIST_P384` that verifies the effective template curve/hash fields and that policy-producing helpers stay consistent.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security`**Overall recommendation:** `REQUEST_CHANGES`
**Findings:** 3 total — 3 posted, 0 skipped

### Posted findings
- [High] [review+review-security] **Default-curve remap overrides explicit P-256 curve requests** — `src/tpm2_wrap.c:8237-8242`
- [High] [review+review-security] **Default-curve remap leaves policy callers using the stale SHA-256 hash** — `src/tpm2_wrap.c:8240-8242`
- [Medium] [review] **No tests cover the new default-curve behavior or hash helper** — `src/tpm2.c:7037-7048`

tests/unit_tests.c
```diff
@@ -2538,10 +2538,57 @@ static void test_TPM2_BrainpoolCurveMapping(void)
...
+    /* The storage primary (wolfTPM2_CreateSRK ECC path) follows the default. */
+    AssertIntEQ(wolfTPM2_GetKeyTemplate_ECC_SRK(&t), 0);
+    AssertIntEQ(t.parameters.eccDetail.curveID, WOLFTPM2_ECC_DEFAULT_CURVE);
+    AssertIntEQ(t.nameAlg, TPM2_GetCurveHashAlg(WOLFTPM2_ECC_DEFAULT_CURVE));
+
+    /* The attestation signing key (AIK) follows the default. */
+    AssertIntEQ(wolfTPM2_GetKeyTemplate_ECC_AIK(&t), 0);
+    AssertIntEQ(t.parameters.eccDetail.curveID, WOLFTPM2_ECC_DEFAULT_CURVE);
+
+    /* EK P256 NV index is TCG-fixed and must NOT follow the default curve. */
```
aidangarske, line 2602: **🟠 [Medium] New EK test assertions fail under NO_ECC256 builds**

The new test is guarded only by `#if !defined(WOLFTPM2_NO_WOLFCRYPT) && defined(HAVE_ECC)` and its comment claims "Assertions hold under any configured default." That is not true for `NO_ECC256` builds. The EK assertions assert the P256 EK template keeps its exact curve/hash: `t.parameters.eccDetail.curveID == TPM_ECC_NIST_P256` and `t.nameAlg == TPM_ALG_SHA256`. But `wolfTPM2_GetKeyTemplate_EKIndex(TPM2_NV_EK_ECC_P256)` -> `wolfTPM2_GetKeyTemplate_EK(curveID=P256)` -> `GetKeyTemplateECC(curve=P256)`, and under `#ifdef NO_ECC256` `GetKeyTemplateECC` remaps `curve = WOLFTPM2_ECC_DEFAULT_CURVE` (e.g. P384) and `nameAlg = TPM2_GetCurveHashAlg(curve)` (SHA384). So with `NO_ECC256 && HAVE_ECC384` (DEFAULT=P384), line 2584 fails (P384 != P256) and line 2585 fails (SHA384 != SHA256). This is exactly the behavior the `GetKeyTemplateECC` comment documents ("applies to every caller (including the named EK/SRK/AIK templates)"), so the test contradicts the documented contract. `make check` would fail in a NO_ECC256+HAVE_ECC384/521 configuration.

**Fix:** Guard the EK P256 assertions with `#ifndef NO_ECC256` (the general/SRK/AIK assertions above are fine because they resolve to WOLFTPM2_ECC_DEFAULT_CURVE in both cases). Also fix the comment claim that assertions hold under any default.

tests/unit_tests.c
```diff
@@ -2538,10 +2538,57 @@ static void test_TPM2_BrainpoolCurveMapping(void)
...
+        TPM_ECC_NIST_P256, TPM_ALG_ECDSA), 0);
+    AssertIntEQ(t.parameters.eccDetail.curveID, WOLFTPM2_ECC_DEFAULT_CURVE);
+    AssertIntEQ(t.nameAlg, TPM2_GetCurveHashAlg(WOLFTPM2_ECC_DEFAULT_CURVE));
+
+    /* The storage primary (wolfTPM2_CreateSRK ECC path) follows the default. */
+    AssertIntEQ(wolfTPM2_GetKeyTemplate_ECC_SRK(&t), 0);
+    AssertIntEQ(t.parameters.eccDetail.curveID, WOLFTPM2_ECC_DEFAULT_CURVE);
+    AssertIntEQ(t.nameAlg, TPM2_GetCurveHashAlg(WOLFTPM2_ECC_DEFAULT_CURVE));
+
+    /* The attestation signing key (AIK) follows the default. */
```
aidangarske: **🔵 [Low] AIK template test does not verify the resolved nameAlg/sigHash**

Unlike the general-API and SRK blocks (which assert both curveID and nameAlg), the AIK block only asserts `curveID == WOLFTPM2_ECC_DEFAULT_CURVE`. The AIK path is the one that exercises the `sigHash` argument of `wolfTPM2_ResolveDefaultEccCurve` (it passes `&sigHash`), so a regression in the name/sig-hash pairing for the signing template would go uncaught. The crypto-callback Brainpool hash mapping (now derived from `TPM2_GetCurveHashAlg`) is also new behavior with no template-level assertion.

**Fix:** Add nameAlg and ECDSA sigHash assertions to the AIK block to lock in the curve/hash pairing for the signing template.

src/tpm2_cryptocb.c
```diff
@@ -234,15 +234,13 @@ int wolfTPM2_CryptoDevCb(int devId, wc_CryptoInfo* info, void* ctx)
                     TPMT_PUBLIC publicTemplate;
                     TPMI_ALG_HASH hashAlg;
 
-                    if (curve_id == TPM_ECC_NIST_P521)
-                        hashAlg = TPM_ALG_SHA512;
-                    else if (curve_id == TPM_ECC_NIST_P384)
-                        hashAlg = TPM_ALG_SHA384;
-                    else
-                        hashAlg = TPM_ALG_SHA256;
+                    hashAlg = TPM2_GetCurveHashAlg(curve_id);
```
aidangarske, line 237: **🔵 [Low] Crypto callback changes ECDSA hash for Brainpool curves**

The replaced inline mapping special-cased only P521 (SHA512) and P384 (SHA384) and used SHA256 for everything else, including Brainpool curves. `TPM2_GetCurveHashAlg` now pairs by curve size, so a TPM-backed ECDSA key created via the callback for BP384 uses SHA384 and BP512 uses SHA512 (previously both SHA256). For NIST P256/P384/P521 behavior is identical. The new pairing is the TCG-recommended combination and is almost certainly the intent, but it is a behavior change for an existing code path (key name algorithm and signature hash for Brainpool ECDSA keys) and is worth confirming against interop expectations.

**Fix:** Confirm the Brainpool ECDSA hash upgrade (BP384->SHA384, BP512->SHA512) is intended for the crypto callback path; if so, no change needed.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review`**Overall recommendation:** `COMMENT`
**Findings:** 5 total — 3 posted, 2 skipped

### Posted findings
- [Medium] **New EK test assertions fail under NO_ECC256 builds** — `tests/unit_tests.c:2582-2586`
- [Low] **AIK template test does not verify the resolved nameAlg/sigHash** — `tests/unit_tests.c:2578-2580`
- [Low] **Crypto callback changes ECDSA hash for Brainpool curves** — `src/tpm2_cryptocb.c:237-246`

### Skipped findings
- [Low] ~~New public API and build option not recorded in ChangeLog.md~~
- [Info] ~~Multi-line explanatory comments exceed one-line policy~~

aidangarske (dismissed):

src/tpm2_wrap.c
```diff
@@ -8280,18 +8292,45 @@ int wolfTPM2_GetKeyTemplate_RSA(TPMT_PUBLIC* publicTemplate,
...
+            WOLFTPM2_ECC_DEFAULT_CURVE != TPM_ECC_NIST_P256) {
+        *curve = WOLFTPM2_ECC_DEFAULT_CURVE;
+        if (nameAlg != NULL)
+            *nameAlg = TPM2_GetCurveHashAlg(*curve);
+        if (sigHash != NULL && *sigHash != TPM_ALG_NULL)
+            *sigHash = TPM2_GetCurveHashAlg(*curve);
+    }
+}
+
 int wolfTPM2_GetKeyTemplate_ECC_ex(TPMT_PUBLIC* publicTemplate,
```
aidangarske, line 8310: **🟠 [Medium] wolfTPM2_GetKeyTemplate_ECC[_ex] silently overrides an explicitly requested P256 curve**

These are public APIs whose `curve` parameter previously meant "use exactly this curve." After this PR, passing `TPM_ECC_NIST_P256` no longer guarantees P256 — in an override build it is replaced by `WOLFTPM2_ECC_DEFAULT_CURVE` (along with nameAlg/sigHash). There is no way for a caller to force P256 through this API once a build overrides the default, which makes the API harder to use correctly and is a semantic change to the existing public contract (it is a no-op only in the shipped P256 default). Several example programs pass P256 explicitly through this path; in an override build they will all silently change curve. This is the documented intent, but it conflates "caller asked for P256" with "caller wants the library default," which are not the same thing (this is the root cause of the ECDHE bug above).

**Fix:** Consider applying the default-curve resolution only in the named convenience templates (SRK/AIK) and the no-curve general helper, while leaving the explicit-curve `_ex`/`wolfTPM2_GetKeyTemplate_ECC(curve,...)` paths to honor exactly what is passed. At minimum, prominently document in the header that an explicit P256 request is not guaranteed in override builds and audit all internal callers (ECDH, examples) for protocol-bound curves.

src/tpm2_wrap.c
```diff
@@ -8224,20 +8224,32 @@ int GetKeyTemplateECC(TPMT_PUBLIC* publicTemplate,
...
-    int curveSz = TPM2_GetCurveSize(curve);
+    int curveSz;
 
-    if (publicTemplate == NULL || curveSz == 0)
+    if (publicTemplate == NULL)
         return BAD_FUNC_ARG;
 
-#if defined(NO_ECC256) && defined(HAVE_ECC384) && ECC_MIN_KEY_SZ <= 384
-    /* make sure we use a curve that is enabled */
+#ifdef NO_ECC256
```
aidangarske, line 8232: **🟠 [Medium] EK ECC template under NO_ECC256 gets substituted curve/nameAlg but mismatched auth policy**

The PR broadens the NO_ECC256 substitution guard from `#if defined(NO_ECC256) && defined(HAVE_ECC384) && ECC_MIN_KEY_SZ <= 384` to plain `#ifdef NO_ECC256`, so the EK ECC P256 template now also gets curve+nameAlg substituted for more configs (e.g. HAVE_ECC521 only). Inside `GetKeyTemplateECC`, `nameAlg` is taken by value and rewritten to the paired hash (e.g. SHA512), so `publicTemplate->nameAlg` becomes SHA512. But the caller `wolfTPM2_GetKeyTemplate_EK` selects the EK auth policy from its own local `nameAlg` (still SHA256), so it copies `TPM_20_EK_AUTH_POLICY` (the SHA256 policy). The resulting EK template has nameAlg != the auth policy it was built for. The new unit test sidesteps this by guarding the EK assertions with `#ifndef NO_ECC256`. The mismatch partially pre-existed for the HAVE_ECC384 case, but this PR widens the set of builds that hit it.

**Fix:** Either keep the EK templates out of the availability substitution (resolve the EK curve at the EK layer so nameAlg and authPolicy stay consistent), or have the EK path recompute its auth policy from the actual `publicTemplate->nameAlg` after `GetKeyTemplateECC` returns.

tests/unit_tests.c
```diff
@@ -2538,10 +2538,67 @@ static void test_TPM2_BrainpoolCurveMapping(void)
     AssertIntEQ(TPM2_GetCurveSize(TPM_ECC_BP_P384_R1), 48);
     AssertIntEQ(TPM2_GetCurveSize(TPM_ECC_BP_P512_R1), 64);
 
+    /* TPM2_GetCurveHashAlg pairs the digest strength to the curve size,
+     * per TCG recommended combinations. */
+    AssertIntEQ(TPM2_GetCurveHashAlg(TPM_ECC_NIST_P256), TPM_ALG_SHA256);
+    AssertIntEQ(TPM2_GetCurveHashAlg(TPM_ECC_NIST_P384), TPM_ALG_SHA384);
+    AssertIntEQ(TPM2_GetCurveHashAlg(TPM_ECC_NIST_P521), TPM_ALG_SHA512);
```
aidangarske, line 2553: **🟠 [Medium] Default-curve upgrade path only verified incidentally; no coverage of the non-P256 case in default CI**

All assertions in the new test compare against `WOLFTPM2_ECC_DEFAULT_CURVE` / `TPM2_GetCurveHashAlg(WOLFTPM2_ECC_DEFAULT_CURVE)`. In the shipped default (P256) build, every assertion reduces to `curveID == P256` / `nameAlg == SHA256`, i.e. it confirms only the no-op path. The actual upgrade behavior (P256 request -> P384 curve + paired SHA384 hash) is never exercised unless someone compiles with `-DWOLFTPM2_ECC_DEFAULT_CURVE=...` or `NO_ECC256`, neither of which is the standard unit-test config. The feature's core behavior therefore has no default-CI coverage, and the ECDHE regression above would not be caught by this suite.

**Fix:** Add a test that forces a non-P256 resolution independent of build flags — e.g. assert the resolver maps P256->expected for a hard-coded override, or add a CI matrix entry building with `-DWOLFTPM2_ECC_DEFAULT_CURVE=TPM_ECC_NIST_P384`. Also add a test asserting `wolfTPM2_ECDHGenKey`/ECDH templates keep the requested curve under an override.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review`**Overall recommendation:** `REQUEST_CHANGES`
**Findings:** 6 total — 6 posted, 0 skipped

### Posted findings
- [High] **ECDHE ephemeral key curve silently remapped by WOLFTPM2_ECC_DEFAULT_CURVE, breaking shared-secret in override builds** — `src/tpm2_wrap.c:6183 (and src/tpm2_cryptocb.c:290)`
- [Medium] **wolfTPM2_GetKeyTemplate_ECC[_ex] silently overrides an explicitly requested P256 curve** — `src/tpm2_wrap.c:8317-8335`
- [Medium] **EK ECC template under NO_ECC256 gets substituted curve/nameAlg but mismatched auth policy** — `src/tpm2_wrap.c:8232-8245, 8443-8453`
- [Medium] **Default-curve upgrade path only verified incidentally; no coverage of the non-P256 case in default CI** — `tests/unit_tests.c:2545-2599`

aidangarske (approved):

PR #521, opened 2026-06-08
Remove flaky SPDM HW tests in CI

aidangarske:
# Description

- Depricate HW CI tests covered by fwtpm CI tests


dgarske (approved):
HW is offline, so let's remove this for now.

PR #531, opened 2026-06-18
PQC parameter encryption for examples, plus bound-session key fixes

dgarske:
# PQC parameter encryption for examples, plus parameter-encryption fixes

Adds post-quantum (ML-DSA / ML-KEM) parameter-encryption support to the examples, and fixes several parameter-encryption bugs in the client and the firmware TPM that were uncovered along the way.

## Features
- PQC-keyed parameter-encryption sessions: `-mlkem[=512|768|1024]` (salt key) and `-mldsa[=44|65|87]` (bind key), plus `-paramkey=` in keygen. Wired through keygen, wrap_test, pcr/quote, nvram/store and nvram/counter, with a shared parser in tpm_test_keys. ML-DSA is sign-only, so the helper adds a transient SRK salt for confidentiality while the ML-DSA key supplies the binding.
- `create_primary` gains `-mldsa[=44|65|87]` for an ML-DSA primary.
- Selecting a PQC param-enc key without `-aes`/`-xor` now defaults the session cipher to AES-CFB instead of silently doing nothing.

## Fixes
- Bound-session session-key derivation: compute `KDFa(bindAuth || salt)` (TPM 2.0 Part 1 Sec.19.6.8) on both the client and the firmware TPM. The firmware TPM had the inputs reversed (`salt || bindAuth`), which broke salt+bind sessions, and a session bound to an EmptyAuth entity previously left the session key empty.
- Parameter-encryption key authValue: the param-enc key is `sessionKey || authValue` even when the session is bound to the entity it authorizes (unlike the command HMAC, which excludes it). The firmware TPM omitted the authValue and silently corrupted parameter-encrypted data for bound-own-entity flows (e.g. `examples/nvram/extend.c`); now fixed on both sides (`TPM2_ParamEncBindKey` / `FwBuildParamEncKey`, with `bindName` on `TPM2_AUTH_SESSION` tracking the bind entity).
- `TPM2_CreateLoaded`: set the response `outHandleCnt`, which was mis-locating the response parameters (and rpHash) and rejecting the reply with `TPM_RC_HMAC` whenever it ran under a parameter-encryption session.
- PQC option parsing: alg/parameter-set outputs are committed only after validation, and an unsupported set (e.g. `-mlkem=999`) is now a fatal usage error in every example instead of silently dropping parameter encryption.
- Single-family builds: the `-mlkem` and `-mldsa` help lines are each guarded by their own family macro, so an ML-KEM-only or ML-DSA-only build only advertises what it supports.

## Tests / testing
- New `unit.test` regressions: bound-EmptyAuth param-enc; `TPM2_CreateLoaded` under param-enc (fails pre-fix with `TPM_RC_HMAC`); and a bound-own-entity param-enc test that writes a POLICYWRITE NV index under a bound policy session and reads the data back, proving the param-enc key is correct (catches a wrong key on either the client or the firmware TPM).
- `run_examples.sh` exercises the ML-KEM-salt and ML-DSA-bind param-enc paths across child-create, attestation and NV.
- Verified against both the firmware TPM and the IBM SW TPM (reference); ML-DSA-only and ML-KEM-only builds link cleanly.

### DIFF
--- a/examples/keygen/create_primary.c
+++ b/examples/keygen/create_primary.c
@@ -90,9 +90,20 @@ static void usage(void)
 #ifdef WOLFTPM_MLDSA
 static int mldsaParamSet(const char* optVal, TPMI_MLDSA_PARAMETER_SET* ps)
 {
-    int n = XATOI(optVal);
+    int n;
+    const char* p;
+
+    if (optVal[0] == '\0') { /* missing or empty suffix, use default */
+        *ps = TPM_MLDSA_65;
+        return TPM_RC_SUCCESS;
+    }
+    /* reject non-digit input (e.g. -mldsa=abc) before XATOI */
+    for (p = optVal; *p != '\0'; p++) {
+        if (*p < '0' || *p > '9')
+            return TPM_RC_FAILURE;
+    }
+    n = XATOI(optVal);
     switch (n) {
-        case 0:  /* missing or empty suffix, use default */
         case 65: *ps = TPM_MLDSA_65; return TPM_RC_SUCCESS;
         case 44: *ps = TPM_MLDSA_44; return TPM_RC_SUCCESS;
         case 87: *ps = TPM_MLDSA_87; return TPM_RC_SUCCESS;
@@ -147,7 +158,7 @@ int TPM2_CreatePrimaryKey_Example(void* userCtx, int argc, char *argv[])
                 argv[argc-1] + 7 : "";
             if (mldsaParamSet(optVal, &mldsaPs) != TPM_RC_SUCCESS) {
                 usage();
-                return 0;
+                return -1;
             }
             alg = TPM_ALG_MLDSA;
         }
--- a/examples/keygen/create_primary.c
+++ b/examples/keygen/create_primary.c
@@ -47,6 +47,10 @@ static void usage(void)
     printf("Primary Key Type:\n");
     printf("\t-rsa: Use RSA for asymmetric key generation (DEFAULT)\n");
     printf("\t-ecc: Use ECC for asymmetric key generation \n");
+#ifdef WOLFTPM_MLDSA
+    printf("\t-mldsa[=44|65|87]: Use ML-DSA for primary key generation "
+           "(v1.85, default 65)\n");
+#endif
     printf("Hierarchy:\n");
     printf("\t-oh: Create keys under the Owner Hierarchy (DEFAULT)\n");
     printf("\t-eh: Create keys under the Endorsement Hierarchy\n");
@@ -77,7 +81,25 @@ static void usage(void)
     printf("\t* Create SRK used by wolfTPM:\n");
     printf("\t\tcreate_primary -rsa -oh -auth=ThisIsMyStorageKeyAuth "
                                        "-store=0x81000200\n");
+#ifdef WOLFTPM_MLDSA
+    printf("\t* Create an ML-DSA-65 primary key:\n");
+    printf("\t\tcreate_primary -mldsa -oh\n");
+#endif
 }
+
+#ifdef WOLFTPM_MLDSA
+static int mldsaParamSet(const char* optVal, TPMI_MLDSA_PARAMETER_SET* ps)
+{
+    int n = XATOI(optVal);
+    switch (n) {
+        case 0:  /* missing or empty suffix, use default */
+        case 65: *ps = TPM_MLDSA_65; return TPM_RC_SUCCESS;
+        case 44: *ps = TPM_MLDSA_44; return TPM_RC_SUCCESS;
+        case 87: *ps = TPM_MLDSA_87; return TPM_RC_SUCCESS;
+        default: return TPM_RC_FAILURE;
+    }
+}
+#endif /* WOLFTPM_MLDSA */
 
 int TPM2_CreatePrimaryKey_Example(void* userCtx, int argc, char *argv[])
 {
@@ -98,6 +120,9 @@ int TPM2_CreatePrimaryKey_Example(void* userCtx, int argc, char *argv[])
 #ifdef WOLFTPM_PROVISIONING
     int useIAKTemplate = 0, useIDevIDTemplate = 0;
 #endif
+#ifdef WOLFTPM_MLDSA
+    TPMI_MLDSA_PARAMETER_SET mldsaPs = TPM_MLDSA_65; /* default param set */
+#endif
 
     if (argc >= 2) {
         if (XSTRCMP(argv[1], "-?") == 0 ||
@@ -114,6 +139,19 @@ int TPM2_CreatePrimaryKey_Example(void* userCtx, int argc, char *argv[])
         else if (XSTRCMP(argv[argc-1], "-ecc") == 0) {
             alg = TPM_ALG_ECC;
         }
+#ifdef WOLFTPM_MLDSA
+        else if (XSTRCMP(argv[argc-1], "-mldsa") == 0 ||
+                 XSTRNCMP(argv[argc-1], "-mldsa=",
+                     XSTRLEN("-mldsa=")) == 0) {
+            const char* optVal = (argv[argc-1][6] == '=') ?
+                argv[argc-1] + 7 : "";
+            if (mldsaParamSet(optVal, &mldsaPs) != TPM_RC_SUCCESS) {
+                usage();
+                return 0;
+            }
+            alg = TPM_ALG_MLDSA;
+        }
+#endif
         else if (XSTRCMP(argv[argc-1], "-aes") == 0) {
             paramEncAlg = TPM_ALG_CFB;
         }
@@ -221,6 +259,17 @@ int TPM2_CreatePrimaryKey_Example(void* userCtx, int argc, char *argv[])
         else
             rc = wolfTPM2_GetKeyTemplate_ECC_SRK(&publicTemplate);
     }
+#ifdef WOLFTPM_MLDSA
+    else if (alg == TPM_ALG_MLDSA) {
+        /* ML-DSA is sign-only and has no AIK/IAK/IDevID variant; the
+         * wrapper enforces TPMA_OBJECT_sign and clears decrypt. */
+        rc = wolfTPM2_GetKeyTemplate_MLDSA(&publicTemplate,
+            TPMA_OBJECT_sign | TPMA_OBJECT_fixedTPM |
+            TPMA_OBJECT_fixedParent | TPMA_OBJECT_sensitiveDataOrigin |
+            TPMA_OBJECT_userWithAuth | TPMA_OBJECT_noDA,
+            mldsaPs, 0 /* allowExternalMu: 0 avoids TPM_RC_EXT_MU */);
+    }
+#endif
     else {
         rc = BAD_FUNC_ARG;
     }
--- a/examples/tpm_test_keys.c
+++ b/examples/tpm_test_keys.c
@@ -430,6 +430,168 @@ int getPrimaryStoragekey(WOLFTPM2_DEV* pDev, WOLFTPM2_KEY* pStorageKey,
     return rc;
 }
 
+#if defined(WOLFTPM_MLDSA) || defined(WOLFTPM_MLKEM)
+int getPrimaryParamEncKey(WOLFTPM2_DEV* pDev, WOLFTPM2_SESSION* session,
+    WOLFTPM2_KEY* pqcKey, TPM_ALG_ID pqcAlg, int paramSet, int paramEncAlg)
+{
+    int rc;
+    TPMT_PUBLIC publicTemplate;
+#ifdef WOLFTPM_MLDSA
+    WOLFTPM2_KEY saltKey; /* transient SRK salt for the ML-DSA bound session */
+#if !defined(NO_RSA)
+    TPM_ALG_ID saltAlg = TPM_ALG_RSA;
+#else
+    TPM_ALG_ID saltAlg = TPM_ALG_ECC;
+#endif
+#endif
+    const char* pqcAuth = "pqcParamEnc"; /* primary userAuth; the bind auth for
+                                          * ML-DSA, immaterial for the ML-KEM
+                                          * salt key */
+
+    if (pDev == NULL || session == NULL || pqcKey == NULL)
+        return BAD_FUNC_ARG;
+
+    XMEMSET(pqcKey, 0, sizeof(*pqcKey));
+    XMEMSET(&publicTemplate, 0, sizeof(publicTemplate));
+#ifdef WOLFTPM_MLDSA
+    XMEMSET(&saltKey, 0, sizeof(saltKey));
+#endif
+
+    if (pqcAlg == TPM_ALG_MLKEM) {
+    #ifdef WOLFTPM_MLKEM
+        /* ML-KEM salt key: restricted decryption, which requires a symmetric
+         * definition (a TPM rejects a restricted key without one). */
+        rc = wolfTPM2_GetKeyTemplate_MLKEM(&publicTemplate,
+            TPMA_OBJECT_decrypt | TPMA_OBJECT_restricted |
+            TPMA_OBJECT_fixedTPM | TPMA_OBJECT_fixedParent |
+            TPMA_OBJECT_sensitiveDataOrigin | TPMA_OBJECT_userWithAuth |
+            TPMA_OBJECT_noDA, (TPMI_MLKEM_PARAMETER_SET)paramSet);
+        if (rc == TPM_RC_SUCCESS) {
+            publicTemplate.parameters.mlkemDetail.symmetric.algorithm =
+                TPM_ALG_AES;
+            publicTemplate.parameters.mlkemDetail.symmetric.keyBits.aes = 128;
+            publicTemplate.parameters.mlkemDetail.symmetric.mode.aes =
+                TPM_ALG_CFB;
+        }
+    #else
+        (void)paramSet;
+        return NOT_COMPILED_IN;
+    #endif
+    }
+    else if (pqcAlg == TPM_ALG_MLDSA) {
+    #ifdef WOLFTPM_MLDSA
+        /* ML-DSA is sign-only; it can only be a session bind entity. */
+        rc = wolfTPM2_GetKeyTemplate_MLDSA(&publicTemplate,
+            TPMA_OBJECT_sign | TPMA_OBJECT_fixedTPM |
+            TPMA_OBJECT_fixedParent | TPMA_OBJECT_sensitiveDataOrigin |
+            TPMA_OBJECT_userWithAuth | TPMA_OBJECT_noDA,
+            (TPMI_MLDSA_PARAMETER_SET)paramSet, 0);
+    #else
+        (void)paramSet;
+        return NOT_COMPILED_IN;
+    #endif
+    }
+    else {
+        return BAD_FUNC_ARG;
+    }
+    if (rc != TPM_RC_SUCCESS)
+        return rc;
+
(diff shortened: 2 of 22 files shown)


examples/keygen/create_primary.c
```diff
@@ -47,6 +47,10 @@ static void usage(void)
     printf("Primary Key Type:\n");
     printf("\t-rsa: Use RSA for asymmetric key generation (DEFAULT)\n");
     printf("\t-ecc: Use ECC for asymmetric key generation \n");
+#ifdef WOLFTPM_PQC
```
aidangarske: **🟠 [Medium] create_primary ML-DSA support guarded by WOLFTPM_PQC but links against the WOLFTPM_MLDSA-only symbol…** · _Logic_

Both the review and review-security scans flagged that the new ML-DSA primary support in create_primary.c is wrapped in `#ifdef WOLFTPM_PQC`, while the function it calls, wolfTPM2_GetKeyTemplate_MLDSA(), is only compiled under `#ifdef WOLFTPM_MLDSA` (src/tpm2_wrap.c:6138) — only its prototype lives under WOLFTPM_PQC (wolftpm/tpm2_wrap.h). Per wolftpm/tpm2_types.h the family macro is opt-out: a WOLFTPM_PQC build with WOLFTPM_NO_MLDSA (ML-KEM only) keeps WOLFTPM_PQC defined but leaves WOLFTPM_MLDSA undefined. In that configuration create_primary.c compiles the call (the enums TPM_ALG_MLDSA/TPM_MLDSA_65 are under WOLFTPM_PQC) but fails to link against the absent definition, producing an undefined-reference error. The PR's own helper getPrimaryParamEncKey() (examples/tpm_test_keys.c:437-473) already uses the precise WOLFTPM_MLDSA/WOLFTPM_MLKEM guards with NOT_COMPILED_IN fallbacks, so this is an internal inconsistency within the same PR. It mirrors the pre-existing keygen.c convention, but the create_primary.c ML-DSA path is newly introduced here. Severity views differ: the review scan rated this Medium (action SUGGEST); the review-security scan rated it Low (Medium confidence). The stricter Medium is retained.

**Fix:** Guard the create_primary.c ML-DSA option/template code with `#ifdef WOLFTPM_MLDSA` (matching the implementation guard and the new getPrimaryParamEncKey() helper) instead of `WOLFTPM_PQC`, so a WOLFTPM_NO_MLDSA build excludes it cleanly.

examples/tpm_test_keys.c
```diff
@@ -418,6 +418,145 @@ int getPrimaryStoragekey(WOLFTPM2_DEV* pDev, WOLFTPM2_KEY* pStorageKey,
...
+    return rc;
+}
+
+int parsePqcParamSet(const char* val, TPM_ALG_ID* alg, int* paramSet)
+{
+    int n;
+
+    if (val == NULL || alg == NULL || paramSet == NULL)
+        return 0;
+#ifdef WOLFTPM_MLKEM
```
aidangarske, line 545: **🔵 [Low] parsePqcParamSet treats a trailing '=' with empty suffix as an error instead of defaulting** · _bug_

For input `mlkem=` (or `mldsa=`) the code computes `n = (val[5] == '=') ? XATOI(&val[6]) : 768;`. With an empty suffix XATOI("") returns 0, which falls through to the else branch and prints `Invalid ML-KEM parameter set: 0` and returns failure, rather than falling back to the documented default (768 / 65). This is inconsistent with create_primary.c's mldsaParamSet(), which explicitly treats case 0 (empty suffix) as the default. Bare mlkem/mldsa (no =) correctly defaults, so only the malformed mlkem= form behaves surprisingly.

**Fix:** Treat n==0 (empty suffix) as the default, e.g. `if (n == 0 || n == 768) *paramSet = TPM_MLKEM_768;` matching mldsaParamSet, or reject the empty-suffix form earlier with a clearer message.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security`**Overall recommendation:** `COMMENT`
**Findings:** 6 total — 4 posted, 2 skipped
**2 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Medium] [review+review-security] **create_primary ML-DSA support guarded by WOLFTPM_PQC but links against the WOLFTPM_MLDSA-only symbol wolfTPM2_GetKeyTemplate_MLDSA** — `examples/keygen/create_primary.c:90-102,142-154,262-270`
- [Medium] [review] **No automated regression test added for the bound-session / EmptyAuth-bind session-key fixes** — `src/tpm2_param_enc.c:201-227, src/tpm2_wrap.c:2627, src/fwtpm/fwtpm_command.c:8306-8325`
- [Low] [review] **parsePqcParamSet treats a trailing '=' with empty suffix as an error instead of defaulting** — `examples/tpm_test_keys.c:512-548`
- [Low] [review] **Selecting -mlkem/-mldsa (or -paramkey=) without -aes/-xor silently does nothing** — `examples/nvram/counter.c:152-175, examples/nvram/store.c:172, examples/pcr/quote.c:218, examples/wrap/wrap_test.c:287, examples/keygen/keygen.c:386`

### Skipped findings
- [Info] ~~Parameter-encryption bind-key selection now depends on session-name ordering~~
- [Info] ~~TPM2_AUTH_SESSION grows by a TPM2B_NAME (bindName) — ABI/size change~~

examples/tpm_test_keys.c
```diff
@@ -418,6 +418,149 @@ int getPrimaryStoragekey(WOLFTPM2_DEV* pDev, WOLFTPM2_KEY* pStorageKey,
     return rc;
 }
 
+#if defined(WOLFTPM_MLDSA) || defined(WOLFTPM_MLKEM)
+int getPrimaryParamEncKey(WOLFTPM2_DEV* pDev, WOLFTPM2_SESSION* session,
+    WOLFTPM2_KEY* pqcKey, TPM_ALG_ID pqcAlg, int paramSet, int paramEncAlg)
+{
+    int rc;
+    TPMT_PUBLIC publicTemplate;
+    /* Auth value for the PQC key; gives the bound (ML-DSA) session extra
```
aidangarske: **🟠 [Medium] ML-DSA bind parameter encryption is derived from a public authValue** · _Security_

The new ML-DSA path creates a transient bind key with the fixed authValue `pqcParamEnc`, then starts an unsalted bound session because a signing-only ML-DSA key cannot provide an encrypted salt. In `wolfTPM2_StartSession`, that bound-only session key is derived from the bind authValue plus public nonces. Since the authValue is hard-coded in the example and is also supplied to `CreatePrimary` in cleartext before this parameter-encryption session exists, a passive observer can derive the same session key and decrypt the parameters protected by the advertised `-mldsa` parameter-encryption mode. The ML-KEM salt path is unaffected because it uses an asymmetric encrypted salt.

**Fix:** Do not present a freshly-created ML-DSA bind key with fixed/public auth as a confidentiality mechanism. Either clearly limit the `-mldsa` example to non-confidential binding semantics, or require a pre-provisioned bind entity whose authValue is secret and was not sent in cleartext on the same channel. If creating the bind key in the helper, generate high-entropy auth, protect its creation under an already confidential session, then zeroize it after session setup.

examples/tpm_test_keys.c
```diff
@@ -418,6 +418,149 @@ int getPrimaryStoragekey(WOLFTPM2_DEV* pDev, WOLFTPM2_KEY* pStorageKey,
...
+    if (rc != TPM_RC_SUCCESS) {
+        printf("PQC param-enc StartSession failed 0x%x: %s\n", rc,
+            TPM2_GetRCString(rc));
+        wolfTPM2_UnloadHandle(pDev, &pqcKey->handle);
+        XMEMSET(pqcKey, 0, sizeof(*pqcKey));
+    }
+    return rc;
+}
+
+int parsePqcParamSet(const char* val, TPM_ALG_ID* alg, int* paramSet)
```
aidangarske, line 538: **🟠 [Medium] Invalid PQC parameter-set parse leaves parser outputs armed** · _Logic / State Management_

`parsePqcParamSet()` writes `*alg` (and may leave `*paramSet` stale) before validating the parameter-set value. On a recognized but invalid option such as `-mlkem=999` or `-mldsa=foo`, it prints an error and returns 0, but `pqcParamEncAlg` has already been changed from `TPM_ALG_NULL`. The new callers in `wrap_test`, `nvram/store`, `nvram/counter`, and `pcr/quote` treat return 0 as an unrecognized-option warning and continue; their later `pqcParamEncAlg != TPM_ALG_NULL` check then enables AES-CFB and attempts to create a PQC parameter-encryption key with parameter set 0 (`TPM_MLKEM_NONE`/`TPM_MLDSA_NONE`) or a stale value, turning a parse error into a low-level TPM failure or wrong PQC key selection. Severity reported as Medium by `review` and Low by `review-security` and `bugs`; stricter view kept.

**Fix:** Parse into local temporaries and assign `*alg`/`*paramSet` only after validation succeeds. Distinguish no-match from invalid-match so callers can reject `-mlkem=999`/`-mldsa=foo` as fatal instead of treating them as ordinary unknown options.

examples/wrap/wrap_test.c
```diff
@@ -265,8 +297,19 @@ int TPM2_Wrapper_TestArgs(void* userCtx, int argc, char *argv[])
     #ifdef NO_RSA
         bindKey = NULL; /* cannot bind to key without RSA enabled */
     #endif
-        rc = wolfTPM2_StartSession(&dev, &tpmSession, bindKey, NULL,
-            TPM_SE_HMAC, paramEncAlg);
+    #if defined(WOLFTPM_MLKEM) || defined(WOLFTPM_MLDSA)
```
aidangarske, line 310: **🟠 [Medium] wrap_test PQC parameter-encryption session is overwritten before use** · _State Management_

The new `-mlkem`/`-mldsa` path creates a PQC-backed parameter-encryption session and installs it in auth slot 0. The wrapper calls immediately afterward (e.g. `wolfTPM2_CreateLoadedKey`) call `wolfTPM2_SetAuthHandle(dev, 0, parent)` internally, resetting slot 0 to password auth before the command is marshaled. A run like `./examples/wrap/wrap_test -aes -mlkem=768` therefore creates a PQC session but does not use it for the advertised parameter encryption; the command proceeds with parent password auth. The other new PQC examples (`keygen`, `quote`, `store`, `counter`) avoid this by installing the secondary encrypt/decrypt session in slot 1.

**Fix:** Install the PQC parameter-encryption session in slot 1, matching `keygen`, `quote`, `store`, and `counter`, so slot 0 stays available for the parent authorization set by wrapper helpers and the PQC session survives into the commands meant to exercise it (cover both the RSA and ECC halves consistently).

tests/unit_tests.c
```diff
@@ -661,6 +661,89 @@ static void test_wolfTPM2_StartSession_SaltedEncryptAttrs(void)
...
+    /* Storage root key with an EmptyAuth (auth NULL, authSz 0). */
+    rc = wolfTPM2_CreateSRK(&dev, &srk, srkAlg, NULL, 0);
+    if (rc != 0) {
+        /* Environmental (TPM busy / unsupported). Treat as skip. */
+        wolfTPM2_Cleanup(&dev);
+        printf("Test TPM Wrapper:\tBound EmptyAuth param-enc:\tSkipped\n");
+        return;
+    }
+
+    /* Bind an HMAC session to the EmptyAuth SRK with AES-CFB param enc. */
```
aidangarske, line 777: **🟠 [Medium] Bound EmptyAuth regression test does not exercise the bound HMAC session** · _Missing Tests / Logic_

The new regression test installs the bound AES-CFB HMAC parameter-encryption session in auth slot 0, then calls `wolfTPM2_CreateAndLoadKey()`. That helper calls `wolfTPM2_CreateKey()` (and `wolfTPM2_LoadKey()`), which call `wolfTPM2_SetAuthHandle(dev, 0, parent)` and overwrite slot 0 with password auth before `TPM2_Create` is sent. As a result, the encrypted command is not authorized by the bound EmptyAuth session the test claims to validate, so the test can pass even if the PR's `wolfTPM2_StartSession` EmptyAuth KDFa change or `TPM2_ParamEncBindKey` behavior is still broken. Severity reported as Medium by `review` and Low by `review-security` and `bugs`; stricter view kept.

**Fix:** Build the regression around a command path that preserves the HMAC session as the authorizing session — use the native `TPM2_Create` setup directly after setting the bound session's authorized handle name, or use slot 1 if only the bound sessionKey derivation is to be exercised. Add an assertion that the command actually uses the bound session and fails on the pre-fix EmptyAuth behavior.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security + bugs`**Overall recommendation:** `COMMENT`
**Findings:** 4 total — 4 posted, 0 skipped
**4 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Medium] [review-security] **ML-DSA bind parameter encryption is derived from a public authValue** — `examples/tpm_test_keys.c:427-495`
- [Medium] [review+review-security+bugs] **Invalid PQC parameter-set parse leaves parser outputs armed** — `examples/tpm_test_keys.c:506-548`
- [Medium] [review+bugs] **wrap_test PQC parameter-encryption session is overwritten before use** — `examples/wrap/wrap_test.c:300-319`
- [Medium] [review+review-security+bugs] **Bound EmptyAuth regression test does not exercise the bound HMAC session** — `tests/unit_tests.c:709-734`

src/fwtpm/fwtpm_command.c
```diff
@@ -196,58 +196,90 @@ static int FwGetPcrBankIndex(UINT16 hashAlg)
...
+            sess->bindAuth.size > 0 && sz + sess->bindAuth.size <= keyBufMax) {
+        XMEMCPY(keyBuf + sz, sess->bindAuth.buffer, sess->bindAuth.size);
+        sz += sess->bindAuth.size;
+    }
+    *keyBufSz = sz;
+    return 0;
+}
+
 /* Decrypt first TPM2B parameter of incoming command (param encryption) */
 static int FwParamDecryptCmd(FWTPM_CTX* ctx, FWTPM_Session* sess,
```
aidangarske, line 224: **🟠 [Medium] fwTPM parameter encryption helpers can return an uninitialized status** · _Uninitialized variable used before assignment_

The PR changed both fwTPM parameter-encryption helpers from `int rc = 0` to `int rc` while keeping the dispatch limited to XOR and AES-CFB and adding no final `else`. `FwCmd_StartAuthSession` stores any non-NULL symmetric definition, and `FWTPM_ProcessCommand` treats any non-NULL symmetric algorithm as an encryption session, so a caller can start a session with an unsupported non-NULL symmetric algorithm or AES mode (e.g. AES-CBC) and use it on a parameter-encryption-capable command. Before this change the path deterministically returned success; after it the helper returns indeterminate stack garbage, producing nondeterministic success/failure in the auth/parameter-encryption path.

**Fix:** Initialize `rc` to a failure code such as `TPM_RC_FAILURE` in both helpers and add an explicit unsupported-symmetric `else` (e.g. `rc = TPM_RC_SYMMETRIC`). Prefer also rejecting unsupported `TPMT_SYM_DEF` values in `FwCmd_StartAuthSession` so invalid sessions fail before command processing.

examples/keygen/create_primary.c
```diff
@@ -77,7 +81,25 @@ static void usage(void)
...
     printf("\t\tcreate_primary -rsa -oh -auth=ThisIsMyStorageKeyAuth "
                                        "-store=0x81000200\n");
+#ifdef WOLFTPM_MLDSA
+    printf("\t* Create an ML-DSA-65 primary key:\n");
+    printf("\t\tcreate_primary -mldsa -oh\n");
+#endif
+}
+
+#ifdef WOLFTPM_MLDSA
+static int mldsaParamSet(const char* optVal, TPMI_MLDSA_PARAMETER_SET* ps)
```
aidangarske, line 91: **🟠 [Medium] create_primary ML-DSA parser treats explicit invalid suffixes as the default** · _bug_

The new `create_primary -mldsa` parser maps any suffix that `XATOI()` converts to zero to the default ML-DSA-65 set. `XATOI` is `atoi`, so `-mldsa=0` and non-numeric typos like `-mldsa=abc` silently create an ML-DSA-65 primary instead of rejecting the unsupported parameter set, despite the comment saying only a missing or empty suffix should use the default. Severity views differ (review: Medium; bugs: Low) — keeping the stricter Medium.

**Fix:** Only use the default for an actually empty suffix, and reject non-digit input before calling `XATOI`. Add usage-error tests for `-mldsa=0` and `-mldsa=abc`.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security + bugs`**Overall recommendation:** `COMMENT`
**Findings:** 4 total — 4 posted, 0 skipped
**2 finding(s) posted as inline comments** (see file-level comments below)
**2 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Medium] [review-security+bugs] **fwTPM parameter encryption helpers can return an uninitialized status** — `src/fwtpm/fwtpm_command.c:221-284`
- [Medium] [review+bugs] **create_primary ML-DSA parser treats explicit invalid suffixes as the default** — `examples/keygen/create_primary.c:91-100`

#### Invalid PQC option can be ignored after a previous valid option
**File:** `examples/wrap/wrap_test.c:181-189; examples/nvram/counter.c:124-132; examples/nvram/store.c:138-146; examples/pcr/quote.c:143-151`
**Function:** `TPM2_Wrapper_TestArgs / TPM2_NVRAM_Counter_Example / TPM2_NVRAM_Store_Example / TPM2_PCR_Quote_Test`
**Severity:** Medium
**Category:** bug

All three scan modes flagged this. The new callers use `pqcParamEncAlg == TPM_ALG_NULL` to decide whether `parsePqcParamEncArg()` reported an invalid PQC parameter set. That only works before any valid PQC option has been parsed. These examples parse argv from right to left, so a command like `wrap_test -mldsa=999 -mlkem=768` first sets `pqcParamEncAlg` to ML-KEM, then `parsePqcParamEncArg()` returns -1 for `-mldsa=999` without clearing the already-committed output. The invalid option is then silently accepted instead of being the fatal usage error this PR intends. Severity views differ (review: Medium; review-security and bugs: Low) — keeping the stricter Medium.

**Recommendation:** Store the parser return value and branch on it directly: `int pqcRc = parsePqcParamEncArg(...); if (pqcRc < 0) { usage(); return -1; } else if (pqcRc > 0) { /* valid PQC option */ }`. Add a regression test with one valid and one invalid PQC option in both argument orders.

_Referenced code: `examples/wrap/wrap_test.c:181-189; examples/nvram/counter.c:124-132; examples/nvram/store.c:138-146; examples/pcr/quote.c:143-149` (7 lines)_

---

#### Zephyr wrap_test target misses the new helper source dependency
**File:** `examples/wrap/wrap_test.c:38,181-189,311-312; examples/wrap/include.am:13-14`
**Function:** `TPM2_Wrapper_TestArgs`
**Severity:** Medium
**Category:** Regression

The PR makes `wrap_test.c` include `examples/tpm_test_keys.h` and call `parsePqcParamEncArg()` / `getPrimaryParamEncKey()` under `WOLFTPM_MLKEM || WOLFTPM_MLDSA`. Automake and the top-level CMake helper were updated to compile/link `examples/tpm_test_keys.c`, but the Zephyr sample `zephyr/samples/wolftpm_wrap_test/CMakeLists.txt` still builds only `examples/wrap/wrap_test.c`. A PQC-enabled Zephyr wrap_test build will have undefined references to the new helper functions.

**Recommendation:** Update the Zephyr wrap_test sample to compile `examples/tpm_test_keys.c` when the same PQC macros can enable these calls, or move the helper linkage into shared Zephyr app sources so the sample stays aligned with the other build systems.

_Referenced code: `examples/wrap/wrap_test.c:38,181-189,311-312; examples/wrap/include.am:13-19` (7 lines)_

---

aidangarske (approved):

PR #536, opened 2026-06-23
feat: add make sbom target

MarkAtwood:
## Summary

Adds `make sbom`, `make install-sbom`, and `make uninstall-sbom` targets to the autotools build.

- Stages a temporary install to extract the built `.so` for hashing
- Dumps preprocessor defines via `cc -dM -E` for the options snapshot
- Calls `gen-sbom` (from wolfssl `scripts/gen-sbom`) to produce CycloneDX and SPDX outputs
- Validates SPDX with `pyspdxtools`

## Usage

```sh
./configure --with-wolfssl=/path/to/wolfssl/install
make
make sbom WOLFSSL_DIR=/path/to/wolfssl/source
```

Outputs: `wolftpm-<version>.cdx.json`, `wolftpm-<version>.spdx.json`, `wolftpm-<version>.spdx`

## Requirements

- `WOLFSSL_DIR` must point to a wolfssl source tree containing `scripts/gen-sbom` (available on the `feat/sbom-embedded` branch of wolfssl)
- `python3` and `pip install spdx-tools` on the build host

## Test plan
- [ ] `./configure --with-wolfssl=... && make && make sbom WOLFSSL_DIR=...`
- [ ] Verify three output files are produced
- [ ] `pyspdxtools --infile wolftpm-*.spdx.json` passes validation

---

## CMake `sbom` target

The cmake build now has an equivalent `sbom` custom target, so SBOM generation works the same way whether you build with autotools or cmake.

```sh
cmake -B build -DWOLFSSL_DIR=/path/to/wolfssl/source .
cmake --build build
cmake --build build --target sbom
```

Outputs land in the build directory: `wolftpm-<version>.cdx.json`, `wolftpm-<version>.spdx.json`, `wolftpm-<version>.spdx`.

Notes:
- Version is read directly from `wolftpm/version.h` (`LIBWOLFTPM_VERSION_STRING`), not from the cmake `project()` call, so the cmake and autotools SBOMs stay bit-identical in the version field even if the two drift.
- The options fingerprint is taken from the generated `wolftpm/options.h` (the real compile-time option set) rather than a raw `cc -dM` dump, which would only capture compiler builtins. This keeps the cmake SBOM consistent with what autotools produces via `--options-h`.
- `WOLFSSL_DIR` and `python3`/`pyspdxtools` are validated at configure time. A missing prerequisite does **not** break a normal `cmake --build`; the `sbom` target simply fails with a clear message when invoked. A wrong `-DWOLFSSL_DIR=` (explicit opt-in with a bad path) hard-errors at configure time.
- On lib64 hosts, pass `-DCMAKE_INSTALL_LIBDIR=lib` if needed; the staged library is read from `lib/`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -809,9 +809,9 @@ if(BUILD_WOLFTPM_LIB)
     # this file drifts out of sync with the header.
     file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/wolftpm/version.h"
          _sbom_version_line
-         REGEX "^#define[ \t]+LIBWOLFTPM_VERSION_STRING[ \t]+\"")
+         REGEX "^#define[ 	]+LIBWOLFTPM_VERSION_STRING[ 	]+\"")
     string(REGEX REPLACE
-           "^#define[ \t]+LIBWOLFTPM_VERSION_STRING[ \t]+\"([^\"]+)\".*" "\\1"
+           "^#define[ 	]+LIBWOLFTPM_VERSION_STRING[ 	]+\"([^\"]+)\".*" "\\1"
            SBOM_VERSION "${_sbom_version_line}")
     if(NOT SBOM_VERSION)
         message(FATAL_ERROR
@@ -866,11 +866,17 @@ if(BUILD_WOLFTPM_LIB)
     set(SBOM_SPDX_TV "${CMAKE_BINARY_DIR}/wolftpm-${SBOM_VERSION}.spdx")
     set(SBOM_STAGING "${CMAKE_BINARY_DIR}/_sbom_staging")
 
-    # Staged install path. install(TARGETS wolftpm) above hardcodes
-    # `LIBRARY DESTINATION lib`, so the .so always lands in lib/, never lib64/.
-    # ${CMAKE_SHARED_LIBRARY_SUFFIX} resolves to .so on Linux / .dylib on mac.
-    set(SBOM_LIB
-        "${SBOM_STAGING}/lib/libwolftpm${CMAKE_SHARED_LIBRARY_SUFFIX}")
+    # Staged install path. Derive the installed artifact name from the target
+    # itself so it tracks the real output - a versioned .so/.dylib, a static
+    # .a/.lib, or the Windows DLL - instead of a hardcoded
+    # libwolftpm${CMAKE_SHARED_LIBRARY_SUFFIX} that only matches a Unix shared
+    # build. install(TARGETS) sends the Windows DLL (RUNTIME) to bin/ and
+    # everything else (LIBRARY/ARCHIVE) to lib/.
+    if(WIN32 AND BUILD_SHARED_LIBS)
+        set(SBOM_LIB "${SBOM_STAGING}/bin/$<TARGET_FILE_NAME:wolftpm>")
+    else()
+        set(SBOM_LIB "${SBOM_STAGING}/lib/$<TARGET_FILE_NAME:wolftpm>")
+    endif()
 
     # wolfTPM links wolfSSL/wolfCrypt, so record wolfSSL as a dependency in the
     # SBOM (matches the autotools SBOM_DEP_WOLFSSL=yes).  Recording it needs the
@@ -899,9 +905,9 @@ if(BUILD_WOLFTPM_LIB)
         VERBATIM
         # Stage a clean install so gen-sbom inspects the same artifact layout
         # the user would actually ship.
-        COMMAND ${CMAKE_COMMAND} -E rm -rf ${SBOM_STAGING}
+        COMMAND ${CMAKE_COMMAND} -E remove_directory ${SBOM_STAGING}
         COMMAND ${CMAKE_COMMAND} --install ${CMAKE_BINARY_DIR}
-                --prefix ${SBOM_STAGING}
+                --prefix ${SBOM_STAGING} --config $<CONFIG>
         COMMAND ${PYTHON3_CMD} ${WOLFSSL_DIR}/scripts/gen-sbom
                 --name wolftpm
                 --version ${SBOM_VERSION}
@@ -918,7 +924,7 @@ if(BUILD_WOLFTPM_LIB)
         # the target.
         COMMAND ${PYSPDXTOOLS_CMD} --infile ${SBOM_SPDX}
                 --outfile ${SBOM_SPDX_TV}
-        COMMAND ${CMAKE_COMMAND} -E rm -rf ${SBOM_STAGING}
+        COMMAND ${CMAKE_COMMAND} -E remove_directory ${SBOM_STAGING}
         COMMENT "Generating SBOM for wolfTPM ${SBOM_VERSION}")
 
     # gen-sbom reads the staged library, so the library must build first.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -793,6 +793,15 @@ if(BUILD_WOLFTPM_LIB)
     set(WOLFSSL_DIR "" CACHE PATH
         "Path to wolfssl source tree with scripts/gen-sbom")
 
+    # wolfTPM is GPLv3-or-later (per the per-file source headers: "either
+    # version 3 of the License, or (at your option) any later version") or
+    # commercial.  Pin the header-accurate SPDX id so the SBOM is correct
+    # regardless of the gen-sbom version's licence detection; commercial
+    # licensees can override it (e.g. LicenseRef-wolfSSL-Commercial).  This
+    # mirrors SBOM_LICENSE_OVERRIDE in the autotools build.
+    set(SBOM_LICENSE_OVERRIDE "GPL-3.0-or-later" CACHE STRING
+        "SPDX licence expression recorded in the SBOM")
+
     # Derive the version from wolftpm/version.h, NOT from PROJECT_VERSION.
     # autotools `make sbom` uses PACKAGE_VERSION, which is sourced from
     # version.h. Reading the same header here keeps the cmake and autotools
@@ -863,6 +872,25 @@ if(BUILD_WOLFTPM_LIB)
     set(SBOM_LIB
         "${SBOM_STAGING}/lib/libwolftpm${CMAKE_SHARED_LIBRARY_SUFFIX}")
 
+    # wolfTPM links wolfSSL/wolfCrypt, so record wolfSSL as a dependency in the
+    # SBOM (matches the autotools SBOM_DEP_WOLFSSL=yes).  Recording it needs the
+    # gen-sbom from wolfSSL/wolfssl#10343; older gen-sbom rejects unknown flags,
+    # so probe --help and only pass --dep-wolfssl when supported.  Against an
+    # older gen-sbom the SBOM is still valid but omits the wolfSSL entry.
+    set(_sbom_dep_wolfssl "")
+    execute_process(
+        COMMAND ${PYTHON3_CMD} ${WOLFSSL_DIR}/scripts/gen-sbom --help
+        OUTPUT_VARIABLE _sbom_help
+        ERROR_QUIET)
+    if(_sbom_help MATCHES "--dep-wolfssl")
+        set(_sbom_dep_wolfssl --dep-wolfssl yes)
+    else()
+        message(WARNING
+            "sbom: ${WOLFSSL_DIR}/scripts/gen-sbom has no --dep-wolfssl; "
+            "the SBOM will omit the wolfSSL dependency entry. Use the gen-sbom "
+            "from wolfSSL/wolfssl#10343 (or master once merged) to record it.")
+    endif()
+
     # ${OPTION_FILE} is the generated wolftpm/options.h, the same compile-time
     # option fingerprint autotools `make sbom` feeds to gen-sbom via
     # --options-h. Using it (rather than a raw `cc -dM` dump, which would only
@@ -881,6 +909,8 @@ if(BUILD_WOLFTPM_LIB)
                 --license-file ${CMAKE_SOURCE_DIR}/LICENSE
                 --options-h ${OPTION_FILE}
                 --lib ${SBOM_LIB}
+                --license-override ${SBOM_LICENSE_OVERRIDE}
+                ${_sbom_dep_wolfssl}
                 --cdx-out ${SBOM_CDX}
                 --spdx-out ${SBOM_SPDX}
         # Validate the SPDX JSON and emit the tag-value rendering as a side
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -770,3 +770,127 @@ install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/wolftpm/
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/wolftpm/
         DESTINATION include/wolftpm
         FILES_MATCHING PATTERN "*.h")
+
+
+####################################################
+# SBOM generation target
+####################################################
+#
+# Usage:
+#   cmake -B build -DWOLFSSL_DIR=/path/to/wolfssl/source .
+#   cmake --build build
+#   cmake --build build --target sbom
+#
+# WOLFSSL_DIR must point to a wolfssl source tree that contains
+# scripts/gen-sbom (available on the feat/sbom-embedded branch).
+#
+# Outputs written to the build directory:
+#   wolftpm-<version>.cdx.json   (CycloneDX)
+#   wolftpm-<version>.spdx.json  (SPDX JSON)
+#   wolftpm-<version>.spdx       (SPDX tag-value, pyspdxtools validation output)
+
+if(BUILD_WOLFTPM_LIB)
+    set(WOLFSSL_DIR "" CACHE PATH
+        "Path to wolfssl source tree with scripts/gen-sbom")
+
+    # Derive the version from wolftpm/version.h, NOT from PROJECT_VERSION.
+    # autotools `make sbom` uses PACKAGE_VERSION, which is sourced from
+    # version.h. Reading the same header here keeps the cmake and autotools
+    # SBOMs bit-identical in the version field even if the project() call in
+    # this file drifts out of sync with the header.
+    file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/wolftpm/version.h"
+         _sbom_version_line
+         REGEX "^#define[ \t]+LIBWOLFTPM_VERSION_STRING[ \t]+\"")
+    string(REGEX REPLACE
+           "^#define[ \t]+LIBWOLFTPM_VERSION_STRING[ \t]+\"([^\"]+)\".*" "\\1"
+           SBOM_VERSION "${_sbom_version_line}")
+    if(NOT SBOM_VERSION)
+        message(FATAL_ERROR
+            "sbom: could not parse LIBWOLFTPM_VERSION_STRING from "
+            "wolftpm/version.h")
+    endif()
+
+    # Validate the SBOM prerequisites at configure time so the user learns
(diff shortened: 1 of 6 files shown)


Makefile.am
```diff
@@ -132,3 +132,70 @@ cppcheck:
...
+	fi
+	rm -rf $(abs_builddir)/_sbom_staging
+	$(MAKE) install DESTDIR=$(abs_builddir)/_sbom_staging
+	$(PYTHON3) $(WOLFSSL_DIR)/scripts/gen-sbom \
+	    --name wolftpm \
+	    --version $(PACKAGE_VERSION) \
+	    --supplier "wolfSSL Inc." \
+	    --license-file $(srcdir)/LICENSE \
+	    --options-h $(abs_builddir)/wolftpm/options.h \
+	    --lib $(abs_builddir)/_sbom_staging$(libdir)/libwolftpm.so.@WOLFTPM_LIBRARY_VERSION_FIRST@.@WOLFTPM_LIBRARY_VERSION_SECOND@.@WOLFTPM_LIBRARY_VERSION_THIRD@ \
```
aidangarske: **🔴 [High] Autotools sbom target reconstructs the libtool library filename incorrectly; breaks on next version-info bump and on…** · _Logic_

The `sbom` target points `gen-sbom --lib` at `libwolftpm.so.@WOLFTPM_LIBRARY_VERSION_FIRST@.@WOLFTPM_LIBRARY_VERSION_SECOND@.@WOLFTPM_LIBRARY_VERSION_THIRD@`, i.e. `libwolftpm.so.\<current>.\<revision>.\<age>`, reconstructed from libtool's `current:revision:age` version-info fields (currently `17:0:0`). libtool actually installs the real ELF file as `libNAME.so.(current-age).age.revision`: the first component should be `current-age` (not `current`) and the trailing two are `age.revision`, swapped relative to this recipe's `revision.age`. configure.ac's own comments confirm the installed soname is `current-age`, making the mapping internally inconsistent. It works today only because all three components collapse to `17.0.0` for `17:0:0`; the next release with a nonzero `revision`/`age` (e.g. `17:1:0`) makes libtool install `libwolftpm.so.17.0.1` while the recipe looks for `libwolftpm.so.17.1.0`, so `--lib` points at a nonexistent file and `make sbom` fails. The hardcoded `.so.X.Y.Z` and `$(libdir)` are also non-portable: macOS installs `libwolftpm.\<n>.dylib`, static-only builds produce `libwolftpm.a`, and MinGW installs the DLL runtime under `$(bindir)` — none of which this path can locate. Severity views across modes: review=High/BLOCK, bugs=Medium, review-security=Low; strictest kept.

**Fix:** Do not reconstruct the versioned name from version-info components. After the staged install, discover the actual installed artifact under the staged `$(libdir)`/`$(bindir)` (glob `libwolftpm.so.*`, `libwolftpm*.dylib`, `*wolftpm*.dll`, or dereference the unversioned `libwolftpm.so` dev symlink) and fail clearly if none is found. If the versioned name must be kept, compute the first component as `current-age` and order the trailing components as `age.revision`.

CMakeLists.txt
```diff
@@ -770,3 +770,127 @@ install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/wolftpm/
...
+            COMMENT "SBOM prerequisites missing")
+        return()
+    endif()
+
+    set(SBOM_CDX     "${CMAKE_BINARY_DIR}/wolftpm-${SBOM_VERSION}.cdx.json")
+    set(SBOM_SPDX    "${CMAKE_BINARY_DIR}/wolftpm-${SBOM_VERSION}.spdx.json")
+    set(SBOM_SPDX_TV "${CMAKE_BINARY_DIR}/wolftpm-${SBOM_VERSION}.spdx")
+    set(SBOM_STAGING "${CMAKE_BINARY_DIR}/_sbom_staging")
+
+    # Staged install path. install(TARGETS wolftpm) above hardcodes
```
aidangarske: **🔴 [High] CMake sbom target hard-codes Unix shared-library layout; fails on Windows, static, and multi-config builds** · _Logic_

The CMake `sbom` target computes `SBOM_LIB = ${SBOM_STAGING}/lib/libwolftpm${CMAKE_SHARED_LIBRARY_SUFFIX}` and passes it to `gen-sbom` after `cmake --install`, hard-coding both the `lib/` staging subdir and a Unix shared-object name rather than the real target artifact. This diverges from the actual install layout in several supported configurations: (1) With `-DBUILD_SHARED_LIBS=OFF` (option defaults ON at line 73) the installed artifact is `libwolftpm.a` under `ARCHIVE DESTINATION lib`, so `--lib` points at a `.so`/`.dylib` that was never produced. (2) On Windows shared builds, `install(TARGETS)` uses `RUNTIME DESTINATION bin`, DLLs go under `bin/` and typically drop the `lib` prefix (`wolftpm.dll`), so the path never matches. (3) For multi-config generators (Visual Studio, Xcode) the `cmake --install` omits `--config $\<CONFIG>`, so building the `sbom` target can install/hash the wrong or unbuilt configuration. In each case `gen-sbom` fails with a bare file-not-found. Severity views: review=High/BLOCK, bugs=Medium, review-security=Low; strictest kept.

**Fix:** Derive the artifact from CMake target metadata via generator expressions (`$\<TARGET_FILE_NAME:wolftpm>` plus the platform-specific staged destination — `bin/` on Windows, `lib/` elsewhere), add `--config $\<CONFIG>` to the install for multi-config generators, and either support static builds explicitly or reject `BUILD_SHARED_LIBS=OFF` for the sbom target with a clear FATAL_ERROR.

CMakeLists.txt
```diff
@@ -770,3 +770,127 @@ install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/wolftpm/
...
+
+    # ${OPTION_FILE} is the generated wolftpm/options.h, the same compile-time
+    # option fingerprint autotools `make sbom` feeds to gen-sbom via
+    # --options-h. Using it (rather than a raw `cc -dM` dump, which would only
+    # capture compiler builtins) keeps the cmake SBOM consistent with autotools.
+    add_custom_target(sbom
+        VERBATIM
+        # Stage a clean install so gen-sbom inspects the same artifact layout
+        # the user would actually ship.
+        COMMAND ${CMAKE_COMMAND} -E rm -rf ${SBOM_STAGING}
```
aidangarske: **🟠 [Medium] CMake sbom target uses cmake -E rm, unavailable on the declared minimum CMake 3.16** · _API Contract Violations_

The `sbom` custom target runs `${CMAKE_COMMAND} -E rm -rf ${SBOM_STAGING}` twice (lines 874 and 891). The `cmake -E rm` subcommand was introduced in CMake 3.17, but this file declares `cmake_minimum_required(VERSION 3.16)` (line 22). On a host with exactly CMake 3.16 — a version the project claims to support — `cmake --build build --target sbom` fails with `cmake -E rm: unknown command` before any SBOM is produced. The main build is unaffected; only the new opt-in target breaks.

**Fix:** Use the 3.16-compatible `${CMAKE_COMMAND} -E remove_directory ${SBOM_STAGING}` (handles a missing directory gracefully and matches `rm -rf` intent), or raise the minimum CMake version for this target path.

Makefile.am
```diff
@@ -132,3 +132,70 @@ cppcheck:
...
+
+# SBOM generation (CRA compliance)
+SBOM_CDX     = wolftpm-$(PACKAGE_VERSION).cdx.json
+SBOM_SPDX    = wolftpm-$(PACKAGE_VERSION).spdx.json
+SBOM_SPDX_TV = wolftpm-$(PACKAGE_VERSION).spdx
+sbomdir      = $(datadir)/doc/$(PACKAGE)
+
+.PHONY: sbom install-sbom uninstall-sbom
+
+sbom:
```
aidangarske: **🟠 [Medium] New SBOM targets have no automated coverage** · _test_

The PR adds non-trivial autotools and CMake SBOM generation paths, but includes no automated smoke coverage for either target; the PR body lists only manual test steps. A lightweight test would have caught the artifact-path and platform assumptions in the new target logic (see the autotools libtool-filename and CMake shared-library-layout findings).

**Fix:** Add CI or a build-system smoke test that provides a minimal fake `WOLFSSL_DIR/scripts/gen-sbom`, runs `make sbom` and `cmake --build --target sbom`, and asserts the expected output files are produced or that prerequisite errors are clear.

aidangarske (commented):
## Skoll Multi-Scan Review

**Findings:** 7 total — 7 posted, 0 skipped
**7 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] [review+review-security+bugs] **Autotools sbom target reconstructs the libtool library filename incorrectly; breaks on next version-info bump and on macOS/static builds** — `Makefile.am:181`
- [High] [review+review-security+bugs] **CMake sbom target hard-codes Unix shared-library layout; fails on Windows, static, and multi-config builds** — `CMakeLists.txt:860-884`
- [Medium] [bugs] **CMake sbom target uses cmake -E rm, unavailable on the declared minimum CMake 3.16** — `CMakeLists.txt:874,891`
- [Medium] [review] **New SBOM targets have no automated coverage** — `Makefile.am:144-188`
- [Low] [review] **install-sbom mixes absolute and relative artifact paths; only works when run from the build dir** — `Makefile.am:184-194`
- [Low] [review+review-security] **CMake return() used at directory scope to skip the SBOM block; fragile if code is appended** — `CMakeLists.txt:844-853`
- [Low] [review-security] **CMake version.h regex uses [ \t] which does not match a tab in CMake's regex dialect** — `CMakeLists.txt:803-806`

Makefile.am
```diff
@@ -132,3 +132,70 @@ cppcheck:
...
+	$(PYTHON3) $(WOLFSSL_DIR)/scripts/gen-sbom \
+	    --name wolftpm \
+	    --version $(PACKAGE_VERSION) \
+	    --supplier "wolfSSL Inc." \
+	    --license-file $(srcdir)/LICENSE \
+	    --options-h $(abs_builddir)/wolftpm/options.h \
+	    --lib $(abs_builddir)/_sbom_staging$(libdir)/libwolftpm.so.@WOLFTPM_LIBRARY_VERSION_FIRST@.@WOLFTPM_LIBRARY_VERSION_SECOND@.@WOLFTPM_LIBRARY_VERSION_THIRD@ \
+	    $(if $(SBOM_LICENSE_OVERRIDE),--license-override $(SBOM_LICENSE_OVERRIDE)) \
+	    $(if $(SBOM_LICENSE_TEXT),--license-text $(SBOM_LICENSE_TEXT)) \
+	    --cdx-out $(abs_builddir)/$(SBOM_CDX) \
```
aidangarske: **🔵 [Low] install-sbom mixes absolute and relative artifact paths; only works when run from the build dir** · _convention_

The `sbom` recipe writes outputs to `$(abs_builddir)/$(SBOM_CDX)` (absolute) at line 184, while `install-sbom` installs the bare `$(SBOM_CDX)`, `$(SBOM_SPDX)`, and `$(SBOM_SPDX_TV)` relative to the current directory at line 192. These only line up when `make install-sbom` runs from the build directory; VPATH/out-of-tree `distcheck`-style builds or invoking from a subdir would look in the wrong place.

**Fix:** Use `$(abs_builddir)/$(SBOM_*)` consistently in install-sbom (and CLEANFILES) to match where the sbom target actually writes the artifacts.

CMakeLists.txt
```diff
@@ -770,3 +770,127 @@ install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/wolftpm/
...
+            "      Check that WOLFSSL_DIR points to a wolfSSL tree with "
+            "SBOM support.")
+    elseif(NOT PYTHON3_CMD)
+        set(_sbom_error "'python3' not found in PATH. Cannot generate SBOM.")
+    elseif(NOT PYSPDXTOOLS_CMD)
+        set(_sbom_error
+            "'pyspdxtools' not found in PATH (install: pip install spdx-tools)")
+    endif()
+
+    if(NOT _sbom_error STREQUAL "")
```
aidangarske, line 853: **🔵 [Low] CMake return() used at directory scope to skip the SBOM block; fragile if code is appended** · _Logic_

When an SBOM prerequisite (python3/pyspdxtools/WOLFSSL_DIR) is missing, the code defines a failing fallback `sbom` target and then calls `return()`. This block is guarded by `if(BUILD_WOLFTPM_LIB)` at top-level directory scope (not inside a function), so `return()` ends processing of the remainder of CMakeLists.txt, not just the SBOM `if` block. It is harmless today only because the SBOM section is the last block in the file (ends at the `endif()` on line 896); any `install()`, target, or export logic added after it would be silently skipped in the common no-SBOM build. Severity views: review=Low/NIT, review-security=Info; strictest kept.

**Fix:** Avoid `return()` at directory scope for control flow. Select the fallback vs. real target with `if()/else()` (or a nested-if scoped to the SBOM section), or move the SBOM logic into a function/included module so the early exit cannot skip future directives.

CMakeLists.txt
```diff
@@ -770,3 +770,127 @@ install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/wolftpm/
...
+        "Path to wolfssl source tree with scripts/gen-sbom")
+
+    # Derive the version from wolftpm/version.h, NOT from PROJECT_VERSION.
+    # autotools `make sbom` uses PACKAGE_VERSION, which is sourced from
+    # version.h. Reading the same header here keeps the cmake and autotools
+    # SBOMs bit-identical in the version field even if the project() call in
+    # this file drifts out of sync with the header.
+    file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/wolftpm/version.h"
+         _sbom_version_line
+         REGEX "^#define[ \t]+LIBWOLFTPM_VERSION_STRING[ \t]+\"")
```
aidangarske: **🔵 [Low] CMake version.h regex uses [ \t] which does not match a tab in CMake's regex dialect** · _Logic_

The SBOM version is parsed with `REGEX "^#define[ \t]+LIBWOLFTPM_VERSION_STRING[ \t]+\""`. CMake's regex engine does not interpret `\t` as a tab escape; inside the bracket expression `[ \t]` is the literal set {space, backslash, 't'}. It matches today only because the distributed `wolftpm/version.h` uses space separators. If version.h were ever regenerated with tab separators, `file(STRINGS ... REGEX)` would return no match, `_sbom_version_line` would be empty, and the target would abort via the FATAL_ERROR at 807-811. Currently benign — defense-in-depth only.

**Fix:** Since version.h is generated with single spaces, match `[ ]+` explicitly, or embed a literal tab in the character class if tab tolerance is actually desired.

dgarske (changes requested):
Please apply this patch (I can't figure out a way to push to your repo):

```patch
From 89eae19fb3fe4079fbb73b5d504969433ad024b5 Mon Sep 17 00:00:00 2001
From: David Garske <david@wolfssl.com>
Date: Fri, 10 Jul 2026 13:49:25 -0700
Subject: [PATCH] SBOM: fix cmake artifact discovery, cmake 3.16 compat, and
 tab regex

---
 CMakeLists.txt | 26 ++++++++++++++++----------
 1 file changed, 16 insertions(+), 10 deletions(-)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index a6a04b86..2a144813 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -809,9 +809,9 @@ if(BUILD_WOLFTPM_LIB)
     # this file drifts out of sync with the header.
     file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/wolftpm/version.h"
          _sbom_version_line
-         REGEX "^#define[ \t]+LIBWOLFTPM_VERSION_STRING[ \t]+\"")
+         REGEX "^#define[ 	]+LIBWOLFTPM_VERSION_STRING[ 	]+\"")
     string(REGEX REPLACE
-           "^#define[ \t]+LIBWOLFTPM_VERSION_STRING[ \t]+\"([^\"]+)\".*" "\\1"
+           "^#define[ 	]+LIBWOLFTPM_VERSION_STRING[ 	]+\"([^\"]+)\".*" "\\1"
            SBOM_VERSION "${_sbom_version_line}")
     if(NOT SBOM_VERSION)
         message(FATAL_ERROR
@@ -866,11 +866,17 @@ if(BUILD_WOLFTPM_LIB)
     set(SBOM_SPDX_TV "${CMAKE_BINARY_DIR}/wolftpm-${SBOM_VERSION}.spdx")
     set(SBOM_STAGING "${CMAKE_BINARY_DIR}/_sbom_staging")
 
-    # Staged install path. install(TARGETS wolftpm) above hardcodes
-    # `LIBRARY DESTINATION lib`, so the .so always lands in lib/, never lib64/.
-    # ${CMAKE_SHARED_LIBRARY_SUFFIX} resolves to .so on Linux / .dylib on mac.
-    set(SBOM_LIB
-        "${SBOM_STAGING}/lib/libwolftpm${CMAKE_SHARED_LIBRARY_SUFFIX}")
+    # Staged install path. Derive the installed artifact name from the target
+    # itself so it tracks the real output - a versioned .so/.dylib, a static
+    # .a/.lib, or the Windows DLL - instead of a hardcoded
+    # libwolftpm${CMAKE_SHARED_LIBRARY_SUFFIX} that only matches a Unix shared
+    # build. install(TARGETS) sends the Windows DLL (RUNTIME) to bin/ and
+    # everything else (LIBRARY/ARCHIVE) to lib/.
+    if(WIN32 AND BUILD_SHARED_LIBS)
+        set(SBOM_LIB "${SBOM_STAGING}/bin/$<TARGET_FILE_NAME:wolftpm>")
+    else()
+        set(SBOM_LIB "${SBOM_STAGING}/lib/$<TARGET_FILE_NAME:wolftpm>")
+    endif()
 
     # wolfTPM links wolfSSL/wolfCrypt, so record wolfSSL as a dependency in the
     # SBOM (matches the autotools SBOM_DEP_WOLFSSL=yes).  Recording it needs the
@@ -899,9 +905,9 @@ if(BUILD_WOLFTPM_LIB)
         VERBATIM
         # Stage a clean install so gen-sbom inspects the same artifact layout
         # the user would actually ship.
-        COMMAND ${CMAKE_COMMAND} -E rm -rf ${SBOM_STAGING}
+        COMMAND ${CMAKE_COMMAND} -E remove_directory ${SBOM_STAGING}
         COMMAND ${CMAKE_COMMAND} --install ${CMAKE_BINARY_DIR}
-                --prefix ${SBOM_STAGING}
+                --prefix ${SBOM_STAGING} --config $<CONFIG>
         COMMAND ${PYTHON3_CMD} ${WOLFSSL_DIR}/scripts/gen-sbom
                 --name wolftpm
                 --version ${SBOM_VERSION}
@@ -918,7 +924,7 @@ if(BUILD_WOLFTPM_LIB)
         # the target.
         COMMAND ${PYSPDXTOOLS_CMD} --infile ${SBOM_SPDX}
                 --outfile ${SBOM_SPDX_TV}
-        COMMAND ${CMAKE_COMMAND} -E rm -rf ${SBOM_STAGING}
+        COMMAND ${CMAKE_COMMAND} -E remove_directory ${SBOM_STAGING}
         COMMENT "Generating SBOM for wolfTPM ${SBOM_VERSION}")
 
     # gen-sbom reads the staged library, so the library must build first.
-- 
2.50.1 (Apple Git-155)
```

MarkAtwood:
@dgarske Applied verbatim as 7f7aebe (kept your authorship via git am). Verified end-to-end against a cmake-built wolfssl 5.9.2: configure extracts 4.0.0 through the tab-class regex, the staged install resolves the artifact via $<TARGET_FILE_NAME:wolftpm>, and the sbom target generates all three files (cdx.json, spdx.json, spdx tag-value) with remove_directory doing the staging cleanup. I don't have a cmake 3.16 on hand, so that compat claim rides on your testing.

dgarske (approved):

PR #537, opened 2026-06-23
Add Transparent TPM_RC_RETRY auto-resubmit

aidangarske:
# Description
The TPM returns `TPM_RC_RETRY` when momentarily busy — e.g. the core
  dictionary-attack logic persisting the `daUsed` flag to NV on the first auth
  use of a non-`noDA` key (an externally provisioned AIK/SUDI key). wolfTPM
  passed this raw code back to the caller, unlike the TCG ESYS stack which
  auto-resubmits. A customer hit it on `wolfTPM2_SignSequenceComplete`.

  This adds optional, bounded auto-resubmit at the command chokepoint
  (`TPM2_SendCommand` / `TPM2_SendCommandAuth`), covering all transports. A
  `TPM_RC_RETRY` response is header-only, so the saved 10-byte header is restored
  and the identical command resent — no re-marshalling.
  
  **Opt-in / disabled by default.** wolfTPM's own key templates all set `noDA`,
  so wolfTPM-created keys never trigger it; defaulting on would change behavior
  for everyone to serve a case none of our keys reach. Existing callers keep the
  raw `TPM_RC_RETRY` unless they opt in.

  - `TPM2_SetCommandRetries(&dev.ctx, N)` — enable at runtime (0 = off)
  - `-DWOLFTPM_MAX_RETRIES=N` — change the compile-time default (default 0)
  - `WOLFTPM_NO_RETRY` — compile the handling out entirely (~528 B)
  
  Note: `TPM_RC_RETRY` is original TPM 2.0 (Part 2 rev 00.99, 2014) core-TPM DA
  behavior — unrelated to SPDM, transport, or SPI wait states.

  Tests cover the resubmit bookkeeping and the config API; ChangeLog and README
  updated
  
  ZD 22030

### DIFF
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -452,7 +452,13 @@ static TPM_RC TPM2_SPDM_SendCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
         packet->buf, packet->pos, tpmResp, &tpmRespSz);
     if (rc != 0) {
         TPM2_ForceZero(tpmResp, sizeof(tpmResp));
-        return rc;
+    #ifdef DEBUG_WOLFTPM
+        printf("SPDM secured exchange failed: %d\n", rc);
+    #endif
+        /* SPDM is active: never downgrade to cleartext. Map the (negative)
+         * transport error to a positive TPM RC so the caller treats it as a
+         * hard failure instead of "SPDM not active". */
+        return TPM_RC_FAILURE;
     }
 
     if (tpmRespSz > MAX_RESPONSE_SIZE) {
@@ -467,6 +473,25 @@ static TPM_RC TPM2_SPDM_SendCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
 }
 #endif /* WOLFTPM_SPDM */
 
+/* Send the finalized command in packet over the SPDM secured channel when a
+ * session is active, otherwise over the raw transport. Once SPDM is active an
+ * exchange failure is returned as-is and never falls back to a cleartext send,
+ * so a forced SPDM failure cannot downgrade the link. */
+static TPM_RC TPM2_DispatchCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
+{
+    TPM_RC rc;
+
+#ifdef WOLFTPM_SPDM
+    rc = TPM2_SPDM_SendCommand(ctx, packet);
+    if (rc >= 0)
+        return rc; /* SPDM active: success or hard failure, no cleartext */
+    /* rc < 0: SPDM not active, use normal transport */
+#endif
+    rc = (TPM_RC)INTERNAL_SEND_COMMAND(ctx, packet);
+
+    return rc;
+}
+
 #ifdef WOLFTPM_NO_RETRY
 /* Submit the finalized command in packet (length cmdSz) and parse the
  * response, returning the TPM response code. */
@@ -478,19 +503,9 @@ static TPM_RC TPM2_TransmitCommand(TPM2_CTX* ctx, TPM2_Packet* packet,
     /* send command requires packet->pos to be the total command length */
     packet->pos = cmdSz;
 
-#ifdef WOLFTPM_SPDM
-    rc = TPM2_SPDM_SendCommand(ctx, packet);
-    if (rc >= 0) {
-        if (rc != TPM_RC_SUCCESS)
-            return rc; /* SPDM active but failed, do not retry cleartext */
-    }
-    else /* rc < 0: SPDM not active, use normal transport */
-#endif
-    {
-        rc = (TPM_RC)INTERNAL_SEND_COMMAND(ctx, packet);
-    }
+    rc = TPM2_DispatchCommand(ctx, packet);
     if (rc != 0)
-        return rc; /* transport error */
+        return rc; /* transport or SPDM error */
 
     /* parse response header and extract the TPM response code */
     rc = TPM2_Packet_Parse(rc, packet);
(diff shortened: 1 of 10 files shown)


src/tpm2.c
```diff
@@ -467,6 +467,60 @@ static TPM_RC TPM2_SPDM_SendCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
...
+ * response, returning the TPM response code. On TPM_RC_RETRY the TPM is
+ * momentarily busy (e.g. persisting the daUsed flag on first auth use of a
+ * non-noDA key) and asks for the identical command to be resubmitted; the
+ * error response is header-only, so restoring the command header and resending
+ * up to ctx->retries times recovers transparently. */
+static TPM_RC TPM2_TransmitWithRetry(TPM2_CTX* ctx, TPM2_Packet* packet,
+    UINT32 cmdSz)
+{
+    TPM_RC rc;
+#ifndef WOLFTPM_NO_RETRY
```
dgarske: Please document this new `WOLFTPM_NO_RETRY` option in README.md. Thanks

src/tpm2_packet.c
```diff
@@ -1625,6 +1625,21 @@ TPM_RC TPM2_Packet_Parse(TPM_RC rc, TPM2_Packet* packet)
     return rc;
 }
 
+int TPM2_Packet_RetryRestore(TPM_RC rc, int* retries, TPM2_Packet* packet,
```
dgarske, line 1629: Wrap with `#ifndef WOLFTPM_NO_RETRY`

wolftpm/tpm2.h
```diff
@@ -2229,6 +2229,9 @@ typedef struct TPM2_CTX {
     void* spdmCtx; /* Pointer to WOLFTPM2_SPDM_CTX when session active */
     unsigned int spdmOnlyDetected:1; /* TPM_RC_DISABLED from Startup */
 #endif
+
+    /* Additional resubmit attempts on TPM_RC_RETRY (0 disables) */
+    int retries;
```
dgarske, line 2235: Wrap with `#ifndef WOLFTPM_NO_RETRY`

wolftpm/tpm2.h
```diff
@@ -3773,6 +3776,36 @@ WOLFTPM_API TPM_RC TPM2_ChipStartup(TPM2_CTX* ctx, int timeoutTries);
...
+    \return TPM_RC_SUCCESS: successful
+    \return BAD_FUNC_ARG: the TPM2 context is NULL or retries is negative
+
+    \param ctx pointer to a TPM2_CTX struct
+    \param retries number of additional resubmit attempts on TPM_RC_RETRY (0 disables)
+
+    \sa TPM2_GetCommandRetries
+    \sa wolfTPM2_SetCommandRetries
+*/
+WOLFTPM_API TPM_RC TPM2_SetCommandRetries(TPM2_CTX* ctx, int retries);
```
dgarske, line 3795: Wrap these with `#ifndef WOLFTPM_NO_RETRY`

wolftpm/tpm2_wrap.h
```diff
@@ -341,6 +341,34 @@ WOLFTPM_API int wolfTPM2_Cleanup_ex(WOLFTPM2_DEV* dev, int doShutdown);
...
+
+    \return TPM_RC_SUCCESS: successful
+    \return BAD_FUNC_ARG: dev is NULL or retries is negative
+
+    \param dev pointer to a populated structure of WOLFTPM2_DEV type
+    \param retries number of additional resubmit attempts on TPM_RC_RETRY (0 disables)
+
+    \sa wolfTPM2_GetCommandRetries
+*/
+WOLFTPM_API int wolfTPM2_SetCommandRetries(WOLFTPM2_DEV* dev, int retries);
```
dgarske: Do we really need wrappers for these? Can't we just expose the tpm2.c versions?

dgarske (changes requested):

dgarske (dismissed):

src/tpm2.c
```diff
@@ -467,6 +467,60 @@ static TPM_RC TPM2_SPDM_SendCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
 }
 #endif /* WOLFTPM_SPDM */
 
+/* Submit the finalized command in packet (length cmdSz) and parse the
```
dgarske, line 516: I think it would be cleaner to have with two distinct `TPM2_TransmitWithRetry` functions  based on WOLFTPM_NO_RETRY. And change the name of API so WithRetry is removed. There is for() loop logic that becomes obsolete with WOLFTPM_NO_RETRY

dgarske (changes requested):

src/tpm2.c
```diff
@@ -467,6 +467,88 @@ static TPM_RC TPM2_SPDM_SendCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
...
+static TPM_RC TPM2_TransmitCommand(TPM2_CTX* ctx, TPM2_Packet* packet,
+    UINT32 cmdSz)
+{
+    TPM_RC rc;
+
+    /* send command requires packet->pos to be the total command length */
+    packet->pos = cmdSz;
+
+#ifdef WOLFTPM_SPDM
+    rc = TPM2_SPDM_SendCommand(ctx, packet);
```
dgarske: Could / should we actually map `INTERNAL_SEND_COMMAND` -> `TPM2_SPDM_SendCommand` to cleanup this section and the duplicated one below?
I suppose not since it is used to check for SPDM active. Do we need to do a better job making sure that someone can't just get SPDM to fail and have it fall back to plain text cmd/rsp?

dgarske (changes requested):

dgarske (approved):

PR #541, opened 2026-06-25
fwTPM: Dictionary Attack (DA) hardening and DA/noDA test coverage

aidangarske:
# Description 

- fTPM DA hardened to TCG spec with noDA on objects, persistent failedTries, self-heal, lockoutAuth lock, Clear/platform handling, FWTPM_DA_USED_RETRY.
- Found/fixed real bugs mid-review: brick-on-reboot, lockout-counter wipe via clock, lockoutAuth brute-force via Clear.
- New client wrappers added - wolfTPM2_DictionaryAttack{LockReset,Parameters}.
- ~22 unit tests + e2e example + isolated RETRY harness + CI job.
- PR #541 opened and server-side + test counterpart to #537

ZD 22030

### DIFF
--- a/src/fwtpm/fwtpm_command.c
+++ b/src/fwtpm/fwtpm_command.c
@@ -696,7 +696,9 @@ static TPM_RC FwCmd_Startup(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
          * is 0, or when there is no clock HAL to time-heal it (clockless
          * targets must keep a reboot recovery path). Otherwise it persists and
          * recovers on the timer, so a power-cycle cannot brute-force
-         * lockoutAuth. */
+         * lockoutAuth. The clock HAL is ms-since-boot, so re-seeding the heal
+         * baseline below measures recovery as continuous post-boot uptime; a
+         * power-cycle restarts the window rather than clearing the lock. */
         if (ctx->daLockoutRecovery == 0 || ctx->clockHal.get_ms == NULL) {
             ctx->lockoutAuthFailed = 0;
         }
--- a/src/fwtpm/fwtpm_command.c
+++ b/src/fwtpm/fwtpm_command.c
@@ -73,6 +73,9 @@ static TPM_RC FwNvCheckAccess(TPM_HANDLE authHandle,
     TPMI_RH_NV_INDEX nvHandle, UINT32 attributes, int isWrite);
 #endif
 static FWTPM_Object* FwFindObject(FWTPM_CTX* ctx, TPM_HANDLE handle);
+#ifndef FWTPM_NO_DA
+static UINT64 FwDaNowMs(FWTPM_CTX* ctx);
+#endif
 #ifdef WOLFTPM_MLDSA
 static FWTPM_SignSeq* FwFindSignSeq(FWTPM_CTX* ctx, TPM_HANDLE handle);
 #endif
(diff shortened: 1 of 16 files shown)


dgarske (dismissed):

src/fwtpm/fwtpm_command.c
```diff
@@ -676,6 +679,31 @@ static TPM_RC FwCmd_Startup(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
...
+         * applied when a clock HAL is present; clockless targets cannot
+         * self-heal, so charging the penalty there would let routine unclean
+         * power-off accumulate into lockout. */
+        if (ctx->clockHal.get_ms != NULL &&
+                !ctx->orderly && ctx->daFailedTries < ctx->daMaxTries) {
+            ctx->daFailedTries++;
+        }
+        ctx->orderly = 1;
+        ctx->daUsed = 0;
+        /* The failed-lockoutAuth lock recovers on reboot when lockoutRecovery
```
dgarske, line 695: **🔵 [Low] lockoutAuth recovery timer restarts on every Startup, so clocked builds measure per-boot uptime rather than elapsed…** · _Logic_

This PR introduces a persistent `lockoutAuthFailed` lock that, on a build with a clock HAL and a non-zero `daLockoutRecovery`, deliberately survives reboot (lines 700-702 only clear it when `daLockoutRecovery == 0 || clockHal.get_ms == NULL`). However, `FwCmd_Startup` unconditionally re-seeds the heal baseline at line 704 (`ctx->daLockoutHealMs = FwDaNowMs(ctx)`), and only the boolean `lockoutAuthFailed` is persisted (not a wall-clock failure timestamp). Because `FwDaSelfHeal` clears the lock only after `(now - daLockoutHealMs)/1000 >= daLockoutRecovery`, the recovery interval effectively measures continuous uptime since the most recent boot, not real elapsed time since the failed `lockoutAuth`. On a device that reboots more frequently than `daLockoutRecovery` (default 86400s = 1 day), the time-based recovery of the lockout hierarchy can never complete. The lock blocks `DictionaryAttackLockReset`, `DictionaryAttackParameters`, and lockoutAuth-authorized `Clear` (the gate at line 16311 rejects any command whose first handle is `TPM_RH_LOCKOUT`). The only remaining recovery is platformAuth `Clear`; if `disableClear` is also set (ClearControl), `FwCmd_Clear` returns `TPM_RC_DISABLED`, leaving no recovery path for the lockout hierarchy on such devices. BEFORE this PR there was no `lockoutAuthFailed` state at all (only volatile `daFailedTries`), so this recovery edge is newly introduced. The behavior errs toward staying locked (safe direction, not an auth bypass), and the docs claim 'clears after lockoutRecovery seconds' without the per-boot-uptime caveat.

**Fix:** Either persist a wall-clock failure timestamp (when an RTC-backed clock HAL is available) so lockoutRecovery measures true elapsed time, or do not re-seed `daLockoutHealMs` at Startup when `lockoutAuthFailed` is already set and being carried across the reboot. At minimum, document in docs/FWTPM.md that on clocked builds the lockoutAuth recovery interval is measured as continuous post-boot uptime, and ensure a platform-hierarchy escape remains available even when disableClear is set.

src/fwtpm/fwtpm_command.c
```diff
@@ -15693,22 +15792,130 @@ static const FWTPM_CMD_ENTRY* FwFindCmdEntry(TPM_CC cc)
...
+         * be a flash-wear / lock-extension DoS via wrong-auth Clear/LockReset. */
+        if (!ctx->lockoutAuthFailed) {
+            ctx->lockoutAuthFailed = 1;
+            ctx->daLockoutHealMs = FwDaNowMs(ctx);
+            ctx->orderly = 0;
+            (void)FWTPM_NV_SaveFlags(ctx);
+        }
+        return 0;
+    }
+    if (!FwHandleIsNoDA(ctx, entityH)) {
```
dgarske, line 15905: **⚪ [Info] Failed-auth path persists FLAGS to NV on every failure (flash wear / NV-bound auth latency)** · _Resource Leak_

To make `daFailedTries` durable, every DA-protected auth failure now calls `FWTPM_NV_SaveFlags(ctx)` (an append to the NV journal), as does the first DA-protected auth use per boot (`daUsed` arming, line 16366) and a failed `lockoutAuth` (line 15899). Before this PR, `daFailedTries` was volatile and failures did not touch NV. The increment is correctly capped at `daMaxTries` so writes are bounded per lockout cycle (default 32), and the `lockoutAuthFailed` write is guarded to occur once per episode. This is sound and explicitly documented in docs/FWTPM.md, but on flash-backed targets it adds write-cycle wear and makes failed-auth latency NV-bound. Noting it so integrators size the NV backend / wear budget accordingly; no code change required.

**Fix:** No change required; the per-cycle bound and documentation are adequate. Optionally consider coalescing the FLAGS write (e.g., persist only on threshold transitions) on extremely wear-sensitive flash backends, and keep the FWTPM_DA_DEFAULT_MAX_TRIES default modest.

dgarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review-security`**Overall recommendation:** `APPROVE`
**Findings:** 2 total — 2 posted, 0 skipped
**2 finding(s) posted as inline comments** (see file-level comments below)

> One or more scans did not complete — the results below are partial.

### Posted findings
- [Low] [review-security] **lockoutAuth recovery timer restarts on every Startup, so clocked builds measure per-boot uptime rather than elapsed wall-clock** — `src/fwtpm/fwtpm_command.c:695-704`
- [Info] [review-security] **Failed-auth path persists FLAGS to NV on every failure (flash wear / NV-bound auth latency)** — `src/fwtpm/fwtpm_command.c:15903-15916`

dgarske (approved):

PR #546, opened 2026-07-02
TPM locality control, fTPM per-PCR enforcement fix, and optional reset HAL

dgarske:
## Summary

Adds runtime TPM locality selection, fixes the fwTPM's per-PCR locality enforcement, and adds an optional nRST reset HAL. Verified on a real ST33KTPM2X32 (SPI, Raspberry Pi 5) and the fwTPM software TPM.

## Features

- **`wolfTPM2_SetLocality(dev, locality)`** switches the active TPM locality at runtime, so PCRs that are not resettable at locality 0 (20-22 at loc 2-4, 17-19 at loc 4) become reachable. Works on the built-in TIS/SPI driver (ACCESS handshake with release-and-retry for non-preempting chips) and on the fwTPM over both its socket and TIS/SHM transports; returns `NOT_COMPILED_IN` where locality is not selectable (I2C, Linux kernel driver, Windows TBS).
- **`examples/pcr/reset -loc=n`** selects the locality (validated 0-4) before reset.
- **`--enable-hal-reset[=LINE]`** adds `TPM2_IoCb_Reset()`, pulsing the TPM nRST line via the Linux GPIO char device (raw v2 uAPI, no libgpiod). Default line: ST33 GPIO24 (Pi pin 18), Nuvoton GPIO4; the `=LINE` argument is validated as numeric.

## Fixes

- **fwTPM per-PCR locality enforcement** replaced with one source-of-truth table (TCG PC Client profile, matching `ibmswtpm2` `s_initAttributes` and a real ST33): corrects the reset map (PCR 18/19 wrongly allowed loc 3; 20-22 wrongly rejected loc 2), adds the missing PCR-extend locality check, and makes `TPM2_GetCapability(TPM_CAP_PCR_PROPERTIES)` report proper `RESET_L*`/`EXTEND_L*`/`DRTM_RESET` bitmaps.
  - Behavior change: DRTM PCRs 17-22 can no longer be extended from locality 0 (now `TPM_RC_LOCALITY`), per the profile.
- **fwTPM `TPM2_GetCapability` paging** for `TPM_CAP_ALGS`/`TPM_CAP_COMMANDS` now honors the property cursor and emits ascending, so a client that follows `moreData` makes progress instead of looping.

## Testing

fwTPM unit tests (reset/extend map, capability report, GetCapability paging), a `wolfTPM2_SetLocality` argument test, and an end-to-end `fwtpm_check.sh` locality check run in CI across socket and TIS/SHM; a build job compiles `--enable-hal-reset`. Hardware-verified on a real ST33KTPM2X32 over SPI (locality switching, the per-PCR reset matrix, and the nRST HAL).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -474,6 +474,34 @@ then
     AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_ST33"
 fi
 
+# Optional TPM hardware reset (nRST) HAL control via Linux GPIO char device.
+# Accepts an optional GPIO line number (default ST33 GPIO24 / Pi pin 18, Nuvoton GPIO4).
+AC_ARG_ENABLE([hal-reset],
+    [AS_HELP_STRING([--enable-hal-reset@<:@=LINE@:>@],[Enable TPM nRST reset HAL (optional GPIO line; default ST33 GPIO24, Nuvoton GPIO4) (default: disabled)])],
+    [ ENABLED_HAL_RESET=$enableval ],
+    [ ENABLED_HAL_RESET=no ]
+    )
+if test "x$ENABLED_HAL_RESET" != "xno"
+then
+    # The reset HAL implementation (hal/tpm_io.c) only builds for the SPI/I2C
+    # example HAL, not the native kernel (/dev/tpm0) or swtpm transports, so
+    # reject those combinations rather than report "yes" with no callable API.
+    # Note swtpm is the default on Linux x86_64/aarch64, so the hardware reset
+    # HAL requires explicitly selecting a hardware interface (--enable-spi/-i2c).
+    if test "x$ENABLED_DEVTPM" = "xyes" || test "x$ENABLED_SWTPM" = "xyes" || test "x$ENABLED_SWTPM" = "xuart"
+    then
+        AC_MSG_ERROR([--enable-hal-reset needs the SPI/I2C hardware HAL; it is not available with the swtpm simulator (default on Linux x86_64/aarch64) or --enable-devtpm. Select a hardware interface, e.g. --enable-spi or --enable-i2c.])
+    fi
+    AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_HAL_RESET"
+    case "x$ENABLED_HAL_RESET" in
+        xyes) ;;
+        x|x*[[!0-9]]*)
+            AC_MSG_ERROR([--enable-hal-reset=LINE requires a numeric GPIO line number (e.g. --enable-hal-reset=24).]) ;;
+        *) AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_RESET_LINE=$ENABLED_HAL_RESET" ;;
+    esac
+    ENABLED_HAL_RESET=yes
+fi
+
 # Microchip ATTPM20
 AC_ARG_ENABLE([mchp],,
     [ ENABLED_MCHP=$enableval ],
@@ -1132,6 +1160,7 @@ if test "x$ENABLED_SWTPM" = "xyes"; then
 fi
 echo "   * WINAPI:                    $ENABLED_WINAPI"
 echo "   * TIS/SPI Check Wait State:  $ENABLED_CHECKWAITSTATE"
+echo "   * HAL Reset (nRST GPIO):     $ENABLED_HAL_RESET"
 
 echo "   * Infineon SLB967X           $ENABLED_INFINEON"
 echo "   * STM ST33:                  $ENABLED_ST"
--- a/examples/pcr/README.md
+++ b/examples/pcr/README.md
@@ -36,6 +36,8 @@ Not all PCRs are equal. The user can perform `extend` operation on all PCRs, but
 * PCR16 is a PCR for debug purposes. This is the PCR used by all tools above by default. It is safe to test and work with PCR16.
 * PCR17-22 are reserved for Dynamic Root of Trust Measurement (DRTM), an advanced topic that is to be covered separately.
 
+Reset locality (TCG PC Client): PCR16/23 reset at localities 0-3, PCR20-22 at locality 2-4, and PCR17-19 at locality 4. Use `-loc=n` to select the locality (see `wolfTPM2_SetLocality`); it applies to the built-in TIS/SPI driver and the fwTPM.
+
 ### Extend
 
 The TPM 2.0 `TPM2_Extend` API uses a SHA1 or SHA256 cryptographic operation to combine the current value of the PCR and with newly provided hash digest.
@@ -50,10 +52,13 @@ The TPM 2.0 `TPM2_Quote` API is a standard operation that encapsulates the PCR d
 
 ```sh
 $ ./examples/pcr/reset -?
-PCR index is out of range (0-23)
+Incorrect arguments
 Expected usage:
-./examples/pcr/reset [pcr]
+./examples/pcr/reset [pcr] [-loc=n]
 * pcr is a PCR index between 0-23 (default 16)
+* -loc=n switch to TPM locality n (0-4) before reset
+    (PCR 17-19 need locality 4; 20-22 need locality 2-4;
+     enforced by the fwTPM and by discrete TPMs like the ST33)
 Demo usage without parameters, resets PCR16.
 ```
 
--- a/examples/pcr/reset.c
+++ b/examples/pcr/reset.c
@@ -43,14 +43,19 @@
 static void usage(void)
 {
     printf("Expected usage:\n");
-    printf("./examples/pcr/reset [pcr]\n");
+    printf("./examples/pcr/reset [pcr] [-loc=n]\n");
     printf("* pcr is a PCR index between 0-23 (default %d)\n", TPM2_TEST_PCR);
+    printf("* -loc=n switch to TPM locality n (0-4) before reset\n");
+    printf("    (PCR 17-19 need locality 4; 20-22 need locality 2-4;\n");
+    printf("     enforced by the fwTPM and by discrete TPMs like the ST33)\n");
     printf("Demo usage without parameters, resets PCR%d.\n", TPM2_TEST_PCR);
 }
 
 int TPM2_PCR_Reset_Test(void* userCtx, int argc, char *argv[])
 {
     int i, j, pcrIndex = TPM2_TEST_PCR, rc = -1;
+    int locality = -1; /* -1 means leave at default locality */
+    int localitySwitched = 0;
     WOLFTPM2_DEV dev;
 
     union {
@@ -63,21 +68,31 @@ int TPM2_PCR_Reset_Test(void* userCtx, int argc, char *argv[])
         byte maxOutput[MAX_RESPONSE_SIZE];
     } cmdOut;
 
-    if (argc == 2) {
-        pcrIndex = XATOI(argv[1]);
-        if (pcrIndex < 0 || pcrIndex > 23 || *argv[1] < '0' || *argv[1] > '9') {
-            printf("PCR index is out of range (0-23)\n");
+    /* Parse arguments: an optional PCR index and an optional -loc=n flag */
+    for (i = 1; i < argc; i++) {
+        if (XSTRNCMP(argv[i], "-loc=", 5) == 0) {
+            /* Require exactly one digit 0-4; XATOI would silently map
+             * non-numeric input (e.g. -loc=abc or -loc=) to locality 0. */
+            if (argv[i][5] < '0' || argv[i][5] > '4' || argv[i][6] != '\0') {
+                printf("Locality is out of range (0-4)\n");
+                usage();
+                goto exit_badargs;
+            }
+            locality = argv[i][5] - '0';
+        }
+        else if (argv[i][0] >= '0' && argv[i][0] <= '9') {
+            pcrIndex = XATOI(argv[i]);
+            if (pcrIndex < 0 || pcrIndex > 23) {
+                printf("PCR index is out of range (0-23)\n");
+                usage();
+                goto exit_badargs;
+            }
+        }
+        else {
+            printf("Incorrect arguments\n");
             usage();
             goto exit_badargs;
         }
-    }
-    else if (argc == 1) {
-        pcrIndex = TPM2_TEST_PCR;
-    }
-    else {
-        printf("Incorrect arguments\n");
-        usage();
-        goto exit_badargs;
     }
 
     printf("Demo how to reset a PCR (clear PCR value)\n");
@@ -87,6 +102,18 @@ int TPM2_PCR_Reset_Test(void* userCtx, int argc, char *argv[])
         goto exit;
     }
     printf("wolfTPM2_Init: success\n");
+
+    if (locality >= 0) {
+        printf("Switching to locality %d...\n", locality);
+        rc = wolfTPM2_SetLocality(&dev, locality);
+        if (rc != TPM_RC_SUCCESS) {
+            printf("wolfTPM2_SetLocality(%d) failed 0x%x: %s\n", locality, rc,
+                TPM2_GetRCString(rc));
+            goto exit;
+        }
+        printf("wolfTPM2_SetLocality: now at locality %d\n", locality);
+        localitySwitched = 1;
+    }
 
     /* Prepare PCR Reset command */
     XMEMSET(&cmdIn.pcrReset, 0, sizeof(cmdIn.pcrReset));
@@ -115,6 +142,13 @@ int TPM2_PCR_Reset_Test(void* userCtx, int argc, char *argv[])
     }
 
 exit:
+
+    /* Return to the default locality so a non-zero one is not left active
+     * (which would block the next TPM startup on TPMs that do not preempt).
+     * Only when a switch actually succeeded - never on a failed Init. */
+    if (localitySwitched) {
+        (void)wolfTPM2_SetLocality(&dev, 0);
+    }
 
     wolfTPM2_Cleanup(&dev);
 
--- a/examples/run_examples.sh
+++ b/examples/run_examples.sh
@@ -1184,6 +1184,22 @@ echo -e "Endorsement Key (EK) and Certificate"
 RESULT=$?
 [ $RESULT -ne 0 ] && echo -e "get_ek_certs failed! $RESULT" && exit 1
 
+# PCR reset locality (-loc): smoke-exercise wolfTPM2_SetLocality. Reset PCR 16
+# at locality 0 (universally resettable, a no-op switch). This runs against
+# whichever backend is present (ibmswtpm2, fwTPM, ...) so it does not assert the
+# backend-specific per-PCR map - that rigor lives in tests/fwtpm_check.sh.
+echo -e "PCR reset locality (-loc)"
+LOC_OUT=$(./examples/pcr/reset 16 -loc=0 2>&1)
+echo "$LOC_OUT" >> $TPMPWD/run.out
+if echo "$LOC_OUT" | grep -q "TPM2_PCR_Reset success"; then
+    : # -loc supported and works
+elif echo "$LOC_OUT" | grep -qiE "not compiled in|NOT_COMPILED_IN"; then
+    # kernel (/dev/tpm0) and Windows TBS own the locality; TPM2_GetRCString
+    # renders NOT_COMPILED_IN as "Feature not compiled in" under wolfCrypt.
+    echo -e "  -loc not supported by this backend, skipping"
+else
+    echo -e "pcr reset -loc failed!" && exit 1
+fi
 
 rm -f keyblob.bin
 
--- a/hal/tpm_io.c
+++ b/hal/tpm_io.c
@@ -239,6 +239,29 @@ int TPM2_IoCb(TPM2_CTX* ctx, const BYTE* txBuf, BYTE* rxBuf,
 }
 
 #endif /* WOLFTPM_ADV_IO */
+
+#ifdef WOLFTPM_HAL_RESET
+/* Pulse the TPM hardware reset (nRST) line to reset the TPM. Dispatches to the
+ * platform implementation. Returns TPM_RC_SUCCESS on success. */
+int TPM2_IoCb_Reset(TPM2_CTX* ctx, void* userCtx)
+{
+    int ret;
+#if defined(__linux__)
+    ret = TPM2_IoCb_Linux_Reset(ctx, userCtx);
+#else
+    /* No reset HAL for this OS - return NOT_COMPILED_IN so callers can tell it
+     * apart from a genuine reset failure (TPM_RC_FAILURE) on a supported OS. */
+    (void)ctx;
+    (void)userCtx;
+    ret = NOT_COMPILED_IN;
+    #ifdef DEBUG_WOLFTPM
+    printf("TPM reset HAL not implemented for this platform\n");
+    #endif
+#endif
+    return ret;
+}
+#endif /* WOLFTPM_HAL_RESET */
+
 #endif /* !(WOLFTPM_LINUX_DEV || WOLFTPM_SWTPM || WOLFTPM_WINAPI) */
 
 /******************************************************************************/
--- a/hal/tpm_io_linux.c
+++ b/hal/tpm_io_linux.c
@@ -64,6 +64,11 @@
     #include <fcntl.h>
     #include <unistd.h>
     #include <errno.h>
+    #ifdef WOLFTPM_HAL_RESET
+        /* GPIO character-device uAPI for optional nRST control */
+        #include <linux/gpio.h>
+        #include <time.h> /* nanosleep (usleep is undefined for >= 1s) */
+    #endif
 
     #ifdef WOLFTPM_I2C
         /* I2C - (Only tested with SLB9673 and ST33 I2C) */
@@ -415,6 +420,110 @@
         return ret;
     }
 #endif /* WOLFTPM_I2C */
+
+#ifdef WOLFTPM_HAL_RESET
+    /* Pulse the TPM nRST (active low) via the Linux GPIO char device (raw GPIO
+     * v2 uAPI, no libgpiod). Default line: Raspberry Pi ST33 = GPIO24 (pin 18),
+     * Nuvoton = GPIO4; override with WOLFTPM_RESET_GPIOCHIP / WOLFTPM_RESET_LINE. */
+    #ifndef WOLFTPM_RESET_GPIOCHIP
+        #define WOLFTPM_RESET_GPIOCHIP "/dev/gpiochip0"
+    #endif
+    #ifndef WOLFTPM_RESET_LINE
+        #if defined(WOLFTPM_NUVOTON)
+            #define WOLFTPM_RESET_LINE 4
+        #else
+            #define WOLFTPM_RESET_LINE 24
+        #endif
+    #endif
+    #ifndef WOLFTPM_RESET_HOLD_US
+        #define WOLFTPM_RESET_HOLD_US 300000    /* reset asserted 300ms */
+    #endif
+    #ifndef WOLFTPM_RESET_SETTLE_US
+        #define WOLFTPM_RESET_SETTLE_US 1000000 /* TPM boot settle 1s */
+    #endif
+
+    /* usleep() is undefined for values >= 1000000 (POSIX); nanosleep has no
+     * such limit and handles the 1s settle and any larger override. */
+    static void TPM2_Reset_DelayUs(unsigned long us)
+    {
+        struct timespec ts;
+        ts.tv_sec  = (time_t)(us / 1000000UL);
+        ts.tv_nsec = (long)((us % 1000000UL) * 1000UL);
+        (void)nanosleep(&ts, NULL);
+    }
+
+    /* Note: this reset HAL is only compile-checked in CI (no GPIO hardware or
+     * gpio-sim there); the open/GET_LINE/SET_VALUES flow, the hold/settle
+     * timing, and the fd lifecycle are functionally regression-verified on real
+     * hardware - the ST33 on a Raspberry Pi 5, nRST wired to GPIO24 (pin 18). */
+    int TPM2_IoCb_Linux_Reset(TPM2_CTX* ctx, void* userCtx)
+    {
+        int ret = TPM_RC_FAILURE;
+        int chipFd, reqFd;
+        struct gpio_v2_line_request req;
+        struct gpio_v2_line_values vals;
+
+        (void)ctx;
+        (void)userCtx;
+
+        chipFd = open(WOLFTPM_RESET_GPIOCHIP, O_RDONLY);
+        if (chipFd < 0) {
+        #ifdef DEBUG_WOLFTPM
+            printf("TPM Reset: open %s failed (errno %d)\n",
+                WOLFTPM_RESET_GPIOCHIP, errno);
+        #endif
+            return TPM_RC_FAILURE;
+        }
+
+        /* Acquire the line as an output driven low (assert reset) */
+        XMEMSET(&req, 0, sizeof(req));
+        req.offsets[0] = (unsigned int)WOLFTPM_RESET_LINE;
+        req.num_lines = 1;
+        req.config.flags = GPIO_V2_LINE_FLAG_OUTPUT;
+        req.config.num_attrs = 1;
+        req.config.attrs[0].attr.id = GPIO_V2_LINE_ATTR_ID_OUTPUT_VALUES;
+        req.config.attrs[0].attr.values = 0; /* drive low (assert reset) */
+        req.config.attrs[0].mask = 1;        /* applies to line index 0 */
+        XMEMCPY(req.consumer, "wolfTPM-reset", sizeof("wolfTPM-reset"));
+
+        if (ioctl(chipFd, GPIO_V2_GET_LINE_IOCTL, &req) < 0 || req.fd < 0) {
+        #ifdef DEBUG_WOLFTPM
+            printf("TPM Reset: GET_LINE ioctl failed (errno %d)\n", errno);
+        #endif
+            close(chipFd);
+            return TPM_RC_FAILURE;
+        }
+        close(chipFd);
+        reqFd = req.fd;
+
+        /* Hold reset asserted, then release (drive high) and let the TPM boot */
+        TPM2_Reset_DelayUs(WOLFTPM_RESET_HOLD_US);
+
+        XMEMSET(&vals, 0, sizeof(vals));
+        vals.mask = 1;
+        vals.bits = 1; /* drive high = release reset */
+        if (ioctl(reqFd, GPIO_V2_LINE_SET_VALUES_IOCTL, &vals) < 0) {
+        #ifdef DEBUG_WOLFTPM
+            printf("TPM Reset: SET_VALUES ioctl failed (errno %d)\n", errno);
+        #endif
+        }
+        else {
+            ret = TPM_RC_SUCCESS;
+        #ifdef DEBUG_WOLFTPM
+            printf("TPM Reset: pulsed nRST on %s line %d\n",
+                WOLFTPM_RESET_GPIOCHIP, (int)WOLFTPM_RESET_LINE);
+        #endif
+            /* Only wait for the TPM to settle after a successful release; on a
+             * failed release the delay would just stall the error path. */
+            TPM2_Reset_DelayUs(WOLFTPM_RESET_SETTLE_US);
+        }
+
+        close(reqFd);
+
+        return ret;
+    }
+#endif /* WOLFTPM_HAL_RESET */
+
 #endif /* __linux__ */
 #endif /* !(WOLFTPM_LINUX_DEV || WOLFTPM_SWTPM || WOLFTPM_WINAPI) */
 #endif /* WOLFTPM_INCLUDE_IO_FILE */
--- a/src/fwtpm/fwtpm_command.c
+++ b/src/fwtpm/fwtpm_command.c
@@ -1076,6 +1076,30 @@ static TPM_RC FwCmd_StirRandom(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
     return rc;
 }
 
+/* Per-PCR reset/extend locality helper (defined with its tables below). */
+static int FwPcrLocalityAllowed(int pcrIndex, int locality, int isReset);
+
+/* Mask a TPM command code to its 16-bit value (strip vendor/reserved bits). */
+#define FW_CC_MASK 0x0000FFFFu
+/* PCRs 0-15 are the SRTM set that TPM_PT_PCR_SAVE reports as saved. */
+#define FWTPM_PCR_SAVE_COUNT 16
+/* pcrProps[].kind: how a TPM_CAP_PCR_PROPERTIES row maps to the locality table. */
+#define FW_PCR_KIND_EXTEND      0  /* extend allowed at loc */
+#define FW_PCR_KIND_RESET       1  /* reset allowed at loc */
+#define FW_PCR_KIND_SAVE        2  /* saved across power cycles (PCR 0-15) */
+#define FW_PCR_KIND_DRTM_RESET  3  /* reset allowed at locality 4 */
+
+/* Overwrite a big-endian UINT32 already appended at buf[pos]. Used to
+ * back-patch a TPML count with the number of entries actually emitted, so a
+ * count/payload mismatch is impossible even if two entries collapse to one. */
+static void FwPatchU32BE(byte* buf, int pos, UINT32 v)
+{
+    buf[pos + 0] = (byte)(v >> 24);
+    buf[pos + 1] = (byte)(v >> 16);
+    buf[pos + 2] = (byte)(v >> 8);
+    buf[pos + 3] = (byte)(v);
+}
+
 /* --- TPM2_GetCapability (CC 0x017A) --- */
 static TPM_RC FwCmd_GetCapability(FWTPM_CTX* ctx, TPM2_Packet* cmd,
     int cmdSize, TPM2_Packet* rsp, UINT16 cmdTag)
@@ -1086,6 +1110,7 @@ static TPM_RC FwCmd_GetCapability(FWTPM_CTX* ctx, TPM2_Packet* cmd,
     UINT32 propertyCount = 0;
     UINT32 i;
     int paramSzPos, paramStart;
+    int moreDataPos;
 
     (void)ctx;
 
@@ -1108,7 +1133,9 @@ static TPM_RC FwCmd_GetCapability(FWTPM_CTX* ctx, TPM2_Packet* cmd,
 
     paramStart = FwRspParamsBegin(rsp, cmdTag, &paramSzPos);
 
-    /* moreData (TPMI_YES_NO) */
+    /* moreData (TPMI_YES_NO) - default NO; a capability case that truncates its
+     * list to propertyCount back-patches this byte to YES via moreDataPos. */
+    moreDataPos = rsp->pos;
     TPM2_Packet_AppendU8(rsp, 0); /* NO */
 
     /* capability */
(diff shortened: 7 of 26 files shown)


src/tpm2_wrap.c
```diff
@@ -6684,6 +6685,72 @@ int wolfTPM2_ResetPCR(WOLFTPM2_DEV* dev, int pcrIndex)
     return rc;
 }
 
+int wolfTPM2_SetLocality(WOLFTPM2_DEV* dev, int locality)
+{
+#if defined(WOLFTPM_LINUX_DEV) || defined(WOLFTPM_WINAPI)
+    /* The Linux kernel driver and Windows TBS own the TPM locality */
+    (void)dev;
+    (void)locality;
```
aidangarske, line 6710: **🔴 [High] SWTPM locality switch serializes the wrong byte on big-endian hosts** · _Logic_

wolfTPM2_SetLocality() records nonzero localities in dev->ctx.locality, declared int (wolftpm/tpm2.h:2209). The existing SWTPM send path (src/tpm2_swtpm.c:553) transmits one byte from &ctx->locality via sizeof(uint8_t). This was latent while locality stayed 0, but the PR now makes 1-4 valid runtime state. On big-endian hosts &ctx->locality points at the most-significant byte (0x00), so localities 1-4 serialize as 0x00 and commands requested for locality 2 or 4 are sent as locality 0 — silently breaking the advertised swtpm/fwTPM socket locality support and letting PCR reset/extend policy checks run under the wrong locality. Severity views: review=High/BLOCK, review-security=Medium; stricter kept.

**Fix:** Serialize through a one-byte temporary, e.g. `byte loc = (byte)ctx->locality; SwTpmTransmit(ctx, &loc, sizeof(loc));`, or store locality in a byte-sized field. Add a regression test verifying locality 2/4 framing is independent of host endianness.

src/fwtpm/fwtpm_command.c
```diff
@@ -1775,6 +1816,59 @@ static TPM_RC FwCmd_PCR_Read(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
...
 /* --- TPM2_PCR_Extend (CC 0x0182) --- */
+/* Per-PCR locality bitmaps (bit L set = locality L may reset/extend the PCR),
+ * per the TCG PC Client profile. Single source of truth for reset/extend
+ * enforcement and the TPM_CAP_PCR_PROPERTIES report.
+ *   reset:  0-15 never; 16,23 loc0-3; 17-19 loc4; 20-22 loc2-4.
+ *   extend: 0-16,23 any; 17,18 loc2-4; 19 loc2-3; 20 loc1-3; 21,22 loc2. */
+#define FW_LOC(l)  (1u << (l))
+#define FW_LOC_ALL (FW_LOC(0) | FW_LOC(1) | FW_LOC(2) | FW_LOC(3) | FW_LOC(4))
+
+static const byte fwPcrResetLocality[IMPLEMENTATION_PCR] = {
```
aidangarske, line 1934: **🔴 [High] fwTPM PCR reset locality table diverges from TCG PC Client reference (PCR 16/23 lose loc4; PCR 20-22 wrongly gain loc3)** · _Correctness_

fwPcrResetLocality is the single source of truth for reset enforcement AND the TPM_CAP_PCR_PROPERTIES report. The sibling fwPcrExtendLocality table matches the TPM reference s_initAttributes extend column exactly, confirming the intended reference is the standard PC Client profile. Against that same reference the reset column should be PCR16=all(0-4), PCR17-19=loc4, PCR20-22=loc2 and loc4 only, PCR23=all(0-4). The table instead encodes PCR16/23=loc0-3 (dropping locality 4) and PCR20-22=loc2,3,4 (adding locality 3). Two defects: (1) PCR16/23 can no longer be reset from locality 4 (FW_LOC(4) appears omitted); (2) PCR20-22 become resettable from locality 3, a security-relevant over-permission in the DRTM path this PR locks down. The reset map also feeds RESET_L3/RESET_L4 in the capability report and the unit tests encode these divergent values, so CI passes while enforcing the wrong policy.

**Fix:** Cross-check the reset bitmaps against the TPM 2.0 reference PCR.c s_initAttributes: PCR16/23 reset should include locality 4, and PCR20-22 reset should be locality 2 and 4 only (not 3). Update the test_fwtpm_pcr_reset_locality_enforced / test_fwtpm_pcr_properties_capability assertions that encode the divergent values.

examples/run_examples.sh
```diff
@@ -1184,6 +1184,20 @@ echo -e "Endorsement Key (EK) and Certificate"
...
 [ $RESULT -ne 0 ] && echo -e "get_ek_certs failed! $RESULT" && exit 1
 
+# PCR reset locality (-loc): smoke-exercise wolfTPM2_SetLocality. Reset PCR 16
+# at locality 0 (universally resettable, a no-op switch). This runs against
+# whichever backend is present (ibmswtpm2, fwTPM, ...) so it does not assert the
+# backend-specific per-PCR map - that rigor lives in tests/fwtpm_check.sh.
+echo -e "PCR reset locality (-loc)"
+LOC_OUT=$(./examples/pcr/reset 16 -loc=0 2>&1)
+echo "$LOC_OUT" >> $TPMPWD/run.out
+if echo "$LOC_OUT" | grep -q "TPM2_PCR_Reset success"; then
```
aidangarske, line 1194: **🔴 [High] run_examples.sh -loc skip branch greps for a token the program never prints (unsupported backends fail CI)** · _Logic_

The new -loc smoke test tries to skip unsupported backends by grepping for the literal string NOT_COMPILED_IN, but reset.c prints TPM2_GetRCString(rc). In a wolfCrypt build, TPM2_GetRCString routes negative codes through wc_GetErrorString()/wolfSSL_ERR_reason_error_string(), returning the human string "Feature not compiled in" — the literal token NOT_COMPILED_IN is only emitted under WOLFTPM2_NO_WOLFCRYPT. So on backends where wolfTPM2_SetLocality returns NOT_COMPILED_IN (WOLFTPM_LINUX_DEV /dev/tpm0, Windows TBS), neither the success string nor the token matches, control falls to the failure branch and run_examples.sh exits 1. The sibling test in tests/fwtpm_check.sh greps TPM_RC_LOCALITY (a positive RC the local switch stringifies), so only the negative NOT_COMPILED_IN path is broken. Severity views: review=High/BLOCK, review-security=Low; stricter kept.

**Fix:** Match the actual emitted text, e.g. `grep -qiE "not compiled in|NOT_COMPILED_IN"`, or have reset.c print a stable machine-parseable marker that both the human string and the script agree on.

examples/pcr/reset.c
```diff
@@ -43,14 +43,17 @@
 static void usage(void)
 {
     printf("Expected usage:\n");
-    printf("./examples/pcr/reset [pcr]\n");
+    printf("./examples/pcr/reset [pcr] [-loc=n]\n");
     printf("* pcr is a PCR index between 0-23 (default %d)\n", TPM2_TEST_PCR);
+    printf("* -loc=n switch to TPM locality n (0-4) before reset\n");
```
aidangarske, line 48: **🟠 [Medium] PCR reset usage text contradicts the implemented locality map and supported transports** · _Documentation_

The help text says `PCR 17-22 require locality 4; only supported on TIS/SPI`, but the source-of-truth reset map (fwPcrResetLocality, src/fwtpm/fwtpm_command.c) allows PCR20-22 at localities other than 4, and the PR explicitly adds swtpm/fwTPM socket transport locality selection. Both halves of the sentence mislead users about the exact feature the PR adds — they may choose the wrong locality or assume socket-mode backends are unsupported. (Remains inaccurate even after the reset-table fix, since PCR20-22 would then be loc2 and loc4, still not simply "locality 4".)

**Fix:** Reword to match the enforced map and supported transports, e.g. describe PCR17-19 as locality 4 and PCR20-22 as their implemented allowed localities, and mention swtpm/fwTPM socket support.

src/tpm2_tis.c
```diff
@@ -330,6 +334,45 @@ int TPM2_TIS_RequestLocality(TPM2_CTX* ctx, int timeout)
     return rc;
 }
 
+int TPM2_TIS_RequestLocality(TPM2_CTX* ctx, int timeout)
```
aidangarske, line 335: **🟠 [Medium] Startup locality self-heal can release a live non-default owner** · _Security_

Before this PR, startup requested the default locality and timed out if another locality was active. The new self-heal loop scans localities 0-4 and, for any active locality other than the default, writes TPM_ACCESS_ACTIVE_LOCALITY to release it before requesting locality 0. There is no check that the active locality is stale or was acquired by the same TPM2_CTX, so a second process or thread calling init can clear a legitimate higher-locality owner. On direct TIS/SPI deployments this can disrupt DRTM/locality-sensitive operations and weakens locality-ownership semantics.

**Fix:** Do not clear arbitrary active non-default localities during normal startup. Make stale-locality recovery an explicit opt-in path, or only release a locality this context previously acquired and tracked; otherwise return the timeout and let the caller decide.

src/tpm2_tis.c
```diff
@@ -286,18 +286,22 @@ int TPM2_TIS_CheckLocality(TPM2_CTX* ctx, int locality, byte* access)
...
+        (access & (TPM_ACCESS_ACTIVE_LOCALITY | TPM_ACCESS_VALID)) ==
                   (TPM_ACCESS_ACTIVE_LOCALITY | TPM_ACCESS_VALID)) {
         ctx->locality = locality;
         return locality;
     }
     return -1;
 }
 
-int TPM2_TIS_RequestLocality(TPM2_CTX* ctx, int timeout)
+int TPM2_TIS_RequestLocalityEx(TPM2_CTX* ctx, int locality, int timeout)
```
aidangarske, line 300: **🟠 [Medium] I2C builds can falsely report nonzero locality as active** · _API Contract Violations_

The new runtime locality path is used for all non-dev/non-swtpm TIS builds, including WOLFTPM_I2C (only WOLFTPM_LINUX_DEV, WOLFTPM_WINAPI, WOLFTPM_SWTPM are excluded). TIS macros encode locality in the upper address bits: TPM_ACCESS(l) = (TPM_BASE_ADDRESS | 0x0004u | ((l) << 12u)). But the I2C HALs transmit only reg & 0xFF (hal/tpm_io_linux.c:131,170), so the (l<<12) bits never reach the TPM. A read of TPM_ACCESS(2) addresses the same lower-8-bit register (0x04) as TPM_ACCESS(0). If locality 0 is active (default after init), wolfTPM2_SetLocality(dev, 2) reads locality-0 ACCESS, CheckLocalityAccessValid sees ACTIVE_LOCALITY|VALID, treats it as a grant, sets ctx->locality=2 and returns success — while subsequent I2C commands still address locality 0 on the bus. CI I2C jobs are compile-only, so this is not exercised.

**Fix:** Return NOT_COMPILED_IN for nonzero locality on WOLFTPM_I2C unless locality-aware I2C addressing is implemented, e.g. guard in wolfTPM2_SetLocality() before the generic TIS handshake.

src/tpm2_wrap.c
```diff
@@ -6684,6 +6685,72 @@ int wolfTPM2_ResetPCR(WOLFTPM2_DEV* dev, int pcrIndex)
...
+#elif defined(WOLFTPM_SWTPM)
+    /* The swtpm/mssim transport carries the locality with each command
+     * (see TPM2_SWTPM_SendCommand), so just record it for subsequent
+     * commands - no TIS handshake is involved. */
+    if (dev == NULL || locality < 0 || locality > 4) {
+        return BAD_FUNC_ARG;
+    }
+    dev->ctx.locality = locality;
+    return TPM_RC_SUCCESS;
+#else
```
aidangarske, line 6721: **🟠 [Medium] wolfTPM2_SetLocality TIS preempt/release/restore branch has no automated coverage** · _Test Coverage_

The #else (built-in TIS driver) branch is the most intricate new logic in the PR: it attempts the target locality, on failure releases the current locality and retries, and on second failure re-requests the old locality, plus an asymmetric else-if release-after-preemption path. This is only reachable on real TIS/SPI hardware, so it is not exercised by the fwtpm unit tests or fwtpm_check.sh (which drive the SWTPM/mssim record-only branch). Given the subtle success semantics (rc == locality means success), a regression here would be silent in CI.

**Fix:** Add a mock/loopback TIS test (or a simulated ACCESS register in the fwTPM TIS harness) exercising switch-up, switch-down, no-preempt-then-release, and restore-on-failure, or document that this branch is hardware-only and was manually verified on the ST33.

configure.ac
```diff
@@ -474,6 +474,23 @@ then
...
 fi
 
+# Optional TPM hardware reset (nRST) HAL control via Linux GPIO char device.
+# Accepts an optional GPIO line number (default ST33 GPIO24 / Pi pin 18, Nuvoton GPIO4).
+AC_ARG_ENABLE([hal-reset],
+    [AS_HELP_STRING([--enable-hal-reset@<:@=LINE@:>@],[Enable TPM nRST reset HAL (optional GPIO line; default ST33 GPIO24, Nuvoton GPIO4) (default: disabled)])],
+    [ ENABLED_HAL_RESET=$enableval ],
+    [ ENABLED_HAL_RESET=no ]
+    )
+if test "x$ENABLED_HAL_RESET" != "xno"
```
aidangarske, line 484: **🟠 [Medium] hal-reset can be reported enabled without exporting the reset API** · _Build_

--enable-hal-reset only adds -DWOLFTPM_HAL_RESET. The TPM2_IoCb_Reset() prototype in hal/tpm_io.h is guarded solely by WOLFTPM_HAL_RESET, but its implementation in hal/tpm_io.c (243-260) sits inside #if !(WOLFTPM_LINUX_DEV || WOLFTPM_SWTPM || WOLFTPM_WINAPI). So a native-transport build such as --enable-devtpm --enable-hal-reset compiles the declaration and prints 'HAL Reset (nRST GPIO): yes' while no callable TPM2_IoCb_Reset() symbol is emitted, yielding a link error if anything references it.

**Fix:** Reject incompatible transport combinations in configure, force the HAL implementation on when --enable-hal-reset is selected, or make the summary report the API as unavailable for those transports.

src/fwtpm/fwtpm_command.c
```diff
@@ -1355,24 +1358,62 @@ static TPM_RC FwCmd_GetCapability(FWTPM_CTX* ctx, TPM2_Packet* cmd,
...
+            for (startIdx = 0; startIdx < totalProps; startIdx++) {
+                if (pcrProps[startIdx].tag >= property)
+                    break;
             }
-            if (numBanks > 1) {
-                TPM2_Packet_AppendU32(rsp, TPM_ALG_SHA384);
+            numOut = totalProps - startIdx;
+            if (numOut < 0)
+                numOut = 0;
+            if ((UINT32)numOut > propertyCount)
```
aidangarske, line 1495: **🔵 [Low] TPM_CAP_PCR_PROPERTIES always reports moreData=NO even when truncated by propertyCount** · _Conformance_

The PCR_PROPERTIES case caps numOut to propertyCount (1394-1395), but the shared moreData byte emitted before the switch is hardcoded to NO (line 1115: TPM2_Packet_AppendU8(rsp, 0)). If a client requests fewer properties than available (or starts mid-list), additional properties are silently not advertised, so a client looping on GetCapability via moreData would stop early. wolfTPM's own callers request all properties at once, so impact is limited, but the report is not strictly conformant.

**Fix:** Set moreData=YES when startIdx + numOut < totalProps, or document that the fwTPM does not implement moreData paging for capabilities.

examples/pcr/reset.c
```diff
@@ -63,22 +66,30 @@ int TPM2_PCR_Reset_Test(void* userCtx, int argc, char *argv[])
         byte maxOutput[MAX_RESPONSE_SIZE];
     } cmdOut;
 
-    if (argc == 2) {
-        pcrIndex = XATOI(argv[1]);
-        if (pcrIndex < 0 || pcrIndex > 23 || *argv[1] < '0' || *argv[1] > '9') {
-            printf("PCR index is out of range (0-23)\n");
+    /* Parse arguments: an optional PCR index and an optional -loc=n flag */
```
aidangarske, line 71: **🔵 [Low] Invalid -loc values are accepted as locality 0** · _Error Handling_

The -loc=n parser uses XATOI() and only range-checks the converted integer. XATOI() maps to atoi(), which returns 0 for non-numeric input, so -loc=abc and -loc= both evaluate to 0 and pass the 0..4 range check, then trigger an explicit switch to locality 0 (differing from the default -1 'leave at default' behavior). For PCR16 this makes an invalid command line succeed; for DRTM PCRs it produces a TPM locality error instead of rejecting the malformed input up front.

**Fix:** Validate exactly one numeric digit in range before converting, e.g. check locStr[0] in '0'..'4' and locStr[1] == '\0', then locality = locStr[0] - '0'.

examples/pcr/reset.c
```diff
@@ -116,6 +138,12 @@ int TPM2_PCR_Reset_Test(void* userCtx, int argc, char *argv[])
 
 exit:
```
aidangarske, line 144: **🔵 [Low] reset.c exit path calls wolfTPM2_SetLocality on dev when wolfTPM2_Init failed** · _Correctness_

If wolfTPM2_Init fails when -loc=N (N>0) was passed, control jumps to exit: where wolfTPM2_SetLocality(&dev, 0) runs before Cleanup, operating on a device that was not successfully initialized. In practice this is benign because Init zero-initializes dev so oldLocality==0==target returns early without touching the bus (bugs mode verified the stronger 'TIS I/O on garbage context' claim is a false positive and dropped it). This remains a clarity nit: gating the restore on a successful switch would be clearer than relying on that implementation detail.

**Fix:** Only restore the locality when the switch actually succeeded (e.g. track a localitySwitched flag) so the restore never runs on an un-initialized device.

hal/tpm_io_linux.c
```diff
@@ -415,6 +419,94 @@
...
+        /* Acquire the line as an output driven low (assert reset) */
+        XMEMSET(&req, 0, sizeof(req));
+        req.offsets[0] = (unsigned int)WOLFTPM_RESET_LINE;
+        req.num_lines = 1;
+        req.config.flags = GPIO_V2_LINE_FLAG_OUTPUT;
+        req.config.num_attrs = 1;
+        req.config.attrs[0].attr.id = GPIO_V2_LINE_ATTR_ID_OUTPUT_VALUES;
+        req.config.attrs[0].attr.values = 0; /* drive low (assert reset) */
+        req.config.attrs[0].mask = 1;        /* applies to line index 0 */
+        XMEMCPY(req.consumer, "wolfTPM-reset", 14);
```
aidangarske: **🔵 [Low] GPIO consumer label copied with hardcoded length 14** · _Style_

XMEMCPY(req.consumer, "wolfTPM-reset", 14) hardcodes the byte count (string length 13 + NUL). Correct today (buffer is 32 bytes) but if the label is ever edited the 14 becomes stale and could truncate or over/under-copy. Using sizeof of the literal keeps the two in sync automatically.

**Fix:** Replace the magic 14 with sizeof("wolfTPM-reset") (or XSTRLEN+1).

src/tpm2_wrap.c
```diff
@@ -6684,6 +6685,72 @@ int wolfTPM2_ResetPCR(WOLFTPM2_DEV* dev, int pcrIndex)
...
+    if (oldLocality == locality) {
+        return TPM_RC_SUCCESS; /* already active */
+    }
+
+    /* TPM2_TIS_RequestLocalityEx returns the granted locality (0-4) on success,
+     * or a TPM_RC_* code on failure (which may be positive), so success is
+     * exactly rc == locality. Try to acquire the target while holding the
+     * current locality (bounded timeout, so it fails fast when the TPM does not
+     * preempt). */
+    rc = TPM2_TIS_RequestLocalityEx(&dev->ctx, locality,
```
aidangarske, line 6762: **⚪ [Info] wolfTPM2_SetLocality can leave ctx-locality inconsistent if both retry and restore fail** · _Logic_

In the TIS path, on the no-preemption branch the current locality is released and the target re-requested. If that re-request fails, the code attempts to restore the old locality but discards its result via (void). ctx->locality is only updated on a successful grant inside TPM2_TIS_CheckLocalityAccessValid (src/tpm2_tis.c:296). So if the restore ALSO fails, the TPM has no active locality while dev->ctx.locality still holds oldLocality, and a subsequent command would target a locality register the TPM does not consider active. The function does return a non-success rc (a well-behaved caller stops), and this requires a double failure, making it a narrow edge case.

**Fix:** Check the restore's return; if it also fails, set a defined ctx->locality/state or return a distinct error so callers know the TPM interface may need re-init.

aidangarske (commented):
## Skoll Multi-Scan Review

**Findings:** 13 total — 13 posted, 0 skipped
**13 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] [review+review-security] **SWTPM locality switch serializes the wrong byte on big-endian hosts** — `src/tpm2_wrap.c:6693-6703`
- [High] [review] **fwTPM PCR reset locality table diverges from TCG PC Client reference (PCR 16/23 lose loc4; PCR 20-22 wrongly gain loc3)** — `src/fwtpm/fwtpm_command.c:1827-1838`
- [High] [review+review-security] **run_examples.sh -loc skip branch greps for a token the program never prints (unsupported backends fail CI)** — `examples/run_examples.sh:1194-1200`
- [Medium] [review+review-security] **PCR reset usage text contradicts the implemented locality map and supported transports** — `examples/pcr/reset.c:48-49`
- [Medium] [review-security] **Startup locality self-heal can release a live non-default owner** — `src/tpm2_tis.c:337-365`
- [Medium] [bugs] **I2C builds can falsely report nonzero locality as active** — `src/tpm2_tis.c:302-335`
- [Medium] [review] **wolfTPM2_SetLocality TIS preempt/release/restore branch has no automated coverage** — `src/tpm2_wrap.c:6704-6751`
- [Medium] [review] **hal-reset can be reported enabled without exporting the reset API** — `configure.ac:484-491`
- [Low] [review] **TPM_CAP_PCR_PROPERTIES always reports moreData=NO even when truncated by propertyCount** — `src/fwtpm/fwtpm_command.c:1394-1417`
- [Low] [bugs] **Invalid -loc values are accepted as locality 0** — `examples/pcr/reset.c:69-78`
- [Low] [review] **reset.c exit path calls wolfTPM2_SetLocality on dev when wolfTPM2_Init failed** — `examples/pcr/reset.c:139-147`
- [Low] [review] **GPIO consumer label copied with hardcoded length 14** — `hal/tpm_io_linux.c:472`
- [Info] [review-security] **wolfTPM2_SetLocality can leave ctx-locality inconsistent if both retry and restore fail** — `src/tpm2_wrap.c:6729-6748`

hal/tpm_io.c
```diff
@@ -239,6 +239,29 @@ int TPM2_IoCb(TPM2_CTX* ctx, const BYTE* txBuf, BYTE* rxBuf,
...
 
 #endif /* WOLFTPM_ADV_IO */
+
+#ifdef WOLFTPM_HAL_RESET
+/* Pulse the TPM hardware reset (nRST) line to reset the TPM. Dispatches to the
+ * platform implementation. Returns TPM_RC_SUCCESS on success. */
+int TPM2_IoCb_Reset(TPM2_CTX* ctx, void* userCtx)
+{
+    int ret;
+#if defined(__linux__)
```
aidangarske, line 249: **🔴 [High] HAL reset can reference an undefined Linux reset symbol outside the Linux SPI/I2C HAL**

The new reset dispatcher calls TPM2_IoCb_Linux_Reset() whenever __linux__ is defined, but that symbol is only defined in hal/tpm_io_linux.c (declared hal/tpm_io.h:153, defined hal/tpm_io_linux.c:455). The file-selection block in hal/tpm_io.c includes tpm_io_fwtpm.c for WOLFTPM_FWTPM_HAL and tpm_io_mmio.c for WOLFTPM_MMIO via #elif, so tpm_io_linux.c is NOT compiled in those configurations. The configure.ac hal-reset guard (lines 484-500) rejects only --enable-devtpm and --enable-swtpm, so combinations such as --enable-mmio --enable-hal-reset or --enable-fwtpm --disable-swtpm --enable-hal-reset pass the checks, define WOLFTPM_HAL_RESET, compile TPM2_IoCb_Reset (the outer guard only excludes LINUX_DEV/SWTPM/WINAPI), and then link with an undefined TPM2_IoCb_Linux_Reset reference. Verified: the reset dispatcher's __linux__ branch is unconditional and no fallback definition exists for the fwTPM/MMIO HAL sources.

**Fix:** Fix the build matrix hole before merge so --enable-hal-reset cannot produce an undefined symbol.

src/fwtpm/fwtpm_command.c
```diff
@@ -1165,8 +1171,10 @@ static TPM_RC FwCmd_GetCapability(FWTPM_CTX* ctx, TPM2_Packet* cmd,
                 { TPM_ALG_NULL,      0x0000 },
             };
             int numAlgs = (int)(sizeof(algList) / sizeof(algList[0]));
-            if (propertyCount < (UINT32)numAlgs)
+            if (propertyCount < (UINT32)numAlgs) {
```
aidangarske: **🟠 [Medium] fwTPM advertises paged capability data without honoring the paging property**

For TPM_CAP_ALGS and TPM_CAP_COMMANDS the handler back-patches moreData=YES when the list is truncated by propertyCount, but both cases still emit entries starting from index 0 and ignore the request's parsed property start value (parsed at line 1105). A spec-following client that asks for one item, sees moreData, and retries with property = last + 1 receives the first item again and can loop or enumerate duplicates. Verified by contrast with the other capability cases in the same function which DO honor the start value (TPM_CAP_TPM_PROPERTIES uses prop >= property at ~1320, TPM_CAP_PCRS uses tag >= property at ~1400, TPM_CAP_HANDLES uses handle >= property at ~1466) — the ALGS/COMMANDS cases are the outliers. The unit tests only request large counts and do not cover paging.

**Fix:** Fix capability paging semantics before merge.

src/fwtpm/fwtpm_command.c
```diff
@@ -1775,6 +1835,62 @@ static TPM_RC FwCmd_PCR_Read(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
...
 /* --- TPM2_PCR_Extend (CC 0x0182) --- */
+/* Per-PCR locality bitmaps (bit L set = locality L may reset/extend the PCR),
+ * per the TCG PC Client profile. Single source of truth for reset/extend
+ * enforcement and the TPM_CAP_PCR_PROPERTIES report. These match the reference
+ * TPM 2.0 implementation (ibmswtpm2 PlatformPCR.c s_initAttributes) byte-for-
+ * byte and the real ST33 GetCapability dump (loc2 reset of PCR 20-22 was
+ * hardware-confirmed over SPI). Deliberately not just "reject locality 0".
+ *   reset:  0-15 never; 16,23 loc0-3; 17-19 loc4; 20-22 loc2-4.
+ *   extend: 0-16,23 any; 17,18 loc2-4; 19 loc2-3; 20 loc1-3; 21,22 loc2. */
+#define FW_LOC(l)  (1u << (l))
```
aidangarske: **🟠 [Medium] fwTPM PCR reset locality table diverges from the cited ibmswtpm2 reference**

The comment above the tables asserts the values match the reference TPM 2.0 implementation (ibmswtpm2 PlatformPCR.c s_initAttributes) byte-for-byte, while the same comment also cites a real ST33 GetCapability dump (loc2 reset of PCR 20-22 hardware-confirmed over SPI) — two sources that only agree if ibmswtpm2 and ST33 match. The fwPcrExtendLocality table does match ibmswtpm2's extendLocality fields, but the fwPcrResetLocality table appears not to: ibmswtpm2 uses 0x14 (localities 2 and 4 only) for PCR 20-22 whereas this table sets FW_LOC(2)|FW_LOC(3)|FW_LOC(4) (adds locality 3), and ibmswtpm2 uses 0x1F (all localities incl. 4) for PCR 16/23 whereas this table sets FW_LOC(0..3) (drops locality 4). Verified the table values in source. The added unit tests enshrine the divergent values, so the suite will not catch a spec mismatch, and the hardware note only confirms locality 2, not locality 3. For DRTM PCRs, permissiveness (allowing reset from locality 3) is the risky direction.

**Fix:** Double-check the reset-locality values against the cited reference and either correct the table+tests or fix the misleading comment; the extend table already matches, so the reset asymmetry looks unintentional.

examples/pcr/README.md
```diff
@@ -36,6 +36,8 @@ Not all PCRs are equal. The user can perform `extend` operation on all PCRs, but
 * PCR16 is a PCR for debug purposes. This is the PCR used by all tools above by default. It is safe to test and work with PCR16.
 * PCR17-22 are reserved for Dynamic Root of Trust Measurement (DRTM), an advanced topic that is to be covered separately.
 
+Reset locality (TCG PC Client): PCR16/23 reset at locality 0, PCR20-22 at locality 2/3, and PCR17-19 at locality 4. Use `-loc=n` to select the locality (see `wolfTPM2_SetLocality`); it applies to the built-in TIS/SPI driver and the fwTPM.
```
aidangarske: **🔵 [Low] PCR reset locality README does not match the implemented table**

The README says PCR16/23 reset at locality 0 and PCR20-22 at locality 2/3, but the implemented table and new unit tests allow PCR16/23 at localities 0-3 and PCR20-22 at localities 2-4. Verified against src/fwtpm/fwtpm_command.c fwPcrResetLocality (PCR16/23 = loc0-3, PCR20-22 = loc2-4, PCR17-19 = loc4). If the code table is intended, the documentation understates the supported localities.

**Fix:** Keep the docs aligned with the source-of-truth locality table.

wolftpm/tpm2_wrap.h
```diff
@@ -2749,6 +2749,30 @@ WOLFTPM_API int wolfTPM2_ReadPCR(WOLFTPM2_DEV* dev,
...
+        commands run at the given locality (e.g. loc 2/3 to reset PCRs not
+        resettable at loc 0). For the built-in TIS driver it performs the ACCESS
+        handshake (request, then release-and-request if the TPM does not preempt,
+        restoring the old locality on failure); for the SWTPM/mssim transport it
+        records the locality (sent with each command). The Linux kernel driver and
+        Windows TBS own the locality and return NOT_COMPILED_IN. Return to locality
+        0 when done.
+
+    \return TPM_RC_SUCCESS: locality granted and now active
+    \return TPM_RC_TIMEOUT: the TPM did not grant the requested locality
```
aidangarske: **🔵 [Low] wolfTPM2_SetLocality documents only TPM_RC_TIMEOUT for a grant failure**

The Doxygen block lists TPM_RC_TIMEOUT as the failure return when the TPM does not grant the locality (verified at wolftpm/tpm2_wrap.h:2765). On the built-in TIS path the returned rc is whatever TPM2_TIS_RequestLocalityEx propagates, which can be a transport error from TPM2_TIS_Write/TPM2_TIS_CheckLocality (e.g. TPM_RC_FAILURE), not only TPM_RC_TIMEOUT. The documented contract is narrower than the actual behavior, so callers switching only on TPM_RC_TIMEOUT may mishandle a real failure.

**Fix:** Broaden the documented failure return so callers don't switch only on TPM_RC_TIMEOUT.

tests/fwtpm_check.sh
```diff
@@ -382,6 +382,36 @@ else
...
+# --- fwTPM PCR reset locality (end-to-end) ---
+# Both transports carry the command locality: the mssim/socket transport in its
+# SEND_COMMAND framing, and the TIS/SHM transport via the ACCESS-register
+# handshake driven by wolfTPM2_SetLocality. So this exercises the fwTPM's
+# per-PCR locality enforcement end-to-end in either mode.
+# Only run when we started our own fwtpm_server: these asserts encode the fwTPM
+# per-PCR map, which differs from an external server (e.g. ibmswtpm2) that CI may
+# have left listening on the port (which fwtpm_check.sh reuses).
+FWTPM_RESET="$BUILD_DIR/examples/pcr/reset"
+if [ "${STARTED_SERVER:-0}" -eq 1 ] && [ -x "$FWTPM_RESET" ]; then
```
aidangarske, line 394: **🔵 [Low] fwtpm_check.sh locality test is silently skipped without a SKIP tally**

The new end-to-end locality check only runs if [ "${STARTED_SERVER:-0}" -eq 1 ] && [ -x "$FWTPM_RESET" ]. Verified the block has no else branch, unlike the immediately preceding guarded block which does echo "SKIP: unit.test not found"; SKIP=$((SKIP + 1)). When either condition is false the locality block produces no output and does not increment SKIP, so a run that reuses an external server or lacks the reset binary reports as if the locality behavior was covered when it silently was not.

**Fix:** Record a skip so coverage reporting reflects when the locality check did not actually run.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review`**Overall recommendation:** `REQUEST_CHANGES`
**Findings:** 6 total — 6 posted, 0 skipped
**6 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **HAL reset can reference an undefined Linux reset symbol outside the Linux SPI/I2C HAL** — `hal/tpm_io.c:249-250`
- [Medium] **fwTPM advertises paged capability data without honoring the paging property** — `src/fwtpm/fwtpm_command.c:1174-1194`
- [Medium] **fwTPM PCR reset locality table diverges from the cited ibmswtpm2 reference** — `src/fwtpm/fwtpm_command.c:1846-1866`
- [Low] **PCR reset locality README does not match the implemented table** — `examples/pcr/README.md:39`
- [Low] **wolfTPM2_SetLocality documents only TPM_RC_TIMEOUT for a grant failure** — `wolftpm/tpm2_wrap.h:2765`
- [Low] **fwtpm_check.sh locality test is silently skipped without a SKIP tally** — `tests/fwtpm_check.sh:394-413`

src/fwtpm/fwtpm_tis.c
```diff
@@ -133,7 +144,8 @@ static void TisHandleRegAccess(FWTPM_CTX* ctx, FWTPM_TIS_REGS* regs)
 
                     procRc = FWTPM_ProcessCommand(ctx,
                         localCmd, (int)localCmdLen,
-                        regs->rsp_buf, &rspSize, 0 /* locality */);
+                        regs->rsp_buf, &rspSize,
+                        (ctx->tisLocality < 0 ? 0 : ctx->tisLocality));
```
aidangarske: **🔴 [High] fwTPM TIS command execution is not gated to the owning locality**

The PR tracks an active TIS locality in ctx->tisLocality, but only the ACCESS register (offset FWTPM_TIS_ACCESS) checks ownership against the decoded loc (bits 12-15 of reg_addr). Verified in src/fwtpm/fwtpm_tis.c: loc is decoded at line 76 and gates only the ACCESS case (lines 92-111). The STS/COMMAND_READY and GO handling (lines 114-180) and the DATA_FIFO/XDATA_FIFO writes (lines 182+) never compare loc to ctx->tisLocality, and GO dispatches with (ctx->tisLocality < 0 ? 0 : ctx->tisLocality) (line 148) rather than the addressed locality. A register access addressed to a non-owner locality can therefore reset the shared FIFO, write a command, and execute it under the current owner locality, defeating the locality model the PR is introducing.

**Fix:** Gate STS/FIFO handling by the active owner locality and add an end-to-end TIS/SHM test proving a non-owner locality cannot execute under the owner locality.

src/fwtpm/fwtpm_command.c
```diff
@@ -1165,8 +1171,10 @@ static TPM_RC FwCmd_GetCapability(FWTPM_CTX* ctx, TPM2_Packet* cmd,
                 { TPM_ALG_NULL,      0x0000 },
             };
             int numAlgs = (int)(sizeof(algList) / sizeof(algList[0]));
```
aidangarske, line 1194: **🔴 [High] GetCapability advertises pagination without honoring the property cursor**

Verified against src/fwtpm/fwtpm_command.c: the request's property cursor is parsed into `property` (lines 1088, 1105) but is never used. For TPM_CAP_ALGS and TPM_CAP_COMMANDS the code back-patches moreData to YES (rsp->buf[moreDataPos] = 1) when propertyCount truncates the list, yet the emission loops always start at index 0 (lines 1180, 1197). A client that follows moreData will request the next page starting from the last-returned algorithm/command + 1 and receive the same first page again, with moreData still set — an infinite paging loop. This is a regression: before the PR the truncated response did not signal that more data was available.

**Fix:** Fix the ALGS and COMMANDS capability paging before merge.

src/tpm2_wrap.c
```diff
@@ -6684,6 +6685,90 @@ int wolfTPM2_ResetPCR(WOLFTPM2_DEV* dev, int pcrIndex)
...
+        return NOT_COMPILED_IN;
+    }
+#endif
+
+    oldLocality = dev->ctx.locality;
+    if (oldLocality == locality) {
+        return TPM_RC_SUCCESS; /* already active */
+    }
+
+    /* TPM2_TIS_RequestLocalityEx returns the granted locality (0-4) on success,
```
aidangarske, line 6757: **🟠 [Medium] No automated coverage for SetLocality release/restore-on-failure path (built-in TIS)**

Verified in src/tpm2_wrap.c: the most delicate branch — target ungrantable, requiring the old locality to be restored (lines 6752-6758) — leaves ctx->locality pointing at oldLocality while rc holds a non-success target request, and desyncs if the restore also fails. There is no fwTPM/TIS hook to force an ungrantable locality, so this release-and-restore branch ships without regression coverage; the immediate-grant and release-and-retry paths are the only ones exercised in CI.

**Fix:** Add a targeted test (fwTPM TIS or a mockable TIS read/write) that forces an ungrantable target, or explicitly document the branch as hardware-only verified.

hal/tpm_io_linux.c
```diff
@@ -415,6 +420,104 @@
...
+     * such limit and handles the 1s settle and any larger override. */
+    static void TPM2_Reset_DelayUs(unsigned long us)
+    {
+        struct timespec ts;
+        ts.tv_sec  = (time_t)(us / 1000000UL);
+        ts.tv_nsec = (long)((us % 1000000UL) * 1000UL);
+        (void)nanosleep(&ts, NULL);
+    }
+
+    int TPM2_IoCb_Linux_Reset(TPM2_CTX* ctx, void* userCtx)
```
aidangarske, line 459: **🔵 [Low] GPIO nRST reset HAL has no functional test**

TPM2_IoCb_Linux_Reset (open gpiochip, GPIO_V2_GET_LINE_IOCTL, hold, SET_VALUES, settle, close) is only compile-checked in CI, with no GPIO hardware or simulated chip. Error paths (open fail, GET_LINE fail, SET_VALUES fail) and the fd lifecycle are not exercised at runtime. The fd handling and ioctl struct population read correctly on inspection, but there is no smoke test even against gpio-sim.

**Fix:** Where feasible add a gpio-sim-backed smoke test; otherwise document that the reset HAL is verified only on real hardware.

examples/pcr/README.md
```diff
@@ -52,8 +54,9 @@ The TPM 2.0 `TPM2_Quote` API is a standard operation that encapsulates the PCR d
 $ ./examples/pcr/reset -?
```
aidangarske, line 54: **🔵 [Low] Reset example README shows stale -? output line**

Verified against examples/pcr/reset.c: for `-?` the parser matches neither `-loc=` nor a leading digit, so it falls into the else branch and prints "Incorrect arguments" (reset.c line 92), not "PCR index is out of range (0-23)" as the README shows (README line 55). The documented usage block (lines 54-60) also omits the two new help lines usage() now prints: "(PCR 17-19 need locality 4; 20-22 need locality 2-4; enforced by the fwTPM and by discrete TPMs like the ST33)" (reset.c lines 49-50).

**Fix:** Update the README code block to reflect the actual output of the new parser and include the added help lines.

src/tpm2.c
```diff
@@ -739,7 +739,10 @@ TPM_RC TPM2_ChipStartup(TPM2_CTX* ctx, int timeoutTries)
         rc = TPM2_TIS_StartupWait(ctx, timeoutTries);
         if (rc == TPM_RC_SUCCESS) {
 
-            /* Request locality for TPM module */
+            /* Request locality for TPM module. This proactively relinquishes a
```
aidangarske: **🔵 [Low] tpm2.c comment overstates RequestLocality self-heal in default builds**

Verified in src/tpm2_tis.c: the release-any-active-locality recovery in TPM2_TIS_RequestLocality is wrapped in #ifdef WOLFTPM_TIS_RESET_STALE_LOCALITY (lines 340-362), which is OFF by default. In a default build the function simply calls TPM2_TIS_RequestLocalityEx(ctx, WOLFTPM_LOCALITY_DEFAULT, timeout) (line 364) and relinquishes nothing, so the tpm2.c comment claiming it "proactively relinquishes a stale non-default locality" describes behavior most builds do not have.

**Fix:** Qualify the comment to note the self-heal is gated on WOLFTPM_TIS_RESET_STALE_LOCALITY, or drop the claim so it does not mislead readers of default builds.

wolftpm/tpm2_wrap.h
```diff
@@ -2749,6 +2749,30 @@ WOLFTPM_API int wolfTPM2_ReadPCR(WOLFTPM2_DEV* dev,
 */
 WOLFTPM_API int wolfTPM2_ResetPCR(WOLFTPM2_DEV* dev, int pcrIndex);
 
+/*!
```
aidangarske, line 2752: **🔵 [Low] wolfTPM2_SetLocality doxygen return set narrower than actual behavior**

Verified: the doxygen (wolftpm/tpm2_wrap.h lines 2764-2767) lists TPM_RC_TIMEOUT as the failure code, but on the built-in TIS path wolfTPM2_SetLocality returns whatever TPM2_TIS_RequestLocalityEx returns on failure, which can be a raw TIS/TPM_RC_* code (the implementation comment at src/tpm2_wrap.c:6740-6742 explicitly notes the return may be "a TPM_RC_* code on failure (which may be positive)"). Callers testing rc == TPM_RC_TIMEOUT rather than rc != TPM_RC_SUCCESS could mis-handle other error codes.

**Fix:** Broaden the failure documentation to indicate a non-success TPM_RC (not exclusively TPM_RC_TIMEOUT) may be returned, and advise checking rc != TPM_RC_SUCCESS.

examples/pcr/README.md
aidangarske: **🔵 [Low] PCR reset locality README under-documents allowed localities**

Verified: README line 39 says "PCR16/23 reset at locality 0, PCR20-22 at locality 2/3, and PCR17-19 at locality 4," but the reset.c usage text (lines 49-50) already states 20-22 need locality 2-4, and the enforcement helper FwPcrLocalityAllowed/fwPcrResetLocality drives the allowed set. The README wording is inconsistent with the tool's own usage text and the enforcement table (PCR20-22 allow 2-4, not just 2/3).

**Fix:** Update the README wording to match the enforcement table and the reset.c usage text.

src/fwtpm/fwtpm_tis.c
```diff
@@ -396,6 +413,9 @@ int FWTPM_TIS_Init(FWTPM_CTX* ctx)
     /* Auto power-on in TIS mode (no platform port to signal power) */
```
aidangarske, line 448: **🔵 [Low] Dead write to regs-access after ACCESS read path reworked**

Verified in src/fwtpm/fwtpm_tis.c: regs->access is assigned only once, at init (line 403). A search of the file shows no other reference to ->access — the ACCESS read handler derives its value from ctx->tisLocality and the ACCESS write handler no longer updates regs->access. The field is therefore dead state in the shared-memory register struct. Not a bug (the field remains part of the shared ABI), but the leftover initialization can mislead maintainers into thinking regs->access still drives ACCESS responses.

**Fix:** Clarify or remove the now-unused regs->access write to avoid future confusion about the source of truth for locality state.

---
> **Note:** Referenced line (`src/fwtpm/fwtpm_tis.c:403`) is outside the diff hunk. Comment anchored to nearest changed region.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review`**Overall recommendation:** `REQUEST_CHANGES`
**Findings:** 9 total — 9 posted, 0 skipped
**9 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **fwTPM TIS command execution is not gated to the owning locality** — `src/fwtpm/fwtpm_tis.c:114-148`
- [High] **GetCapability advertises pagination without honoring the property cursor** — `src/fwtpm/fwtpm_command.c:1173-1194`
- [Medium] **No automated coverage for SetLocality release/restore-on-failure path (built-in TIS)** — `src/tpm2_wrap.c:6740-6768`
- [Low] **GPIO nRST reset HAL has no functional test** — `hal/tpm_io_linux.c:455-518`
- [Low] **Dead write to regs-access after ACCESS read path reworked** — `src/fwtpm/fwtpm_tis.c:403`
- [Low] **Reset example README shows stale -? output line** — `examples/pcr/README.md:54-60`
- [Low] **tpm2.c comment overstates RequestLocality self-heal in default builds** — `src/tpm2.c:742-746`
- [Low] **wolfTPM2_SetLocality doxygen return set narrower than actual behavior** — `wolftpm/tpm2_wrap.h:2752-2774`
- [Low] **PCR reset locality README under-documents allowed localities** — `examples/pcr/README.md:39`

aidangarske (dismissed):

src/fwtpm/fwtpm_command.c
```diff
@@ -1799,6 +1975,15 @@ static TPM_RC FwCmd_PCR_Extend(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
         }
     }
 
+    /* Enforce the per-PCR extend locality (e.g. DRTM PCRs 17-22 cannot be
```
tmael, line 2001: This adds extend-locality enforcement that didn't exist before — DRTM PCRs 17–22 can no longer be extended from locality 0 (returns TPM_RC_LOCALITY). Correct per PC Client, but it's a behavior change for existing fwTPM callers; worth a note in the changelog/PR description.

src/fwtpm/fwtpm_command.c
```diff
@@ -1775,6 +1895,62 @@ static TPM_RC FwCmd_PCR_Read(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
...
+
+static const byte fwPcrExtendLocality[IMPLEMENTATION_PCR] = {
+    /*  0- 7 */ FW_LOC_ALL, FW_LOC_ALL, FW_LOC_ALL, FW_LOC_ALL,
+                FW_LOC_ALL, FW_LOC_ALL, FW_LOC_ALL, FW_LOC_ALL,
+    /*  8-15 */ FW_LOC_ALL, FW_LOC_ALL, FW_LOC_ALL, FW_LOC_ALL,
+                FW_LOC_ALL, FW_LOC_ALL, FW_LOC_ALL, FW_LOC_ALL,
+    /* 16 */ FW_LOC_ALL,
+    /* 17 */ FW_LOC(2) | FW_LOC(3) | FW_LOC(4),
+    /* 18 */ FW_LOC(2) | FW_LOC(3) | FW_LOC(4),
+    /* 19 */ FW_LOC(2) | FW_LOC(3),
```
tmael: The extend map here is asymmetric with the reset map (e.g. PCR 20 extend allows loc 1 but reset does not; PCR 19 extend is loc 2–3 while reset is loc 4; 21/22 extend loc 2 only). These are the least intuitive values — worth a byte-for-byte confirmation against the cited s_initAttributes, since the whole enforcement + capability report derives from this table.

tmael: Dropped my comment. Verified the reset/extend tables against ibmswtpm2 s_initAttributes — byte-for-byte match, incl. the 16/23 loc-4 exclusion.

aidangarske (approved):

PR #547, opened 2026-07-07
Harden ECDH buffer bounds, RSA/ECC key conversion, identity auth, and fwTPM PQC TestParms

aidangarske:
```
F-6738, F-6739, F-6740, F-6741, F-6742, F-6743, F-6744, F-6745, F-6746, F-6747, F-6748, F-6749, F-6750
```

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -6329,8 +6329,7 @@ int wolfTPM2_ECDHGenKey(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* ecdhKey, int curve_id,
     return rc;
 }
 
-/* Copy an ECDH shared-secret x-coordinate from a TPM response into the caller
- * buffer, rejecting a response larger than the caller's capacity (*outSz) */
+/* Copy the ECDH shared secret to the caller buffer, reject if larger */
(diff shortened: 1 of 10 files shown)


src/tpm2_wrap.c
```diff
@@ -10834,6 +10854,19 @@ int wolfTPM2_SetIdentityAuth(WOLFTPM2_DEV* dev, WOLFTPM2_HANDLE* handle,
     enum wc_HashType hashType = WC_HASH_TYPE_SHA256;
     uint8_t digest[TPM_SHA256_DIGEST_SIZE];
 
+    if (handle == NULL) {
+        return BAD_FUNC_ARG;
+    }
+
+#if !defined(WOLFTPM_SLB9672) && !defined(WOLFTPM_SLB9673) && \
```
dgarske: Technically this masterPassword concept is spec driven, but currently only supported with the STSAFE-TPM, so I think its odd that this comment says anything about Infineon. And the macros should be WOLFTPM_ST33 || WOLFTPM_AUTODETECT driven if we are trapping on defaults. To be honest a lot of these changes around this don't need to be made, but please see if you can improve it.

dgarske (changes requested):

dgarske (approved):
Merging with unrelated failures fixed in https://github.com/wolfSSL/wolfTPM/pull/546

PR #549, opened 2026-07-10
tpm2_types.h: freestanding build support (WOLFTPM_NO_STD_HEADERS) and NO_RETRY/MAX_RETRIES config cleanup

tmael:
Two small, independent build-configuration cleanups in `wolftpm/tpm2_types.h`. No functional change for existing hosted builds.

### 1. Support freestanding builds — `WOLFTPM_NO_STD_HEADERS`

A target without a hosted C standard library currently **cannot build wolfTPM**: `tpm2_types.h` unconditionally includes `<stdint.h>`, and in the standalone `WOLFTPM2_NO_WOLFCRYPT` path also `<stdio.h>` / `<stdlib.h>` / `<string.h>`. A `-nostdinc` / freestanding compile fails at the first include.

Guard the standard type/libc headers behind `WOLFTPM_NO_STD_HEADERS`. When defined, the target supplies the fixed-width integer types and the mem/str functions (used via the `X*` wrappers) through `user_settings.h` instead. Platform delay/sleep headers (`unistd.h`/`time.h`) are unchanged — they remain steered by the existing `XTPM_WAIT` / `XSLEEP_MS` hooks and their platform conditionals.

### 2. Error on conflicting `WOLFTPM_NO_RETRY` + `WOLFTPM_MAX_RETRIES`

`-DWOLFTPM_NO_RETRY -DWOLFTPM_MAX_RETRIES=N` (N > 0) is contradictory: compile the retry mechanism out *and* retry N times. The old block silently `#undef`d the user's `MAX_RETRIES` and forced 0, so the build ignored an explicit flag with no diagnostic. `#undef`-ing a caller-supplied command-line macro is also fragile.

Define the default first, then `#error` on the actual conflict. Ordering the default before the check keeps it `-Wundef`-clean, including the `NO_RETRY`-without-`MAX_RETRIES` case.

### Testing
- Default `make` (with `--enable-swtpm`), including unit tests: builds clean, no regression.
- Freestanding: `wolftpm/tpm2_types.h` compiles under `-nostdinc` with `-DWOLFTPM_NO_STD_HEADERS -DWOLFTPM2_NO_WOLFCRYPT` and types supplied via `user_settings.h` — no standard header reached.
- Retry macros exercised across all permutations under `-Wundef -Werror`; the contradictory combo now fails with a clear `#error`.

### DIFF
--- a/.github/workflows/freestanding-build.yml
+++ b/.github/workflows/freestanding-build.yml
@@ -1,0 +1,63 @@
+name: Freestanding Build (WOLFTPM_NO_STD_HEADERS)
+
+# Builds the standalone (WOLFTPM2_NO_WOLFCRYPT) library with no hosted C library
+# (-nostdinc) to prove WOLFTPM_NO_STD_HEADERS keeps the standard headers out.
+# The fixed-width types and mem/str prototypes are supplied via user_settings.h,
+# exactly as a bare-metal integrator would. -nostdinc is applied only to the
+# build (make CFLAGS), not to configure, whose feature probes need libc.
+
+on:
+  push:
+    branches: [ 'master', 'main', 'release/**' ]
+  pull_request:
+    branches: [ '*' ]
+    types: [opened, synchronize, reopened, ready_for_review]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  freestanding:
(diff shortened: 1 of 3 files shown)


dgarske (changes requested):
Thank you! I'd like a new GitHub CI build for this `WOLFTPM_NO_STD_HEADERS`. Can you add it @tmael ?

dgarske (approved):

PR #551, opened 2026-07-13
Harden crypto callback, ASN.1 parsing, param-enc and marshalling; add negative tests

aidangarske:
```
F-6839, F-6840, F-6841, F-6842, F-6843, F-6844, F-6845, F-6846, F-6847, F-6848, F-6849, F-6850
```


src/tpm2_packet.c
embhorn, line 1: For all these size checks, would it be better to throw an error instead of truncating the stored value? Not sure if it really matters here.

aidangarske: normally I think we just truncate without an error. AppendSensitive already does this I think the F-6844 was just a constancy thing

embhorn (approved):

PR #553, opened 2026-07-16
Fix WOLFTPM_IS_COMMAND_UNAVAILABLE to handle vendor error code variants

AmumuI:
## Summary
Some TPM chips (e.g. Nations Technologies NS350) return TPM_RC_COMMAND_CODE
with additional vendor bits set (0xb0143 instead of the standard 0x143).
The `WOLFTPM_IS_COMMAND_UNAVAILABLE` macro previously used exact equality,
causing it to fail to recognize these vendor-extended error codes. This
resulted in the bench program treating the error as fatal and aborting
before reaching SHA/RSA/ECC benchmarks.
图片
## Fix
Apply a `0x1FF` mask to compare only the base VER1 error code bits,
stripping any vendor-specific high bits. This matches the approach already
used elsewhere in the codebase (e.g. `(rc & TPM_RC_MODE) == TPM_RC_MODE`).
## Verified
Tested on Nations Technologies NS350 TPM 2.0 module. After this fix, the
bench program correctly skips symmetric encryption tests and continues
to produce SHA, RSA, and ECC benchmark results.

图片

### DIFF
--- a/wolftpm/tpm2.h
+++ b/wolftpm/tpm2.h
@@ -2151,9 +2151,9 @@ struct wolfTPM_winContext {
 #define TPM_E_COMMAND_BLOCKED (0x80280400)
 #endif
 
-#define WOLFTPM_IS_COMMAND_UNAVAILABLE(code) ((code) == (int)TPM_RC_COMMAND_CODE || (code) == (int)TPM_E_COMMAND_BLOCKED)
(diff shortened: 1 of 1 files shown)


dgarske:
Hi @AmumuI ,

Thank you for this report! Can you tell us more about your use of wolfTPM and your project? If you'd like to keep it private you can email support at wolfssl dot com and mention this PR. Are you planning any additional pull requests? If not we may opt to treat this as a bug report and fix ourself since setting up the contributor agreement is not trivial.

Thanks, David Garske, wolfSSL

PR #557, opened 2026-07-21
GPLv2 license exception for Cisco Systems, Inc. use of wolfTPM in U-Boot only

dgarske:
Pairs with https://github.com/wolfSSL/wolfssl/pull/10976


toddouska (approved):
Thanks, looks good.

PR #560, opened 2026-07-24
Add caller-supplied policy authorization for firmware upgrade

dgarske:
## Summary

Lets callers authorize the firmware-upgrade start command with their own platform-hierarchy policy (multi-branch PolicyOR, SHA-256/384/512) instead of the fixed library authorization. Backward compatible - no existing API changes.

## Features
- New `_ex` APIs taking a caller-satisfied session: `wolfTPM2_FirmwareUpgradeHash_ex`, `wolfTPM2_FirmwareUpgrade_ex`, `wolfTPM2_FirmwareUpgradeRecover_ex`; the existing calls forward with a NULL session.
- New public wrappers: `wolfTPM2_PolicyOR`, `wolfTPM2_SetPrimaryPolicy`, `wolfTPM2_PolicyCommandCodeMake`, `wolfTPM2_IsAlgSupported`.
- Infineon: a caller session leaves the platform primary policy untouched; ST33: it replaces the default `TPM_RS_PW`.
- Firmware examples: non-destructive `--policytest` self-test and `--policy`/`--policyor` `[--sha256|--sha384|--sha512]` end-to-end modes; shared helpers factored into `examples/firmware/firmware_policy.c`.

## Fixes
- `wolfTPM2_PolicyOR` validates branch count and each branch digest size against its buffer (out-of-bounds read, CWE-125).
- Examples skip a policy hash the TPM does not implement (via `TPM_CAP_ALGS`) instead of failing with `TPM_RC_SIZE`.
- Example arg parsing rejects unknown/extra options; the caller session is released only when not consumed by a started upgrade; a provisioned platform policy is cleared on the failure path.
- CMake builds the new shared example source.

### DIFF
--- a/examples/firmware/README.md
+++ b/examples/firmware/README.md
@@ -54,7 +54,13 @@ The TPM has a vendor capability for getting the key group id. This is populated 
 Infineon Firmware Update Usage:
         ./ifx_fw_update (get info)
         ./ifx_fw_update --abandon (cancel)
-        ./ifx_fw_update <manifest_file> <firmware_file>
+        ./ifx_fw_update --policytest (safe policy auth self-test)
+        ./ifx_fw_update [policy opts] <manifest_file> <firmware_file>
+        ./ifx_fw_update <manifest_file> <firmware_file> (default auth)
+Policy options (caller-supplied authorization):
+        --policy    provision+satisfy a PolicyCommandCode
+        --policyor  provision+satisfy a PolicyOR (multi-branch)
+        --sha256|--sha384|--sha512  policy hash (default SHA-256)
 
 # Run without arguments to display the current firmware information including key group id and operational mode
 ./ifx_fw_update
@@ -131,7 +137,13 @@ The `st33_fw_update` tool automatically detects the firmware format.
 ST33 Firmware Update Usage:
 	./st33_fw_update (get info)
 	./st33_fw_update --abandon (cancel)
-	./st33_fw_update <firmware.fi>
+	./st33_fw_update --policytest (safe policy auth self-test)
+	./st33_fw_update [policy opts] <firmware.fi>
+	./st33_fw_update <firmware.fi> (default password auth)
+Policy options (caller-supplied authorization):
+	--policy    provision+satisfy a PolicyCommandCode
+	--policyor  provision+satisfy a PolicyOR (multi-branch)
+	--sha256|--sha384|--sha512  policy hash (default SHA-256)
 
 Firmware format is auto-detected from TPM firmware version:
       - Firmware < 512: Non-LMS format (177 byte manifest)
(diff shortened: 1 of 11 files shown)


aidangarske (commented):
## Skoll Multi-Scan Review

**5 finding(s) not tied to a diff line** (full detail below)

#### Failure cleanup cannot clear the command-bound platform policy it installed
**File:** `examples/firmware/firmware_policy.c:174-288`
**Function:** `firmware_policy_clear / firmware_policy_session_setup`
**Modes:** review+review-security
**Severity:** High
**Category:** Logic
**Confidence:** High

After wolfTPM2_SetPrimaryPolicy() installs an authPolicy whose only satisfied branch is PolicyCommandCode(FieldUpgradeStart) (or the PolicyOR containing only FieldUpgradeStart and NV_Read), the failure path calls firmware_policy_clear() using the device's default password authorization. TPM2_SetPrimaryPolicy is itself authorized by the hierarchy's current authorization policy, and neither installed branch permits TPM_CC_SetPrimaryPolicy; the clear command therefore fails. A post-provisioning failure (e.g. StartSession_ex or PolicyOR failing) leaves the platform hierarchy locked behind the policy until a TPM reset, while the example and README claim it restores default authorization. The same cleanup is retried after failed updates in both firmware examples. Severity noted as High by the review mode and Medium by the security mode; the stricter High is kept.

**Recommendation:** Make rollback independently authorizable: include a PolicyCommandCode(TPM_CC_SetPrimaryPolicy) branch in the provisioned PolicyOR, create and activate a session satisfying that branch before clearing, and propagate/report a rollback failure. Otherwise do not claim automatic rollback and require a TPM reset after any post-provisioning failure. Add a hardware-backed or transport-mocked regression test that forces a failure after provisioning and verifies default platform authorization is usable afterward.

_Referenced code: `examples/firmware/firmware_policy.c:174-288` (6 lines)_

---

#### Public digest helper writes without a knowable output capacity
**File:** `src/tpm2_wrap.c:10973-10995`
**Function:** `wolfTPM2_PolicyCommandCodeMake`
**Modes:** review
**Severity:** High
**Category:** bug
**Confidence:** High

The new public API accepts only a byte* output and an output-only digestSz, then unconditionally zeroes and hashes hashSz bytes. A caller with a smaller buffer (e.g. a one-byte buffer with TPM_ALG_SHA512) is overwritten by both XMEMSET and wc_HashFinal. The documented requirement that the buffer be large enough cannot be checked by this API because no capacity is supplied and *digestSz is overwritten before hashing.

**Recommendation:** Add an explicit output-capacity argument, or define *digestSz as an input/output capacity and return BUFFER_E when it is smaller than hashSz; update the Doxygen, examples, and tests accordingly. Add a short-buffer regression test under ASan or an equivalent bounds-checking configuration.

_Referenced code: `src/tpm2_wrap.c:10973-10995` (5 lines)_

---

#### Caller-supplied policies requiring an auth value are serialized as empty authorization
**File:** `src/tpm2_wrap.c:11277-11316, src/tpm2_wrap.c:11709-11738`
**Function:** `tpm2_ifx_firmware_start / tpm2_st33_firmware_start_common`
**Modes:** review-security
**Severity:** Medium
**Category:** Logic
**Confidence:** High

Before this PR, the vendor start commands were used only with the library-created policy session, which requires no policy password or auth value. The new _ex path accepts an arbitrary caller-satisfied WOLFTPM2_SESSION but forwards only startSession->handle.hndl to the raw vendor command. Those commands construct an all-zero TPMS_AUTH_COMMAND; they do not serialize startSession->handle.auth, policyPass, policyAuth, nonces, or a command HMAC. This bypasses the normal wrapper authorization path that computes the session HMAC. A platform policy containing PolicyAuthValue or PolicyPassword therefore fails at FieldUpgradeStart, although the new README explicitly presents PolicyAuthValue as a supported policy branch.

**Recommendation:** Route caller sessions through the normal authorization machinery so nonce, attributes, password/auth-value handling, and HMAC are serialized, or explicitly reject policy-password/auth-value sessions and narrow the API documentation. Add tests for PolicyAuthValue and PolicyPassword sessions in both vendor start paths.

_Referenced code: `src/tpm2_wrap.c:11277-11316, src/tpm2_wrap.c:11709-11738` (11 lines)_

---

#### PolicyOR wrapper accepts a one-branch list that TPM2_PolicyOR rejects
**File:** `src/tpm2_wrap.c:10751-10764`
**Function:** `wolfTPM2_PolicyOR`
**Modes:** review-security
**Severity:** Low
**Category:** Logic
**Confidence:** High

The new validation treats any nonzero pHashList->count as valid and the public header documents a range beginning at one. TPM2_PolicyOR requires at least two digests; the repository's TPM implementation enforces digestCount < 2 as TPM_RC_VALUE. A caller passing one digest therefore receives a TPM error instead of the documented wrapper-side BAD_FUNC_ARG, making the new validation contract inaccurate.

**Recommendation:** Require pHashList->count >= 2, update the API documentation, and add a unit test for a one-branch list.

_Referenced code: `src/tpm2_wrap.c:10751-10764` (4 lines)_

---

#### No test exercises the non-NULL caller-session firmware path
**File:** `tests/unit_tests.c:324-350`
**Function:** `test_wolfTPM2_ST33_FirmwareUpgrade`
**Modes:** review-security
**Severity:** Low
**Category:** Missing Tests
**Confidence:** High

The added _ex tests pass a NULL device and NULL startSession, so they only verify early argument rejection and legacy forwarding. They never execute the changed branches that consume a caller session in either vendor start implementation. The standalone policy self-test covers PolicyOR digest construction but does not call a firmware _ex API. This gap leaves the new wire authorization format and policy cleanup behavior unverified.

**Recommendation:** Add a mocked TPM transport test that invokes each _ex path with a populated policy session and asserts the outgoing authorization area, including policy-auth-value/password behavior and cleanup after a rejected start.

_Referenced code: `tests/unit_tests.c:324-350` (4 lines)_

---

aidangarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review] **wolfTPM2_IsAlgSupported tri-state return (1/0/rc) is easy to misuse** — `src/tpm2_wrap.c:1030-1056`
- [Low] [review-security] **Policy self-test hard-fails when TPM implements a hash wolfCrypt was not built with** — `examples/firmware/firmware_policy.c:153-173`
- [Low] [review] **Cleanup comment/guard says library 'zeroes' the session handle, but it sets TPM_RH_NULL** — `examples/firmware/ifx_fw_update.c:311-314; examples/firmware/st33_fw_update.c:472-475`
- [Low] [review] **No unit coverage for wolfTPM2_PolicyOR / wolfTPM2_IsAlgSupported success paths** — `tests/unit_tests.c:8200-8209`
- [Info] [review-security] **README describes consumed session as 'zeroed' but library sets TPM_RH_NULL** — `examples/firmware/README.md:270-276`

src/tpm2_wrap.c
```diff
@@ -1024,6 +1024,37 @@ int wolfTPM2_GetCapabilities(WOLFTPM2_DEV* dev, WOLFTPM2_CAPS* cap)
     return wolfTPM2_GetCapabilities_NoDev(cap);
 }
 
+/* Return 1 if the TPM implements the given algorithm, 0 otherwise.
+ * Queries TPM_CAP_ALGS: the TPM returns algorithms with ID >= property, so a
+ * match at index 0 for a single-property query means it is implemented. */
+int wolfTPM2_IsAlgSupported(WOLFTPM2_DEV* dev, TPM_ALG_ID alg)
```
aidangarske: **🟠 [Medium] wolfTPM2_IsAlgSupported tri-state return (1/0/rc) is easy to misuse** · _api_

The new public API returns 1 (supported), 0 (not supported), or a raw non-zero `rc` (query failure / BAD_FUNC_ARG). A natural boolean use — `if (wolfTPM2_IsAlgSupported(dev, alg)) { use it }` — would treat a capability-query error (non-zero rc) as "supported", which is the opposite of safe-fail. The two in-tree callers handle it correctly (checking `== 0` and `!= 1` separately) and the doxygen documents the contract, but the shape is unusual for wolfTPM where wrappers normally return TPM_RC and use an out-parameter for booleans. This raises misuse risk for external callers.

**Fix:** Consider an out-parameter form (e.g. `int wolfTPM2_IsAlgSupported(WOLFTPM2_DEV* dev, TPM_ALG_ID alg, int* isSupported)` returning TPM_RC) to match wolfTPM conventions, or at minimum keep the prominent doxygen warning. Not blocking.

examples/firmware/firmware_policy.c
```diff
@@ -0,0 +1,388 @@
...
+    else {
+        printf("  %s PolicyOR: ERROR 0x%x: %s\n",
+            name, rc, TPM2_GetRCString(rc));
+    }
+
+    wolfTPM2_UnloadHandle(dev, &sess.handle);
+    return rc;
+}
+
+int firmware_policy_selftest_all(WOLFTPM2_DEV* dev)
```
aidangarske, line 193: **🔵 [Low] Policy self-test hard-fails when TPM implements a hash wolfCrypt was not built with** · _Logic_

The new `--policytest` self-test iterates SHA2-256/384/512 unconditionally and gates each only on `wolfTPM2_IsAlgSupported()` (a TPM-side `TPM_CAP_ALGS` query). It does not verify that the *local* wolfCrypt build supports the algorithm (`WOLFSSL_SHA384`/`WOLFSSL_SHA512`). If a TPM implements e.g. SHA2-512 but wolfCrypt was compiled without `WOLFSSL_SHA512`, `wolfTPM2_IsAlgSupported` returns 1 (supported), the code proceeds into `wolfTPM2_PolicyCommandCodeMake` → `wolfTPM2_PolicyHash` → `wc_HashInit(SHA512)` which returns a negative/`NOT_COMPILED_IN` error. That negative rc is treated as a hard failure (`rc != 0 && rc != 1`), so `firmware_policy_selftest_all` returns -1 and `--policytest` reports failure for what is only a build-configuration mismatch rather than skipping the branch. `TPM2_GetHashDigestSize()` is a static table lookup and does not catch this because it returns 64 regardless of build flags. This is example-only robustness (no security impact), but the self-test is advertised as a safe, non-destructive check.

**Fix:** Guard the SHA-384/512 self-test entries with `#ifdef WOLFSSL_SHA384` / `#ifdef WOLFSSL_SHA512`, or treat a local `wc_HashInit`/`TPM2_GetHashType` failure as a skip (like the unsupported-algorithm case) rather than a hard failure, so a build/TPM capability mismatch does not make the non-destructive self-test report failure.

examples/firmware/ifx_fw_update.c
```diff
@@ -212,6 +307,23 @@ int TPM2_IFX_Firmware_Update(void* userCtx, int argc, char *argv[])
             rc, TPM2_GetRCString(rc));
     }
 
+#ifdef HAVE_FW_POLICY
+    /* The library zeroes the session handle when the TPM consumes it on a
```
aidangarske: **🔵 [Low] Cleanup comment/guard says library 'zeroes' the session handle, but it sets TPM_RH_NULL** · _style_

The cleanup comment states "The library zeroes the session handle when the TPM consumes it on a successful start; a non-zero handle means it was not consumed", and the code guards cleanup with `if (policySession.handle.hndl != 0)`. On a successful start the library actually sets `startSession->handle.hndl = TPM_RH_NULL` (0x40000007 — see `tpm2_ifx_firmware_start` / `tpm2_st33_firmware_start_common`), which is non-zero. So the `!= 0` guard does NOT distinguish a consumed session from a live one — it is true in both cases, and the call is only safe because `wolfTPM2_UnloadHandle` internally treats `TPM_RH_NULL` as a no-op. The comment is factually wrong and the guard does not do what it claims. Functionally harmless today, but misleading and fragile. Note the header doxygen (`wolftpm/tpm2_wrap.h:5245`) states the TPM_RH_NULL behavior correctly, so the example comments contradict the API docs.

**Fix:** Fix the comment to reflect that the handle is set to TPM_RH_NULL (not zeroed) and that UnloadHandle is a no-op on it, in both ifx_fw_update.c and st33_fw_update.c.

**Suggestion:**
```suggestion
/* On a successful start the TPM consumes the session and the library sets
     * handle.hndl to TPM_RH_NULL; wolfTPM2_UnloadHandle is a no-op on that, so
     * this only flushes a still-loaded session. */
    if (policySession.handle.hndl != 0) {
        wolfTPM2_UnloadHandle(&dev, &policySession.handle);
    }
```

tests/unit_tests.c
```diff
@@ -7807,6 +8197,16 @@ int unit_tests(int argc, char *argv[])
     test_wolfTPM2_ST33_FirmwareUpgrade();
     #endif
     #endif
+    test_wolfTPM2_PolicyOR();
```
aidangarske, line 8457: **🔵 [Low] No unit coverage for wolfTPM2_PolicyOR / wolfTPM2_IsAlgSupported success paths** · _test_

Argument validation for the new wrappers is well covered, and `wolfTPM2_PolicyCommandCodeMake` has strong known-answer vectors. However the success paths of `wolfTPM2_PolicyOR` (a real 2-branch OR satisfied on a session) and `wolfTPM2_IsAlgSupported` (a positive/negative capability result) are only exercised by the hardware/SWTPM example self-test (`firmware_policy_selftest`), not by the unit suite. The SWTPM-gated `test_wolfTPM2_SetPrimaryPolicy_rollback` and `test_wolfTPM2_FirmwareUpgrade_ex_session` show the pattern for adding a simulator-only positive test.

**Fix:** Optionally add SWTPM-gated positive tests: query a known-supported alg (e.g. SHA-256 → 1) and an implausible one for IsAlgSupported, and a real 2-branch PolicyOR satisfied on a started session verified via GetPolicyDigest. Acceptable to defer given the example self-test covers it on hardware.

examples/firmware/README.md
```diff
@@ -199,3 +211,70 @@ Success: Please reset or power cycle TPM
...
+    manifest, manifest_sz, fwDataCb, fwCbCtx, &session);
+
+cleanup:
+/* On a successful FieldUpgradeStart the library zeroes session.handle.hndl (the
+ * TPM consumed the session), so this only releases a still-loaded session. */
+if (session.handle.hndl != 0)
+    wolfTPM2_UnloadHandle(&dev, &session.handle);
+```
+
+Passing `NULL` for the final `startSession` argument makes `wolfTPM2_FirmwareUpgradeHash_ex()` behave exactly like `wolfTPM2_FirmwareUpgradeHash()` (library-managed authorization), so existing code is unaffected.
```
aidangarske, line 273: **⚪ [Info] README describes consumed session as 'zeroed' but library sets TPM_RH_NULL** · _Logic_

On a successful FieldUpgradeStart the library marks the caller session consumed with `startSession->handle.hndl = TPM_RH_NULL` (0x40000007), not 0. The new README guidance states "the library zeroes session.handle.hndl" and shows `if (session.handle.hndl != 0) wolfTPM2_UnloadHandle(...)`. Functionally this is safe: `wolfTPM2_UnloadHandle` treats `TPM_RH_NULL` as a no-op (src/tpm2_wrap.c:6962), so both the example's `!= 0` guard and a naive user's `== 0` consumption check remain safe against double-flush/leak. The only issue is the documentation wording: a library user who literally trusts "zeroes" and tests `hndl == 0` to detect consumption will misclassify a consumed session (it will read as 0x40000007), though the subsequent UnloadHandle still no-ops. No security or resource-leak impact; documentation accuracy only.

**Fix:** Update the README to say the library sets `session.handle.hndl` to `TPM_RH_NULL` on consumption (which `wolfTPM2_UnloadHandle` safely no-ops), rather than "zeroes", so users do not rely on an `== 0` consumption check.

examples/firmware/ifx_fw_update.c
aidangarske: Testing new feature with "apply suggestion" on skoll

aidangarske (approved):
Looks good!

PR #561, opened 2026-07-27
Fix fwTPM response buffer overflow, SPDM clear-frame command bypass, and other misc hardening

aidangarske:
```
F-7026, F-7260, F-7261, F-7263, F-7264, F-7265, F-7266, F-7267, F-7268,
F-7269, F-7270, F-7271, F-7272, F-7273, F-7274, F-7275, F-7276, F-7277, 
F-7279, F-7280, F-7281, F-7282, F-7283, F-7284, F-7285
```

### DIFF
--- a/src/fwtpm/fwtpm_command.c
+++ b/src/fwtpm/fwtpm_command.c
@@ -1686,6 +1686,181 @@ static TPM_RC FwCmd_GetCapability(FWTPM_CTX* ctx, TPM2_Packet* cmd,
 /* --- TPM2_TestParms (CC 0x018A) --- */
 /* Validates that the given algorithm parameters are supported.
  * No auth, no output params. */
+/* Validate a hash selector carried inside a scheme's details union. */
+static TPM_RC FwTestHashAlg(UINT16 hashAlg)
+{
+    TPM_RC rc = TPM_RC_SUCCESS;
+
+    if (TPM2_GetHashDigestSize(hashAlg) <= 0) {
+        rc = TPM_RC_HASH;
+    }
+    return rc;
+}
+
+/* Validate a TPMT_SYM_DEF(_OBJECT). TPM_ALG_NULL means no symmetric alg. */
+static TPM_RC FwTestSymDef(const TPMT_SYM_DEF* sym)
+{
+    TPM_RC rc = TPM_RC_SUCCESS;
+
+    if (sym->algorithm == TPM_ALG_NULL) {
+        rc = TPM_RC_SUCCESS;
+    }
+#ifndef NO_AES
+    else if (sym->algorithm == TPM_ALG_AES) {
+        UINT16 mode = sym->mode.sym;
+        if (sym->keyBits.aes != 128 && sym->keyBits.aes != 192 &&
+                sym->keyBits.aes != 256) {
+            rc = TPM_RC_KEY_SIZE;
+        }
+        else if (mode != TPM_ALG_CFB && mode != TPM_ALG_CBC &&
+                mode != TPM_ALG_CTR && mode != TPM_ALG_OFB &&
+                mode != TPM_ALG_ECB) {
+            rc = TPM_RC_MODE;
+        }
+    }
+#endif
+    else {
+        rc = TPM_RC_SYMMETRIC;
+    }
+    return rc;
+}
+
+#ifndef NO_RSA
+/* Validate a TPMT_RSA_SCHEME selector and its hash, when it carries one. */
+static TPM_RC FwTestRsaScheme(const TPMT_RSA_SCHEME* scheme)
+{
+    TPM_RC rc = TPM_RC_SUCCESS;
+    UINT16 alg = scheme->scheme;
+
+    if (alg == TPM_ALG_NULL || alg == TPM_ALG_RSAES) {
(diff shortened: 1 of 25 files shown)


src/fwtpm/fwtpm_command.c
```diff
@@ -16221,13 +16410,24 @@ int FWTPM_ProcessCommand(FWTPM_CTX* ctx,
 #endif
     int pj, hj;                     /* Loop indices for auth validation */
     int authFail;                   /* Password comparison result */
+    int rspCap;                     /* Caller's response buffer capacity */
+    int rspTruncated = 0;           /* Response did not fit the buffer */
+    TPM2B_NONCE savedNonce[FWTPM_MAX_CMD_AUTHS];
```
dgarske: This is a large stack addition. Is this save/restore really needed? If so can we isolate this into the block of code below it is used and double check the size of this... does it really need to be this large?

src/fwtpm/fwtpm_command.c
```diff
@@ -16221,13 +16410,24 @@ int FWTPM_ProcessCommand(FWTPM_CTX* ctx,
...
 
     if (ctx == NULL || cmdBuf == NULL || rspBuf == NULL || rspSize == NULL) {
         return BAD_FUNC_ARG;
     }
 
+    /* rspSize is in/out: capacity in, bytes written out. Callers that leave
```
dgarske, line 16420: Condense this comment section and the one after it.

src/fwtpm/fwtpm_main.c
```diff
@@ -81,19 +81,48 @@ static void usage(const char* progname)
 }
 
 #ifdef WOLFTPM_SPDM_RESPONDER
+/* Minimal TPM_ST_NO_SESSIONS error frame carrying TPM_RC_SIZE. */
+static int FwtpmBuildSizeError(byte* resp)
+{
+    resp[0] = 0x80; resp[1] = 0x01;
```
dgarske: Comment on each line what part of the response structure it is... Looks like we are building manually an error response.

src/fwtpm/fwtpm_main.c
```diff
@@ -81,19 +81,48 @@ static void usage(const char* progname)
...
+}
+
 static int fwtpmSpdmTpmDispatch(void* userCtx,
     const byte* cmd, word32 cmdSz,
     byte* resp, word32 respBufSz, word32* respSz)
 {
+    /* The dispatcher requires a full-size response buffer, and the SPDM
+     * envelope is smaller, so stage here and copy what fits. The server
+     * handles one command at a time. */
+    static byte stageBuf[FWTPM_MAX_COMMAND_SIZE];
```
dgarske, line 105: Why `static`? Why do we need a "stage" buffer? That's a lot of extra memory.

src/tpm2_asn.c
```diff
@@ -366,31 +366,65 @@ int TPM2_ASN_DecodeRsaPubKey(uint8_t* input, int inputSz,
...
 
-    if (*sigSz < 3)
-        return rc;
+/* 1 when a is less than b, for values well below the unsigned range */
+static int TPM2_ASN_CtLt(unsigned int a, unsigned int b)
+{
+    return (int)(((a - b) >> TPM2_ASN_CT_SHIFT) & 1u);
+}
 
-    if (sig[idx++] == 0x00 && sig[idx++] == 0x01) {
```
dgarske, line 378: This code was lifted out of wolfBoot and has also been involved in DO-178, so please explain why its changed and keep the code simple and readable. It. can be changed, but I'd like it to better match the old code if possible.

wolftpm/tpm2_packet.h
```diff
@@ -98,6 +98,9 @@ typedef struct TPM2_Packet {
     byte* buf;
     int pos;
     int size;
+    /* Set when an append was dropped because it would not fit in buf, so
```
dgarske: Condense this comment... fairly obvious. Fix on same line if possible. Also consider unsigned and bit-field for future room.

dgarske (changes requested):

src/fwtpm/fwtpm_command.c
aidangarske: makes sense, moved into session block

src/fwtpm/fwtpm_main.c
aidangarske: done

src/fwtpm/fwtpm_main.c
aidangarske: ctx->rspBuf cant be used so it needs to be static

aidangarske: writing to it would corrupt the response being built

src/tpm2_asn.c
aidangarske: Im gonna revert this one that makes sense this was a low timing corruption issue but not worth fixing

aidangarske: documented here https://github.com/wolfSSL/wolfTPM/pull/561/changes/6cbf2d14d7e12c86b68a9c3b0a3b43a433b24f00

dgarske (approved):

PR #574, opened 2026-08-13
Add finer per-command-group gating macros in fTPM

dgarske:
## Summary

Adds opt-in macros that let a firmware-TPM (fwTPM) build compile out command groups it does not need, so wolfTPM's fwTPM can fit deeply memory-constrained targets. All new macros default OFF, so the default fwTPM command set is unchanged and no existing build is affected.

## What is added

Six new per-group gates plus one umbrella, following the existing `FWTPM_NO_POLICY` / `NO_ATTESTATION` / `NO_CREDENTIAL` / `NO_DA` / `NO_PARAM_ENC` precedent:

| Macro | Compiles out |
|-------|--------------|
| `FWTPM_NO_KEY_MIGRATION` | Import, Duplicate, Rewrap (keeps the shared wrap/import helpers, LoadExternal, CreateLoaded) |
| `FWTPM_NO_ECDH` | ECDH_KeyGen, ECDH_ZGen, EC_Ephemeral, ZGen_2Phase, ECC_Parameters (and the `ecEphemeral*` context fields) |
| `FWTPM_NO_HASH_CMDS` | Hash, HMAC and the hash sequence commands (kept alive when `WOLFTPM_MLDSA` still needs the shared sequence path) |
| `FWTPM_NO_CONTEXT` | ContextSave, ContextLoad (keeps FlushContext) |
| `FWTPM_NO_SYM_ENCRYPT` | EncryptDecrypt, EncryptDecrypt2 |
| `FWTPM_NO_CLOCK` | ReadClock, ClockSet, ClockRateAdjust |

There is intentionally no "minimal" umbrella macro: dropping a command group removes real TPM functionality, so a build must select each gate deliberately rather than flipping one switch. Applying all six new gates plus the five pre-existing `FWTPM_NO_*` gates leaves a core fTPM (Startup / GetCapability / GetRandom / PCR / Create / Load / Sign / VerifySignature / NV / sessions).

## Why it is safe

`TPM2_GetCapability` derives the advertised command list and `TPM_PT_TOTAL_COMMANDS` from the same `fwCmdTable[]` dispatch table, so removing a table row removes the command from capability reporting automatically - no separate GetCapability edits are needed and a gated command cannot be advertised-but-missing. Shared helpers used by the retained commands (key wrap/import, hash sequence lookups used by MLDSA signing, FlushContext) are explicitly kept. The MLDSA sign path shares the TPM sequence commands, so those are guarded with `#if !defined(FWTPM_NO_HASH_CMDS) || defined(WOLFTPM_MLDSA)` to avoid breaking a gated + MLDSA build.

## Footprint impact

On a 32-bit RISC-V `-Os` build, enabling all of these gates removes roughly 20 KB of code from the fwTPM command engine. Combined with a hardware-TRNG entropy source this is what lets an ECC-only fwTPM fit a 192 KB soft-core target (see the AMD SCU35 example in wolftpm-examples, which selects the gate set explicitly in its user_settings.h).

## Testing

- Compile-clean across every new `-DFWTPM_NO_*` flag individually and with all of them enabled together, including the `FWTPM_NO_HASH_CMDS` + `WOLFTPM_MLDSA` combination.
- CI (`.github/workflows/fwtpm-test.yml`): build-only matrix entries for each new flag plus an all-gates-together build standalone and with MLDSA. Functional unit tests continue to run on the default (full) build; gated commands are guarded there by the matching `#ifndef`.

## Docs

`docs/FWTPM.md` feature-group table (new macros + the previously missing `FWTPM_NO_PARAM_ENC` row), `src/fwtpm/README.md` (macro list + build-only matrix rows), root `README.md`, and a `ChangeLog.md` entry.

### DIFF
--- a/.github/workflows/fwtpm-test.yml
+++ b/.github/workflows/fwtpm-test.yml
@@ -149,6 +149,64 @@ jobs:
             build_only: true
             extra_cflags: -DFWTPM_NO_PARAM_ENC
 
+          # Build-only: finer per-command-group FWTPM_NO_* macros
+          - name: fwtpm-no-key-migration
+            os: ubuntu-latest
+            wolftpm_config: --enable-fwtpm --enable-swtpm
+            wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-keygen
+            build_only: true
+            extra_cflags: -DFWTPM_NO_KEY_MIGRATION
+
+          - name: fwtpm-no-ecdh
+            os: ubuntu-latest
+            wolftpm_config: --enable-fwtpm --enable-swtpm
+            wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-keygen
+            build_only: true
+            extra_cflags: -DFWTPM_NO_ECDH
+
+          - name: fwtpm-no-hash-cmds
+            os: ubuntu-latest
+            wolftpm_config: --enable-fwtpm --enable-swtpm
+            wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-keygen
+            build_only: true
+            extra_cflags: -DFWTPM_NO_HASH_CMDS
+
+          - name: fwtpm-no-context
+            os: ubuntu-latest
+            wolftpm_config: --enable-fwtpm --enable-swtpm
+            wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-keygen
+            build_only: true
+            extra_cflags: -DFWTPM_NO_CONTEXT
+
+          - name: fwtpm-no-sym-encrypt
+            os: ubuntu-latest
+            wolftpm_config: --enable-fwtpm --enable-swtpm
+            wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-keygen
+            build_only: true
+            extra_cflags: -DFWTPM_NO_SYM_ENCRYPT
+
+          - name: fwtpm-no-clock
+            os: ubuntu-latest
+            wolftpm_config: --enable-fwtpm --enable-swtpm
+            wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-keygen
+            build_only: true
+            extra_cflags: -DFWTPM_NO_CLOCK
+
+          # Build-only: fewer PCRs than the 24 standard ones. The locality
+          # tables are sized to a fixed 24 while IMPLEMENTATION_PCR sets the
+          # implemented count, so this leg proves the two boundaries compile
+          # apart. PLATFORM_PCR must drop with it or PCR_SELECT_MIN exceeds
+          # PCR_SELECT_MAX. Runtime coverage is in fwtpm-gated-runtime.
+          - name: fwtpm-reduced-pcr
+            os: ubuntu-latest
+            wolftpm_config: --enable-fwtpm --enable-swtpm
+            wolfssl_config: --enable-wolftpm --enable-pkcallbacks --enable-keygen
+            build_only: true
+            extra_cflags: -DIMPLEMENTATION_PCR=8 -DPLATFORM_PCR=8
+
+          # Note: the all-gates combinations are built AND run by the
+          # fwtpm-gated-runtime job below, so they are not duplicated here.
+
           # Build-only: cross-algorithm + feature macro combinations
           - name: fwtpm-no-rsa-no-policy
             os: ubuntu-latest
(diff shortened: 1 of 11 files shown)


src/fwtpm/fwtpm_command.c
```diff
@@ -16167,19 +16200,29 @@ static const FWTPM_CMD_ENTRY fwCmdTable[] = {
...
     /* --- Hash/HMAC --- */
+#ifndef FWTPM_NO_HASH_CMDS
     { TPM_CC_Hash,               FwCmd_Hash,                 0, 0, 0, FW_CMD_FLAG_ENC | FW_CMD_FLAG_DEC },
     { TPM_CC_HMAC,               FwCmd_HMAC,                 1, 1, 0, FW_CMD_FLAG_ENC | FW_CMD_FLAG_DEC },
     { TPM_CC_HMAC_Start,         FwCmd_HMAC_Start,           1, 1, 1, FW_CMD_FLAG_ENC },
     { TPM_CC_HashSequenceStart,  FwCmd_HashSequenceStart,    0, 0, 1, FW_CMD_FLAG_ENC },
+#endif /* !FWTPM_NO_HASH_CMDS */
+    /* SequenceUpdate/SequenceComplete are shared with MLDSA sign sequences -
+     * keep them whenever MLDSA is built, even if the hash commands are gated. */
+#if !defined(FWTPM_NO_HASH_CMDS) || defined(WOLFTPM_MLDSA)
```
aidangarske, line 16242: **🟠 [Medium] ML-DSA build retains an unusable SequenceComplete command** · _Logic_

With both FWTPM_NO_HASH_CMDS and WOLFTPM_MLDSA defined, the PR applies the MLDSA exception to both sequence commands, retaining and advertising SequenceComplete even though only SequenceUpdate supports an MLDSA sign-sequence handle. The producers of hash sequence handles (HMAC_Start / HashSequenceStart) and FwAllocHashSeq are compiled out, but FwCmd_SequenceComplete looks up FwFindHashSeq() / FWTPM_HashSeq exclusively, so every constructible handle returns TPM_RC_HANDLE; given an MLDSA sign-sequence handle it additionally frees that sequence as misrouted. The advertised command therefore can never succeed and leaves dead code in a footprint-sensitive build. Severity views differ (review: Medium; review-security: Low); the stricter Medium is kept — both agree the command is unusable and should not be advertised.

**Fix:** Keep SequenceUpdate under the combined hash/MLDSA condition, but compile and advertise SequenceComplete only when FWTPM_NO_HASH_CMDS is not defined. Guard both the FwCmd_SequenceComplete definition and its dispatch-table entry with #ifndef FWTPM_NO_HASH_CMDS, and update the docs that claim both commands are shared with MLDSA.

wolftpm/fwtpm/fwtpm.h
```diff
@@ -110,6 +110,30 @@
...
+ * groups a minimal build does not need. All default OFF (the full command set is
+ * built). Removing a group drops its handlers AND its TPM2_GetCapability
+ * advertisement (the command list is derived from the dispatch table).
+ *   FWTPM_NO_POLICY        - policy session commands (TPM2_Policy*)
+ *   FWTPM_NO_ATTESTATION   - Quote/Certify/CertifyCreation/GetTime/NV_Certify
+ *   FWTPM_NO_CREDENTIAL    - MakeCredential/ActivateCredential
+ *   FWTPM_NO_DA            - dictionary-attack lockout protection (see below)
+ *   FWTPM_NO_PARAM_ENC     - command/response parameter encryption
+ *   FWTPM_NO_NV            - all NV_* commands
+ *   FWTPM_NO_KEY_MIGRATION - Import/Duplicate/Rewrap
```
aidangarske, line 123: **🟠 [Medium] Feature gates retain per-instance FWTPM_CTX storage that is no longer reachable** · _Logic_

The new gates remove handlers but leave their per-instance storage and initialization unconditional. hashSeq[FWTPM_MAX_HASH_SEQ] alone occupies ~3,520 bytes on the current ABI, the context-protection key/counters/live-handle array remains, and the ecEphemeral* fields (ecEphemeralCounter/ecEphemeralKey[FWTPM_MAX_PRIVKEY_DER]/ecEphemeralKeySz/ecEphemeralCurve) remain, ~1,280 bytes (or 256 bytes in an ECC-only NO_RSA build). Context init and context/ECDH-only helpers also stay compiled. For FWTPM_NO_ECDH specifically, all users of the ecEphemeral state are removed yet FWTPM_CTX still carries those members and the startup path still initializes them under HAVE_ECC alone (fwtpm_command.c ~13848-14282), directly contradicting the documented behavior that FWTPM_NO_ECDH removes the ecEphemeral* fields. This materially undermines the memory-constrained-target goal of the gates.

**Fix:** Guard the associated FWTPM_CTX members and their init, lookup, cleanup, and private helpers consistently. Gate the ecEphemeralCounter/ecEphemeralKey/ecEphemeralKeySz/ecEphemeralCurve members and their startup init with HAVE_ECC && !FWTPM_NO_ECDH, and preserve only the sign-sequence state needed by the MLDSA exception.

aidangarske (commented):
## Skoll Multi-Scan Review

### Posted findings
- [Medium] [review+review-security] **ML-DSA build retains an unusable SequenceComplete command** — `src/fwtpm/fwtpm_command.c:16211-16214`
- [Medium] [review+review-security] **Feature gates retain per-instance FWTPM_CTX storage that is no longer reachable** — `wolftpm/fwtpm/fwtpm.h:123-129`
- [Medium] [review+review-security] **New feature gates receive no behavioral/runtime test coverage** — `.github/workflows/fwtpm-test.yml:153-232`
- [Medium] [review] **Reduced-PCR table change lacks a reduced-PCR build test** — `src/fwtpm/fwtpm_command.c:2148-2188`
- [Medium] [review] **Command coverage docs still label newly gated commands as always enabled** — `src/fwtpm/README.md:232-245`

.github/workflows/fwtpm-test.yml
```diff
@@ -149,6 +149,88 @@ jobs:
             build_only: true
             extra_cflags: -DFWTPM_NO_PARAM_ENC
 
+          # Build-only: finer per-command-group FWTPM_NO_* macros
+          - name: fwtpm-no-key-migration
```
aidangarske, line 153: **🟠 [Medium] New feature gates receive no behavioral/runtime test coverage** · _Missing Tests_

Every newly added matrix configuration sets build_only: true and no test or example source references any of the six new macros, so CI verifies compilation only. It never checks that a gated command returns TPM_RC_COMMAND_CODE, disappears from TPM_CAP_COMMANDS, or updates TPM_PT_TOTAL_COMMANDS, nor that the FWTPM_NO_HASH_CMDS + MLDSA sequence path works. This gap is exactly what let the retained-but-unusable MLDSA SequenceComplete entry pass. Additionally, the documented all-gates ECC-only/NO_RSA minimal configuration is not built because the all-gates entry does not disable RSA. Severity views differ (review: Medium; review-security: Low); stricter Medium kept.

**Fix:** Add at least one runtime gated configuration that checks command rejection and capability/count reporting, plus an MLDSA sequence test under FWTPM_NO_HASH_CMDS. Also build the documented all-gates ECC-only/NO_RSA configuration.

src/fwtpm/fwtpm_command.c
```diff
@@ -2145,7 +2145,13 @@ static TPM_RC FwCmd_PCR_Read(FWTPM_CTX* ctx, TPM2_Packet* cmd, int cmdSize,
  * dump (loc2 reset of PCR 20-22 hardware-confirmed over SPI).
  *   reset:  0-15 never; 16,23 loc0-3; 17-19 loc4; 20-22 loc2-4.
  *   extend: 0-16,23 any; 17,18 loc2-4; 19 loc2-3; 20 loc1-3; 21,22 loc2. */
-static const byte fwPcrResetLocality[IMPLEMENTATION_PCR] = {
+/* The TCG PCR locality attributes are defined for the 24 standard PCRs. Size the
```
aidangarske, line 2164: **🟠 [Medium] Reduced-PCR table change lacks a reduced-PCR build test** · _Missing Tests_

The PR changes table sizing and validation specifically to support IMPLEMENTATION_PCR < 24, but none of the added CI configurations overrides IMPLEMENTATION_PCR. The motivating configuration therefore remains uncompiled and the new dual-boundary behavior (IMPLEMENTATION_PCR vs FW_PCR_LOCALITY_TABLE) is untested.

**Fix:** Add a CI build with a representative value such as -DIMPLEMENTATION_PCR=8, and exercise the highest implemented PCR plus the first unavailable PCR.

src/fwtpm/README.md
```diff
@@ -221,6 +229,26 @@ EncryptDecrypt, EncryptDecrypt2
 - `FWTPM_NO_PARAM_ENC`: Disables parameter encryption/decryption for command and
   response parameters. Sessions still work for HMAC auth, but encrypted transport
   is disabled. Reduces code size by removing AES-CFB and XOR param encryption.
+- `FWTPM_NO_KEY_MIGRATION`: Import, Duplicate, Rewrap (3 commands). Shared key
```
aidangarske, line 257: **🟠 [Medium] Command coverage docs still label newly gated commands as always enabled** · _Documentation_

The new macro descriptions say commands such as ReadClock, ContextSave, hash sequences, and Import can be disabled, while the preceding command-coverage section still lists them under "Always enabled (47 commands)". The new feature makes that count and category inaccurate for gated builds.

**Fix:** Rename that section to describe the default command set, or move the newly gated commands into conditional groups and adjust the count. Also include the six new macros in the earlier feature-disable macro summary.

aidangarske (approved):
Looks good!

PR #576, opened 2026-08-17
Add support for the NVIDIA Jetson Orin OP-TEE firmware TPM

dgarske:
Adds the **NVIDIA Jetson Orin** (Tegra234) as a supported target. Its TPM 2.0 is a **firmware TPM** running as an **OP-TEE** trusted application rather than a discrete part on a bus, reached through the Linux kernel driver at `/dev/tpmrm0`. `wolfTPM2_ParseCapabilities()` now recognizes the `MSFT` manufacturer string and reports the new `TPM_MFG_MSFT` (appended to `WOLFTPM2_MFG`, existing values unchanged, and deliberately without `req_wait_state` since there is no TIS bus); it previously read `Mfg MSFT (0)` = `TPM_MFG_UNKNOWN`. Also adds **`docs/DEVTPM.md`**, the first documentation for the kernel `/dev/tpmX` transport, plus README identification and benchmark entries, a `devtpm-autodetect` CI build entry, and a packaging fix for `docs/FWTPM.md`, which was never listed in `docs/include.am`. Identification reads `Mfg MSFT (7), Vendor SSE fTPM, Fw 8216.1808 (0x105300), FIPS 140-2, CC-EAL4 0` — note `Fw 8216.1808` is `0x20180710`, a build **date** rather than a version, and is captioned as such in the README.

Two behavior changes and one real bug. The bug: **`--enable-autodetect` only autodetected for the wrapper API** — `TPM2_LINUX_TryOpen()` was called solely from `wolfTPM2_Init_ex()`, so the native `TPM2_Init_ex()` left `fd = -1`, fell through to TIS/SPI, and every native-API caller failed at init with `TPM_RC_FAILURE`, including the shipped `native_test`. Fixed by probing in the native init and making `TryOpen` idempotent so the wrapper's call no longer leaks a descriptor. Separately, `wolfTPM2_Reset()` no longer sends `TPM2_Shutdown`/`TPM2_Startup` on the kernel and TBS transports, matching `wolfTPM2_Init()` which already skips startup there — to be clear about scope, I expected the resource manager to *reject* those commands and that proved **false**: on Linux 5.15 it passes `TPM_SU_CLEAR` through and it succeeds, so this is hardening (one caller should not make a global state change on a TPM shared with every process) rather than a crash fix. Configure now also emits a notice when it auto-selects the software TPMs, because on Linux x86_64/aarch64 a bare `./configure` enables swTPM and thereby suppresses kernel-device autodetect — the build silently talks to a simulator on port 2321; the README text claiming otherwise is corrected, while the default itself is left unchanged since probing `/dev/tpm*` at configure time would break cross-compiles. **Tested** on Jetson Linux R36.4.4 / kernel 5.15.148-tegra: `run_examples.sh` reports `Success!` (native, wrapper, NV, policy, PKCS7, full TLS matrix for RSA and ECC over TLS 1.2/1.3 with and without parameter encryption, attestation, PCR quote/policy, Secure Boot ROT, Seal/Unseal, EK cert), every benchmark operation supported with none skipped, all identification values cross-checked against `tpm2_getcap`, and x86_64 regression builds clean in `devtpm`, `autodetect`, `infineon` and default configurations. Two items are documented in `docs/DEVTPM.md` rather than patched, since neither is a wolfTPM defect: `/dev/tpmrm0` virtualizes transient handles and flushes them when the descriptor closes, so `run_examples.sh`'s provisioning block (which creates IAK/IDevID with `-keep` in one process and references `0x80000000` from another) cannot pass on the resource manager by construction; and `run_examples.sh:190` prints `$RESULT$RESULT`, reporting an exit code of `1` as `failed! 11`.

### DIFF
--- a/.github/workflows/make-test-swtpm.yml
+++ b/.github/workflows/make-test-swtpm.yml
@@ -193,6 +193,19 @@ jobs:
             needs_swtpm: false
             test_command: "true"
 
+          # Kernel-device autodetect: /dev/tpmrm0 -> /dev/tpm0 -> SPI fallback.
+          # The only entry defining WOLFTPM_LINUX_DEV_AUTODETECT, so it is also
+          # the only one compiling that path's DEBUG_WOLFTPM code -- hence
+          # --enable-debug, which keeps those printfs under a compiler.
+          # Build-only: the runner has no /dev/tpm* and this config cannot talk
+          # to swtpm, so wolfTPM2_Init has no transport and unit.test fails at
+          # its first assertion. Runtime coverage needs a self-hosted runner
+          # with a kernel-bound TPM -- see docs/DEVTPM.md.
+          - name: devtpm-autodetect
+            wolftpm_config: --enable-autodetect --enable-debug
+            needs_swtpm: false
+            test_command: "true"
+
           # Negative tests: configure must error on conflicting flag combos,
           # and we verify the SPECIFIC error message (not just a non-zero
           # exit). wolfSSL is installed by the earlier `Setup wolfSSL` step,
@@ -252,7 +265,10 @@ jobs:
             wolftpm_config: --enable-advio --disable-fwtpm
             needs_swtpm: false
 
-          # Autodetect (default configure, /dev/tpm0 + SPI dual support)
+          # Default configure. NOTE this does NOT build the /dev/tpmX path:
+          # WOLFTPM_SWTPM is still auto-enabled here, which suppresses
+          # WOLFTPM_LINUX_DEV_AUTODETECT (wolftpm/tpm2_types.h). The kernel
+          # device path is covered by the devtpm-autodetect entry above.
           - name: autodetect
             wolftpm_config: "--disable-fwtpm"
             needs_swtpm: false
--- a/configure.ac
+++ b/configure.ac
@@ -309,6 +309,15 @@ then
 
     AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_SWTPM"
 
+    # WOLFTPM_SWTPM suppresses the /dev/tpmX autodetect (see
+    # WOLFTPM_LINUX_DEV_AUTODETECT in wolftpm/tpm2_types.h), so when it is
+    # reached by default rather than by request, a host with a real kernel
+    # TPM would silently build against the simulator. Say so.
+    if test "x$WOLFTPM_DEFAULT_SWTPM" = "xyes" && test "x$enable_swtpm" = "x"; then
+        AC_MSG_NOTICE([no hardware TPM interface selected: defaulting to the software TPM (swTPM) so 'make check' works without hardware.])
+        AC_MSG_NOTICE([this build will NOT use /dev/tpmrm0 or /dev/tpm0. For a real TPM use --enable-autodetect, --enable-devtpm, or --enable-<vendor>.])
+    fi
+
     if test "x$ENABLED_SWTPM" = "xuart"
     then
         AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_SWTPM_UART"
--- a/docs/DEVTPM.md
+++ b/docs/DEVTPM.md
@@ -1,0 +1,173 @@
+# wolfTPM with the Linux Kernel TPM Device (/dev/tpmX)
+
+On Linux the kernel's TPM driver stack exposes a TPM through a character device, and wolfTPM can use it directly instead of driving SPI or I2C itself. This is the right transport whenever the kernel already owns the TPM: a discrete chip bound to a kernel driver, a Windows-style firmware TPM, or a TEE-resident firmware TPM such as the one on NVIDIA Jetson platforms.
+
+With `--enable-devtpm` there is **no TIS layer and no HAL IO callback**: `hal/tpm_io.c` is compiled out entirely and `TPM2_IoCb` is `NULL` (see `hal/tpm_io.h`), so pass `NULL` for the callback argument of `TPM2_Init` / `wolfTPM2_Init`.
+
+With `--enable-autodetect` this is **not** the case. The TIS/SPI HAL stays compiled in on purpose - it is the fallback - and `TPM2_IoCb` is a real function. Keep passing it, or the SPI fallback that build exists to provide is unreachable.
+
+## Two device nodes
+
+The kernel presents up to two nodes per TPM:
+
+* `/dev/tpm0` - the raw device. One user at a time, no resource management. Whatever you send reaches the TPM.
+* `/dev/tpmrm0` - the in-kernel resource manager (kernel 4.12+, practical from 5.12+). It virtualizes handles, swaps transient objects and sessions in and out as needed, and flushes everything belonging to a connection when that connection closes.
+
+wolfTPM prefers `/dev/tpmrm0` and falls back to `/dev/tpm0`. The resource manager is the better default: a TPM has very few transient object slots, and without it a program that leaks a handle wedges the TPM for everything else on the system.
+
+Build-time overrides, honored by both `--enable-devtpm` and `--enable-autodetect`:
+
+* `-DWOLFTPM_USE_TPMRM` - use `/dev/tpmrm0` only, with no fallback to the raw device.
+* `CFLAGS='-DTPM2_LINUX_DEV="/dev/tpm1"'` - use a specific node. The inner quotes are required: the macro is used directly as a C string literal, so an unquoted value does not compile.
+
+## Startup, shutdown, and shared state
+
+The TPM is started by firmware long before Linux runs, and on the resource manager it is shared with every other process on the system. Restarting or shutting it down is therefore not an individual caller's decision, so wolfTPM stays out of the way on this transport:
+
+* `wolfTPM2_Init` skips the startup and self-test sequence.
+* `wolfTPM2_Reset` and `wolfTPM2_Shutdown` send no TPM command and return `NOT_COMPILED_IN` (-174), the same way `wolfTPM2_SetLocality` does on this transport. A `wolfTPM2_Reset(dev, 0, 0)` that asked for neither a shutdown nor a startup still returns `TPM_RC_SUCCESS`, since nothing was declined. Treat `NOT_COMPILED_IN` here as "the OS owns this", not as a failure.
+* `wolfTPM2_SetLocality` returns `NOT_COMPILED_IN` - the kernel owns the locality.
+
+Be aware that the kernel does **not** reliably stop you here. Command filtering on `/dev/tpmrm0` is primarily about handle isolation, not about blocking global state changes, and behavior varies by kernel version and TPM implementation. On Linux 5.15 with the Jetson OP-TEE fTPM, a `TPM2_Shutdown(TPM_SU_CLEAR)` sent through the resource manager is passed straight through and returns success - both from wolfTPM and from `tpm2_shutdown`. So this is a case where the library declining to send the command is what protects other users of the TPM, rather than the kernel doing it for you.
+
+If you genuinely need to control TPM startup state, you need `/dev/tpm0` and exclusive use of the TPM, or direct SPI access with wolfTPM's own TIS driver.
+
+## What the native API does on autodetect builds
+
+Two behaviors worth knowing if you use `TPM2_Init` / `TPM2_Init_ex` directly rather than the `wolfTPM2_*` wrapper.
+
+**The kernel device wins over your callback.** If `/dev/tpmrm0` or `/dev/tpm0` opens, every command is routed there and the HAL IO callback you passed is never invoked. On a host that has both a kernel-bound TPM and a discrete SPI part, that means you now talk to a different TPM than a pre-autodetect build did. Pin the part you want with `--enable-devtpm`, `--enable-spi` / `--enable-<vendor>`, or `-DTPM2_LINUX_DEV`.
+
+**Init now acquires a descriptor.** `TPM2_Init*` opens the device on autodetect builds, and `TPM2_Cleanup()` is what closes it. Native callers that skipped cleanup previously leaked nothing; now they leak a descriptor per context. This matters most on hosts exposing only the raw `/dev/tpm0`, which permits a single open - a context that merely initialized holds the TPM exclusively for its lifetime, and a second context in the same process falls through to a different transport.
+
+`TPM2_Init_minimal()` is unaffected: it performs no IO and still succeeds with no device present.
+
+## Transient handles do not outlive a process
+
+This is the difference most likely to break an existing application.
+
+On `/dev/tpmrm0` the kernel gives each open file description its own handle space. Transient object handles are **virtualized** - the value the TPM assigned is not the value you get back - and everything in that space is **flushed when the file descriptor closes**. So a transient key created by one process is gone by the time a second process runs, and the handle number it printed is meaningless to anyone else.
+
+Creating a primary key on the Jetson fTPM through the resource manager returns:
+
+```
+Create Primary Handle: 0x80ffffff
+```
+
+not the `0x80000000` a raw device would report. Query the transient handles from a separate process afterwards and the list is empty:
+
+```bash
+tpm2_getcap handles-transient      # no output - the space was torn down
+```
+
+Two practical consequences:
+
+* A "create a key, keep it, use it from the next command" workflow does not work across processes. Do the whole sequence in one process, or make the object persistent with `TPM2_EvictControl` so it gets a stable `0x81xxxxxx` handle that does survive.
+* Passing a hard-coded transient handle such as `0x80000000` on a command line will fail. The kernel rejects the reference before it reaches the TPM, and because that happens at the file-descriptor layer the error surfaces as an `errno 22 = Invalid argument` on `read()`, which wolfTPM reports as `TPM_RC_FAILURE` rather than as a handle error. If you see `TPM_RC_FAILURE` alongside `Failed to read from /dev/tpmrm0 ... errno 22`, suspect a stale or cross-process transient handle before suspecting the TPM.
+
+wolfTPM's own `examples/run_examples.sh` hits exactly this: its provisioning section creates IAK and IDevID primaries with `-keep` in one process and then references `0x80000000` / `0x80000001` from another. That block cannot pass on the resource manager by construction. Everything either side of it is unaffected. Use `/dev/tpm0` with exclusive access if you need to run it as written.
+
+## Building
+
+```bash
+./autogen.sh
+./configure --enable-devtpm
+make
+```
+
+`--enable-devtpm` uses the kernel node only. Use `--enable-autodetect` instead if you want wolfTPM to try `/dev/tpmrm0`, then `/dev/tpm0`, and finally fall back to probing SPI - useful for one binary that has to run on several boards.
+
+Only one transport can be enabled at a time. `--enable-devtpm` conflicts with `--enable-swtpm` and `--enable-winapi`, and configure will stop if you ask for more than one.
+
+### The x86_64 / aarch64 default
+
+A bare `./configure` on Linux `x86_64` or `aarch64` does **not** produce a build that talks to `/dev/tpmX`. On those hosts wolfTPM auto-enables the software TPMs (swTPM and fwTPM) so that `make check` passes with no hardware attached, and defining `WOLFTPM_SWTPM` suppresses the kernel-device autodetect path. The result talks to a simulator on TCP port 2321.
+
+Selecting any hardware path explicitly turns that default back off - `--enable-autodetect`, `--enable-devtpm`, or any `--enable-<vendor>`. Configure prints a notice when the software default is taken, so check the tail of its output if a build unexpectedly fails to find your TPM.
+
+This bites hardest on single-board `aarch64` machines with a firmware TPM, where the kernel device is the only transport there is.
+
+## Permissions
+
+The TPM character devices are not world-accessible. On a typical system they are mode `0660` owned by group `tss`:
+
+```
+crw-rw---- 1 tss root  10,   224 /dev/tpm0
+crw-rw---- 1 tss tss  252, 65536 /dev/tpmrm0
+```
+
+wolfTPM detects `EACCES` and reports it plainly:
+
+```
+Permission denied on /dev/tpm0
+Use sudo or add tss group to user.
+```
+
+The fix is to put your user in the owning group and start a new login session:
+
+```bash
+sudo usermod -aG tss $USER
+```
+
+Note that the `tss` group is created by tpm2-tss, and on distributions that ship it the group frequently exists with no members - so this step is required even though the group looks correctly set up.
+
+To use a group of your own instead, add a udev rule:
+
+1) Create the group and add your user:
+
+```bash
+sudo addgroup wolftpm
+sudo adduser [username] wolftpm
+```
+
+2) Create `/etc/udev/rules.d/wolftpm-udev.rules` containing:
+
+```
+KERNEL=="tpm[0-9]*", TAG+="systemd", MODE="0660", GROUP="wolftpm"
+```
+
+3) Reload the rules: `sudo udevadm control -R`, then re-plug or reboot.
+
+## NVIDIA Jetson Orin (Tegra234) firmware TPM
+
+Jetson Orin platforms carry a TPM 2.0 implemented in firmware, running as a trusted application inside OP-TEE rather than as a discrete package on a bus. Linux reaches it through the `tpm_ftpm_tee` driver, which speaks to the TA over the TEE interface and registers an ordinary TPM chip - so from wolfTPM's point of view it is just another `/dev/tpmrm0`.
+
+Confirm the device is present before building:
+
+```bash
+lsmod | grep tpm_ftpm_tee
+ls -l /dev/tpm*
+cat /sys/class/tpm/tpm0/tpm_version_major     # expect 2
+```
+
+If the module is missing, try `sudo modprobe tpm_ftpm_tee` and check that the kernel was configured with `CONFIG_TCG_FTPM_TEE`. On NVIDIA's Jetson Linux (L4T) images the driver is present and an `fTPM Device Provisioning Service` systemd unit runs at boot; you can see it complete in the boot log.
+
+Note that an OP-TEE boot message about silicon-identity fTPM provisioning not being enabled refers to a separate NVIDIA feature and does **not** mean the TPM 2.0 device is unavailable.
+
+Build as above with `--enable-devtpm` or `--enable-autodetect`, then confirm with:
+
+```bash
+./examples/wrap/caps
+```
+
+Because this is a firmware TPM, expect two differences from a discrete part. There is no TIS bus, so the `TPM2: Caps/Did/Vid/Rid` values do not exist and the device is identified purely from `TPM2_GetCapability` properties. Under `--enable-devtpm` the `DEBUG_WOLFTPM` line is still printed but reads all zeros; under `--enable-autodetect` `wolfTPM2_Init_ex` returns as soon as the kernel device opens, before that printf, so the line is absent entirely. And a firmware TPM's algorithm coverage is set by its firmware build rather than by a datasheet, so it is worth checking rather than assuming; where an operation is absent the benchmark reports it as unsupported rather than failing. The Jetson Orin fTPM supports every operation the benchmark exercises - see the README results.
+
+See the main [README.md](../README.md#device-identification) for this platform's identification values and benchmark results.
+
+## Testing
+
+The examples run unchanged on this transport:
+
+```bash
+./examples/wrap/caps
+./examples/native/native_test
+./examples/wrap/wrap_test
+./examples/bench/bench
+./examples/run_examples.sh
+```
+
+`run_examples.sh` already skips the locality test on backends that do not support it.
+
+## CI coverage
+
+Both `--enable-devtpm` and `--enable-autodetect` are build-tested in CI, but not run - GitHub-hosted runners have no `/dev/tpm*` node. Runtime coverage of this transport requires a self-hosted runner with a real TPM bound to the kernel driver.
--- a/src/tpm2.c
+++ b/src/tpm2.c
@@ -939,9 +939,30 @@ TPM_RC TPM2_Init_ex(TPM2_CTX* ctx, TPM2HalIoCb ioCb, void* userCtx,
     /* Set the active TPM global */
     TPM2_SetActiveCtx(ctx);
 
+#ifdef WOLFTPM_LINUX_DEV_AUTODETECT
+    /* Probe here, not only in wolfTPM2_Init_ex, so the native API autodetects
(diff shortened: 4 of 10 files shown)


docs/DEVTPM.md
```diff
@@ -0,0 +1,163 @@
...
+
+* `-DWOLFTPM_USE_TPMRM` - use `/dev/tpmrm0` only, with no fallback to the raw device.
+* `CFLAGS='-DTPM2_LINUX_DEV="/dev/tpm1"'` - use a specific node. The inner quotes are required: the macro is used directly as a C string literal, so an unquoted value does not compile.
+
+## Startup, shutdown, and shared state
+
+The TPM is started by firmware long before Linux runs, and on the resource manager it is shared with every other process on the system. Restarting or shutting it down is therefore not an individual caller's decision, so wolfTPM stays out of the way on this transport:
+
+* `wolfTPM2_Init` skips the startup and self-test sequence.
+* `wolfTPM2_Reset` and `wolfTPM2_Shutdown` are no-ops that return success.
```
aidangarske: **🔴 [High] New docs/DEVTPM.md contradicts the return value this PR gives wolfTPM2_Reset/Shutdown** · _Documentation_

The new doc states `wolfTPM2_Reset` and `wolfTPM2_Shutdown` "are no-ops that return success". The same PR makes them return `NOT_COMPILED_IN` (`-174`, see `wolftpm/tpm2_types.h:279`) on `WOLFTPM_LINUX_DEV`/`WOLFTPM_WINAPI` (`src/tpm2_wrap.c:8502`) and on the autodetect kernel path (`src/tpm2_wrap.c:8511`). The doxygen added in the same PR (`wolftpm/tpm2_wrap.h:3617-3623`, `3639-3645`) correctly says `NOT_COMPILED_IN` is returned, so the two pieces of new documentation disagree about the API contract of the one function this PR changes. This is the file the README now points users to for this transport, so it is the version a porter will read first. Both scan modes flagged the same doc/code disagreement; the security pass additionally notes that a user following DEVTPM.md as written will write `if (rc != 0) fail;` and take an error branch on every clean shutdown.

**Fix:** Pick one contract and make code, header Doxygen, and docs/DEVTPM.md agree. If `NOT_COMPILED_IN` is the intended return, correct the DEVTPM.md bullet so it matches the doxygen added to `wolftpm/tpm2_wrap.h` in this PR, and add a ChangeLog.md migration note for the changed return on devtpm/TBS builds.

**Suggestion:**
```suggestion
* `wolfTPM2_Reset` and `wolfTPM2_Shutdown` send no TPM command and return `NOT_COMPILED_IN` (-174), matching `wolfTPM2_SetLocality` on this transport.
```

src/tpm2.c
```diff
@@ -939,9 +939,28 @@ TPM_RC TPM2_Init_ex(TPM2_CTX* ctx, TPM2HalIoCb ioCb, void* userCtx,
     /* Set the active TPM global */
     TPM2_SetActiveCtx(ctx);
 
+#ifdef WOLFTPM_LINUX_DEV_AUTODETECT
+    /* Probe here, not only in wolfTPM2_Init_ex, so the native API autodetects
+     * /dev/tpmX too. TPM_RC_FAILURE = present but unopenable (permissions),
+     * so do not mask it with SPI; TPM_RC_INITIALIZE = absent, fall back to SPI. */
+    rc = TPM2_LINUX_TryOpen(ctx);
```
aidangarske: **🟠 [Medium] EACCES on /dev/tpm* now aborts native init in autodetect builds, defeating the SPI fallback** · _Logic_

BEFORE: `TPM2_Init_ex()` never probed the kernel device. In an `--enable-autodetect` build a native-API caller that passed `TPM2_IoCb` went straight to `TPM2_ChipStartup()` (TIS/SPI), regardless of the state of `/dev/tpm*`.

AFTER: `TPM2_Init_ex()` calls `TPM2_LINUX_TryOpen()` and returns immediately when it yields `TPM_RC_FAILURE`. `TPM2_LINUX_TryOpen()` (src/tpm2_linux.c:248-258) returns `TPM_RC_FAILURE` specifically when the last `open()` attempt failed with `EACCES`, which is the *normal* state for an unprivileged process on any Linux host with a kernel-bound TPM (`/dev/tpm0` and `/dev/tpmrm0` ship mode 0660 root:tss). So the single most common kernel-device failure mode is now converted into a hard init failure with no attempt at the fallback transport, even though `ctx->ioCb` is non-NULL and the TIS/SPI HAL is compiled in.

This directly contradicts the purpose of the build mode as stated in the PR's own new documentation (`docs/DEVTPM.md`: "The TIS/SPI HAL stays compiled in on purpose - it is the fallback") and README ("If the kernel driver is available it will be used; otherwise wolfTPM falls back to direct SPI access"). The distinction matters because `TPM2_LINUX_TryOpen()` already has a distinct "device not usable, keep going" return (`TPM_RC_INITIALIZE`) used for every other errno. See also the related finding on `TPM2_LINUX_TryOpen`'s errno handling in src/tpm2_linux.c, which can suppress the EACCES branch entirely.

**Fix:** Only treat the EACCES failure as fatal when there is no fallback transport available (`ctx->ioCb == NULL`). Otherwise fall through so `TPM2_ChipStartup()` can run, preserving the documented SPI fallback. The permission-denied message printed by `TPM2_LINUX_TryOpen()` is still emitted, so diagnosability is unchanged.

**Suggestion:**
```suggestion
rc = TPM2_LINUX_TryOpen(ctx);
    if (rc == TPM_RC_FAILURE && ctx->ioCb == NULL) {
        return rc;
    }
```

src/tpm2_wrap.c
```diff
@@ -8474,13 +8482,36 @@ int wolfTPM2_HmacFinish(WOLFTPM2_DEV* dev, WOLFTPM2_HMAC* hmac,
...
 
+#if defined(WOLFTPM_LINUX_DEV) || defined(WOLFTPM_WINAPI)
+    /* The kernel driver and Windows TBS own TPM startup state, as they own the
+     * locality (see wolfTPM2_SetLocality, which returns NOT_COMPILED_IN here
+     * for the same reason). On the shared /dev/tpmrm0 a TPM_SU_CLEAR from one
+     * caller would hit every other process using the TPM, and the kernel does
+     * not block it - on Linux 5.15 it passes through and succeeds. */
+    (void)doShutdown;
+    (void)doStartup;
+    rc = NOT_COMPILED_IN;
```
aidangarske: **🟠 [Medium] wolfTPM2_Reset/wolfTPM2_Shutdown now return NOT_COMPILED_IN, breaking callers and the degenerate (0,0) call** · _Logic_

On `WOLFTPM_LINUX_DEV`/`WOLFTPM_WINAPI` builds (src/tpm2_wrap.c:8494-8513) the function now returns `NOT_COMPILED_IN` unconditionally, before looking at `doShutdown`/`doStartup`. Refusing to send `TPM2_Shutdown`/`TPM2_Startup` on a shared kernel device is the point of the change and is reasonable, but two regressions follow.

1. Unannounced public-API contract change: every external caller written as `rc = wolfTPM2_Shutdown(&dev, 1); if (rc != TPM_RC_SUCCESS) { /* error */ }` now takes an error branch on a normal clean shutdown. `wolfTPM2_Shutdown()` (src/tpm2_wrap.c:8550) forwards to `wolfTPM2_Reset()` and inherits the new return. The two in-tree callers (`examples/wrap/caps.c:141`, `examples/wrap/wrap_test.c:1097`) discard the return, so CI will not catch this.
2. The caller that asks for neither (`wolfTPM2_Reset(dev, 0, 0)`) previously fell through both blocks and returned `TPM_RC_SUCCESS`. That call is now a hard error despite the library having been asked to do nothing. The autodetect branch has the same shape (`rc = NOT_COMPILED_IN` is set whenever `dev->ctx.fd >= 0`, regardless of the requested flags).

Additionally, on `WOLFTPM_LINUX_DEV_AUTODETECT` builds the return depends on the *runtime* transport (`dev->ctx.fd >= 0`), so a caller cannot compile-time-detect which contract applies and must special-case `NOT_COMPILED_IN` in all builds. The two modes propose different fixes: the review pass would return success only for the no-op `(0,0)` case, the security pass would return `TPM_RC_SUCCESS` for the whole deliberate no-op to match what DEVTPM.md already tells users.

**Fix:** At minimum return `TPM_RC_SUCCESS` when neither a shutdown nor a startup was requested, and reserve `NOT_COMPILED_IN` for the case where a command was actually declined; apply the same guard in the `WOLFTPM_LINUX_DEV_AUTODETECT` branch (only set `rc = NOT_COMPILED_IN` if `doShutdown` or `doStartup` was 1 before being cleared). If the intent is that the whole call is a documented no-op, return `TPM_RC_SUCCESS` outright and fix the header doxygen; either way add a ChangeLog.md entry calling out the changed return on devtpm/TBS builds.

**Suggestion:**
```suggestion
if (doShutdown == 0 && doStartup == 0) {
        rc = TPM_RC_SUCCESS;
    }
    else {
        rc = NOT_COMPILED_IN;
    }
```

src/tpm2_linux.c
```diff
@@ -215,27 +215,45 @@ int TPM2_LINUX_SendCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
...
     #ifdef DEBUG_WOLFTPM
-        printf("Opened /dev/tpm0\n");
+        printf("Opened %s\n", TPM2_LINUX_DEV_FALLBACK);
     #endif
         return TPM_RC_SUCCESS;
     }
+#endif
 
-    /* Distinguish "not available" from "permission denied" */
+    /* Distinguish "not available" from "permission denied". Name the node
```
aidangarske: **🟠 [Medium] TryOpen still misattributes EACCES: permission failure on /dev/tpmrm0 is reported as "device absent"** · _Logic_

The rewritten block's comment claims it now names "the node actually tried", but the name is selected at compile time by `#ifdef TPM2_LINUX_DEV_FALLBACK` and the `errno` tested is only whatever the *last* `open()` left behind. If `/dev/tpmrm0` fails with `EACCES` and `/dev/tpm0` then fails with `ENOENT`, `errno` is `ENOENT`, the permission message is suppressed entirely, and the function returns `TPM_RC_INITIALIZE` — which `TPM2_Init_ex` (src/tpm2.c:946) and `wolfTPM2_Init_ex` both read as "no kernel device, fall back to SPI". That combination is exactly what a container that bind-mounts only `/dev/tpmrm0` produces, and it is now more visible because this PR routes the native API through the same probe. The user gets "Neither /dev/tpm0 nor SPI bus produced a TPM response" instead of the tss-group hint the code is trying to give.

**Fix:** Capture the per-open errno (declare `int errPrimary = 0; int errFallback = 0;` at the top of the function, C89 style) and report whichever node returned `EACCES`, returning `TPM_RC_FAILURE` if either did. That makes the code match the comment and stops a permissions problem from being silently downgraded to "not present".

tests/unit_tests.c
```diff
@@ -427,7 +427,13 @@ static void test_wolfTPM2_FirmwareUpgrade_ex_session(void)
     rc = wolfTPM2_GetCapabilities(&dev, &caps);
     AssertIntEQ(rc, 0);
     if (caps.mfg != TPM_MFG_UNKNOWN) {
```
aidangarske, line 429: **🟠 [Medium] TPM_MFG_MSFT widens the "real TPM, skip" gate to cover the ms-tpm-20-ref simulator** · _Missing Tests_

Three unit tests gate on `caps.mfg != TPM_MFG_UNKNOWN` as a proxy for "not the simulator" (lines 429, 556, 651). Adding `TPM_MFG_MSFT` makes that gate true for `ms-tpm-20-ref`, which `docs/SWTPM.md:122` documents as a supported simulator, so those three tests silently degrade to "Skipped (real TPM)" for anyone running `make check` against it. CI is unaffected today only by accident — it uses `ibmswtpm2` (`.github/workflows/make-test-swtpm.yml:428`, manufacturer `IBM`) and the wolfTPM fwTPM reports `WOLF` (`src/fwtpm/fwtpm_command.c:1377`) — so nothing will flag the loss. The PR documents the trade-off in a comment rather than fixing it, and every future manufacturer added to `WOLFTPM2_MFG` erodes this gate further.

**Fix:** Invert the gate so it detects the simulator positively rather than inferring it from an unrecognized manufacturer — e.g. `#if defined(WOLFTPM_SWTPM) || defined(WOLFTPM_FWTPM_HAL)` at compile time, or a helper that checks the vendor string. That keeps the tests running on every simulator and keeps the gate stable as new manufacturers are recognized.

src/tpm2.c
```diff
@@ -939,9 +939,28 @@ TPM_RC TPM2_Init_ex(TPM2_CTX* ctx, TPM2HalIoCb ioCb, void* userCtx,
...
+    if (rc == TPM_RC_FAILURE) {
+        return rc;
+    }
+    /* The SPI fallback needs an IO callback. With no kernel device and no
+     * callback there is no transport at all, and TPM2_TIS_SendCommand would
+     * call a NULL ctx->ioCb on the first command, so fail init here instead.
+     * wolfTPM2_Init_ex makes the same check for the wrapper API. */
+    if (ctx->fd < 0 && ctx->ioCb == NULL) {
```
aidangarske: **🔵 [Low] TPM2_Init_minimal() and NULL-callback native init now fail in autodetect builds with no kernel device** · _Logic_

BEFORE: `TPM2_Init_minimal(ctx)` (= `TPM2_Init_ex(ctx, NULL, NULL, 0)`, src/tpm2.c:977-980) returned `TPM_RC_SUCCESS` in an autodetect build — it only initialized the context and performed no IO, which is the documented purpose of the "minimal" variant.

AFTER: the new guard `if (ctx->fd < 0 && ctx->ioCb == NULL) return TPM_RC_FAILURE;` makes it return `TPM_RC_FAILURE` whenever `/dev/tpmrm0` and `/dev/tpm0` are absent or unopenable. The same applies to any native caller that deliberately passes a NULL HAL callback, which is the pattern the README documents for kernel-device builds ("The `TPM2_Init` or `wolfTPM2_Init` calls should use NULL for the HAL IO callback argument").

This is a public API (`WOLFTPM_API TPM_RC TPM2_Init_minimal`, wolftpm/tpm2.h:3718). The in-tree caller `tests/unit_tests.c:3997` casts the return to `(void)` and the later assertions still pass, so `make check` will not surface the change — and the new `devtpm-autodetect` CI entry is build-only, so no CI job runs this code path at all.

**Fix:** Only hard-fail when the caller actually asked for a transport bring-up (`timeoutTries > 0`, i.e. `TPM2_Init`), so `TPM2_Init_minimal()` keeps its no-IO success contract. If the hard failure is intended for all variants, document it on `TPM2_Init_minimal` in wolftpm/tpm2.h and add a ChangeLog entry.

**Suggestion:**
```suggestion
if (ctx->fd < 0 && ctx->ioCb == NULL && timeoutTries > 0) {
        return TPM_RC_FAILURE;
    }
    rc = TPM_RC_SUCCESS;
```

src/tpm2.c
aidangarske: **🔵 [Low] TPM2_Init_ex now acquires a file descriptor, so native callers that skip TPM2_Cleanup leak it (and can hold /dev/tpm0…** · _Resource Leak_

BEFORE: in an autodetect build `TPM2_Init_ex()` acquired no OS resources for native-API callers; the descriptor was only opened by the wrapper (`wolfTPM2_Init_ex`) or lazily in `TPM2_LINUX_SendCommand()`.

AFTER: every `TPM2_Init_ex()`/`TPM2_Init()`/`TPM2_Init_minimal()` call opens `/dev/tpmrm0` (or `/dev/tpm0`) when present. `TPM2_Cleanup()` closes it (src/tpm2.c:1022-1026), but native callers written against the old contract never call it.

Concrete in-tree instance introduced by this change: `tests/unit_tests.c:6584` (`test_wolfTPM2_thread_local_storage_work_thread`) calls `TPM2_Init(&tpm2Ctx, NULL, NULL)` on a stack context and never calls `TPM2_Cleanup()`. On any host with an accessible kernel TPM node, that test now leaks one descriptor per thread (two per run). A second consequence on hosts that expose only the raw `/dev/tpm0`: the kernel permits a single open of that node, so a context that merely initializes now holds the TPM exclusively for its lifetime — a second `TPM2_CTX` in the same process gets `EBUSY`, falls through to `TPM_RC_INITIALIZE`, and silently ends up on a different transport (TIS/SPI) than the first.

Related: `TPM2_Cleanup()` closes `ctx->fd` without resetting it to -1, so the pre-existing double-cleanup/close-of-reused-descriptor hazard now applies to native-API contexts as well.

**Fix:** Add `TPM2_Cleanup(&tpm2Ctx);` before the worker thread returns in tests/unit_tests.c, and set `ctx->fd = -1;` after `close(ctx->fd)` in `TPM2_Cleanup()` so repeat cleanup is idempotent. Document in docs/DEVTPM.md that on autodetect builds `TPM2_Init*` now acquires a descriptor that `TPM2_Cleanup()` must release.

.github/workflows/make-test-swtpm.yml
```diff
@@ -193,6 +193,16 @@ jobs:
             needs_swtpm: false
             test_command: "true"
 
+          # Kernel-device autodetect: /dev/tpmrm0 -> /dev/tpm0 -> SPI fallback.
+          # Compiles a different path than devtpm (WOLFTPM_LINUX_DEV_AUTODETECT
```
aidangarske: **🔵 [Low] New devtpm-autodetect CI entry is build-only and never compiles the DEBUG_WOLFTPM paths it added** · _Missing Tests_

`devtpm-autodetect` is the only matrix entry that defines `WOLFTPM_LINUX_DEV_AUTODETECT` (the `autodetect` and `autodetect-debug` entries are swTPM builds — with no explicit hardware selection `WOLFTPM_DEFAULT_SWTPM=yes` (configure.ac:269-285) defines `WOLFTPM_SWTPM`, which suppresses `WOLFTPM_LINUX_DEV_AUTODETECT` per wolftpm/tpm2_types.h:531-534), and it builds without `--enable-debug` and with `test_command: "true"`. The workflow skips the test step entirely when `test_command == 'true'` (line 483), so the only job that compiles the new code path never runs `make check` or any example.

Two consequences. (1) The `#ifdef DEBUG_WOLFTPM printf("Opened %s\n", TPM2_LINUX_DEV)` lines this PR added to `TPM2_LINUX_TryOpen` (src/tpm2_linux.c:229, 238) are not compiled anywhere in CI, so a typo or format-specifier mismatch in them would ship. (2) None of the behavioral changes in this PR — the new probe in `TPM2_Init_ex()`, the `TPM2_LINUX_TryOpen()` idempotency guard, the `wolfTPM2_Reset()` no-op return, and the `TPM_MFG_MSFT` parse branch — have any automated coverage; these are all return-value/control-flow changes that a build cannot catch. The review pass rated this a NIT (debug lines uncompiled), the security pass rated it a coverage gap for every behavior change in the PR; the stricter view is recorded here.

**Fix:** Add `--enable-debug` to the new entry (or add a second `devtpm-autodetect-debug` entry) so the debug-only code on the kernel-device path is at least compile-tested. Then add host-independent unit coverage that does not need `/dev/tpm*` — assert `TPM2_LINUX_TryOpen()` is idempotent (second call with `fd >= 0` returns success and does not change `fd`), assert the documented return of `wolfTPM2_Reset()`/`wolfTPM2_Shutdown()` on kernel/TBS transports, and add a `wolfTPM2_ParseCapabilities` case feeding `TPM_PT_MANUFACTURER = 0x4D534654` and checking `caps.mfg == TPM_MFG_MSFT` — and set `test_command: "make check"` for this entry.

**Suggestion:**
```suggestion
wolftpm_config: --enable-autodetect --enable-debug
```

.github/workflows/make-test-swtpm.yml
```diff
@@ -252,7 +262,10 @@ jobs:
             wolftpm_config: --enable-advio --disable-fwtpm
             needs_swtpm: false
 
-          # Autodetect (default configure, /dev/tpm0 + SPI dual support)
+          # Default configure. NOTE this does NOT build the /dev/tpmX path:
+          # WOLFTPM_SWTPM is still auto-enabled here, which suppresses
+          # WOLFTPM_LINUX_DEV_AUTODETECT (wolftpm/tpm2_types.h). The kernel
+          # device path is covered by the devtpm-autodetect entry below.
```
aidangarske: **🔵 [Low] CI comment points "below" at an entry that is above it** · _Documentation_

The `autodetect` entry's new comment says the kernel device path "is covered by the devtpm-autodetect entry below", but `devtpm-autodetect` is defined at line 198, roughly 70 lines above.

**Fix:** Change "below" to "above".

**Suggestion:**
```suggestion
# device path is covered by the devtpm-autodetect entry above.
```

configure.ac
```diff
@@ -309,6 +309,15 @@ then
 
     AM_CFLAGS="$AM_CFLAGS -DWOLFTPM_SWTPM"
 
+    # WOLFTPM_SWTPM suppresses the /dev/tpmX autodetect (see
+    # WOLFTPM_LINUX_DEV_AUTODETECT in wolftpm/tpm2_types.h), so when it is
+    # reached by default rather than by request, a host with a real kernel
+    # TPM would silently build against the simulator. Say so.
+    if test "x$WOLFTPM_DEFAULT_SWTPM" = "xyes" && test "x$enable_swtpm" = "x"; then
```
aidangarske, line 316: **🔵 [Low] Configure notice claims fwTPM was defaulted on even when --disable-fwtpm was passed** · _Documentation_

The notice fires whenever `WOLFTPM_DEFAULT_SWTPM=yes` and `enable_swtpm` is empty. `--disable-fwtpm` does not set `WOLFTPM_HW_SELECTED` (configure.ac:255-262), so `./configure --disable-fwtpm` — which is what most of the CI matrix uses — prints "defaulting to the software TPMs (swTPM/fwTPM)" even though fwTPM was explicitly turned off. The message is otherwise a good addition and the gating on `enable_swtpm` is correct.

**Fix:** Name only swTPM in the notice, or condition the "/fwTPM" wording on `ENABLED_FWTPM` actually being yes.

**Suggestion:**
```suggestion
AC_MSG_NOTICE([no hardware TPM interface selected: defaulting to the software TPM (swTPM) so 'make check' works without hardware.])
```

docs/DEVTPM.md
```diff
@@ -0,0 +1,163 @@
...
+
+Build as above with `--enable-devtpm` or `--enable-autodetect`, then confirm with:
+
+```bash
+./examples/wrap/caps
+```
+
+Because this is a firmware TPM, expect two differences from a discrete part. There is no TIS bus, so the `TPM2: Caps/Did/Vid/Rid` values do not exist and the device is identified purely from `TPM2_GetCapability` properties. Under `--enable-devtpm` the `DEBUG_WOLFTPM` line is still printed but reads all zeros; under `--enable-autodetect` `wolfTPM2_Init_ex` returns as soon as the kernel device opens, before that printf, so the line is absent entirely. And its algorithm coverage is set by the firmware build rather than by a datasheet, so some operations the benchmark exercises may report as unsupported; that is expected, not a fault.
+
+See the main [README.md](/README.md#device-identification) for this platform's identification values and benchmark results.
```
aidangarske: **🔵 [Low] Root-absolute link to README.md will 404 on GitHub** · _Documentation_

`[README.md](/README.md#device-identification)` is a site-root-absolute path, so on GitHub it resolves to `github.com/README.md` rather than to the repository README. Every other cross-reference added in this PR uses a relative path (e.g. `README.md:246` uses `docs/DEVTPM.md#...`, `src/fwtpm/README.md:163` uses `../../docs/FWTPM.md`).

**Fix:** Use the relative path `../README.md#device-identification`.

**Suggestion:**
```suggestion
See the main [README.md](../README.md#device-identification) for this platform's identification values and benchmark results.
```

docs/DEVTPM.md
```diff
@@ -0,0 +1,163 @@
...
+
+Note that an OP-TEE boot message about silicon-identity fTPM provisioning not being enabled refers to a separate NVIDIA feature and does **not** mean the TPM 2.0 device is unavailable.
+
+Build as above with `--enable-devtpm` or `--enable-autodetect`, then confirm with:
+
+```bash
+./examples/wrap/caps
+```
+
+Because this is a firmware TPM, expect two differences from a discrete part. There is no TIS bus, so the `TPM2: Caps/Did/Vid/Rid` values do not exist and the device is identified purely from `TPM2_GetCapability` properties. Under `--enable-devtpm` the `DEBUG_WOLFTPM` line is still printed but reads all zeros; under `--enable-autodetect` `wolfTPM2_Init_ex` returns as soon as the kernel device opens, before that printf, so the line is absent entirely. And its algorithm coverage is set by the firmware build rather than by a datasheet, so some operations the benchmark exercises may report as unsupported; that is expected, not a fault.
```
aidangarske: **🔵 [Low] DEVTPM.md and README.md disagree on Jetson benchmark algorithm coverage** · _Documentation_

DEVTPM.md tells the reader that on this platform "some operations the benchmark exercises may report as unsupported; that is expected", while README.md:963 says of the same board "every operation the benchmark exercises is supported" and the pasted benchmark output shows no skips. One of the two statements is stale.

**Fix:** Reword the DEVTPM.md sentence as a general caveat about firmware TPMs rather than about this board specifically, or drop it, so it does not contradict the measured README results.

src/tpm2.c
```diff
@@ -939,9 +939,28 @@ TPM_RC TPM2_Init_ex(TPM2_CTX* ctx, TPM2HalIoCb ioCb, void* userCtx,
     /* Set the active TPM global */
     TPM2_SetActiveCtx(ctx);
 
+#ifdef WOLFTPM_LINUX_DEV_AUTODETECT
```
aidangarske, line 942: **⚪ [Info] Native API in autodetect builds now silently prefers the kernel device over the caller-supplied HAL IO callback** · _Logic_

BEFORE: `TPM2_Init(ctx, TPM2_IoCb, userCtx)` on an autodetect build always brought up TIS/SPI via `TPM2_ChipStartup()` and used the supplied callback for all traffic.

AFTER: if `/dev/tpmrm0` or `/dev/tpm0` opens, `ctx->fd >= 0`, `TPM2_ChipStartup()` is skipped (the condition at line 960-965 gained `&& ctx->fd < 0`), and `TPM2_LINUX_AUTODETECT_SendCommand()` routes every command to the kernel device — the explicitly supplied HAL callback is never invoked.

This is the intended fix (it makes the native API match the wrapper, and `examples/native/native_test.c:503` depends on it), and it is not attacker-influenced: the device paths are compile-time constants under root-owned `/dev`, so there is no TOCTOU or symlink vector. It is flagged only because it is a silent transport switch for existing native-API applications on hosts that have both a kernel TPM and an SPI TPM — such an application will now talk to a different physical TPM than it did before the upgrade, with no diagnostic outside `DEBUG_WOLFTPM`.

**Fix:** Add a ChangeLog.md entry noting that on `--enable-autodetect` builds the native `TPM2_Init`/`TPM2_Init_ex` now prefer the kernel device over a supplied HAL IO callback, and point users who need the SPI part specifically at `--enable-spi`/`--enable-\<vendor>` or `-DWOLFTPM_USE_TPMRM`-style pinning. docs/DEVTPM.md already documents the wrapper behavior; extend that section to cover the native API.

aidangarske (changes requested):
## Skoll Multi-Scan Review

**Findings:** 16 total — 14 posted, 2 skipped
**13 finding(s) posted as inline comments** (see file-level comments below)
**1 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [High] [review+review-security] **New docs/DEVTPM.md contradicts the return value this PR gives wolfTPM2_Reset/Shutdown** — `docs/DEVTPM.md:28`
- [Medium] [review-security] **EACCES on /dev/tpm* now aborts native init in autodetect builds, defeating the SPI fallback** — `src/tpm2.c:946-949`
- [Medium] [review+review-security] **wolfTPM2_Reset/wolfTPM2_Shutdown now return NOT_COMPILED_IN, breaking callers and the degenerate (0,0) call** — `src/tpm2_wrap.c:8502`
- [Medium] [review] **TryOpen still misattributes EACCES: permission failure on /dev/tpmrm0 is reported as "device absent"** — `src/tpm2_linux.c:245-259`
- [Medium] [review] **TPM_MFG_MSFT widens the "real TPM, skip" gate to cover the ms-tpm-20-ref simulator** — `tests/unit_tests.c:429`
- [Low] [review-security] **TPM2_Init_minimal() and NULL-callback native init now fail in autodetect builds with no kernel device** — `src/tpm2.c:954-956`
- [Low] [review-security] **TPM2_Init_ex now acquires a file descriptor, so native callers that skip TPM2_Cleanup leak it (and can hold /dev/tpm0 exclusively)** — `src/tpm2.c:946`
- [Low] [review+review-security] **New devtpm-autodetect CI entry is build-only and never compiles the DEBUG_WOLFTPM paths it added** — `.github/workflows/make-test-swtpm.yml:197`
- [Low] [review] **CI comment points "below" at an entry that is above it** — `.github/workflows/make-test-swtpm.yml:268`
- [Low] [review] **Configure notice claims fwTPM was defaulted on even when --disable-fwtpm was passed** — `configure.ac:316`
- [Low] [review] **Root-absolute link to README.md will 404 on GitHub** — `docs/DEVTPM.md:145`
- [Low] [review] **DEVTPM.md and README.md disagree on Jetson benchmark algorithm coverage** — `docs/DEVTPM.md:143`
- [Info] [review-security] **Native API in autodetect builds now silently prefers the kernel device over the caller-supplied HAL IO callback** — `src/tpm2.c:942-963`

#### Native-init probe makes the wrapper's SPI-fallback check dead and degrades its error message
**File:** `src/tpm2_wrap.c:109-115`
**Function:** `wolfTPM2_Init_ex`
**Severity:** Low
**Category:** Logic

`TPM2_Init_ex` now returns `TPM_RC_FAILURE` for exactly the condition phase 3 checks (`src/tpm2.c:954`: no kernel device and `ctx->ioCb == NULL`), so `wolfTPM2_Init_ex`'s phase-3 `if (ioCb == NULL)` and its `"Kernel driver not available and no IO callback for SPI"` message are unreachable. In their place the caller now gets phase 1's unconditional `printf("TPM2_Init failed 0x%x: %s\n", ...)` = `TPM2_Init failed 0x101: TPM_RC_FAILURE`, which says nothing about the missing IO callback. Relatedly, phase 2's `TPM2_LINUX_TryOpen` call is now redundant when a device was found (it early-returns) and re-attempts both `open()` calls when one was not, so the no-TPM path probes `/dev/tpmrm0` and `/dev/tpm0` twice per init.

**Recommendation:** Either drop the now-dead phase-2/phase-3 duplication in `wolfTPM2_Init_ex` and let `TPM2_Init_ex` own the probe, or move the "no kernel device and no IO callback" diagnostic into `TPM2_Init_ex` (under `DEBUG_WOLFTPM`) so the useful message survives for both APIs.

_Referenced code: `src/tpm2_wrap.c:109-115` (7 lines)_

---

### Skipped findings
- [Low] ~~README advice to pass a NULL IO callback now sits under the autodetect instructions, where NULL fails init~~
- [Info] ~~TPM2_LINUX_TryOpen sits outside the __linux__ guard but now depends on macros defined inside it~~

aidangarske (approved):

PR #583, opened 2026-08-25
Fix ST33 generation 1 manifest size and refuse oversized commands

dgarske:
ST33 manifest (blob0) sizing was wrong for generation 1 parts. The manifest is a 33 byte fixed header plus the firmware digest and the signature over it, so its size follows the algorithms of that generation: 321 bytes on gen 1 (SHA-256 + RSAPSS-2048), 177 on gen 9 below 512 (SHA-384 + ECDSA P-384), 2697 for gen 9 LMS. Only 177 and 2697 were known, so a gen 1 image was split 144 bytes short and `FieldUpgradeStart` was fed the wrong manifest. Also fixes the LMS gate, which tested `fwVerMinor >= 512` without the major version, so a part running 1.771 demanded an LMS manifest and rejected every valid gen 1 image.

The example now confirms its choice against the file: everything after blob0 is a chain of `[type][length]` records ending exactly at end of file, so only the correct manifest size lands on the final byte. That also replaces the guess used when resuming in firmware upgrade mode, where TPM capabilities are unavailable.

Separately, a command too large for the packet buffer was going out silently truncated. The marshalling helpers drop what does not fit and only raise `packet->overflow`, which was set in seven places and read in none, so `TPM2_Packet_Finalize` stamped the short length and the TPM got a command missing its parameters. `--enable-smallstack` drops the buffer to 1350 bytes, below the 2697 byte ST33 LMS manifest, so this is reachable on a supported build. `TPM2_DispatchCommand` now returns `BUFFER_E` instead.

Fixes ZD 22193

### DIFF
--- a/examples/firmware/README.md
+++ b/examples/firmware/README.md
@@ -2,7 +2,7 @@
 
 Currently wolfTPM supports firmware update capability for:
 - Infineon SLB9672 (SPI) and SLB9673 (I2C) TPM 2.0 modules. Infineon has open sourced their firmware update.
-- STMicroelectronics ST33KTPM TPM 2.0 modules. Support includes both Generation 1 firmware versions (< 512, without LMS signature) and Generation 2 firmware versions (>= 512, with LMS signature requirement).
+- STMicroelectronics ST33KTPM TPM 2.0 modules. Support covers generation 1 firmware (RSA signed manifest), generation 9 firmware below 512 (ECDSA signed manifest) and generation 9 firmware at 512 and above (LMS signature requirement).
 
 ## Infineon Firmware
 
@@ -117,15 +117,19 @@ KeyGroupId 0x7, FwCounter 1253 (254 same)
 
 ST33KTPM firmware update automatically detects the required format based on TPM firmware version:
 
-- **Legacy firmware (< 512, e.g., 9.257)**: Non-LMS format
-  - Manifest size: 177 bytes
-  - Generation 1 firmware (ECC-only)
+The manifest (blob0) is a 33 byte fixed header followed by the firmware digest and the signature over it, so its size follows the algorithms that generation signs with:
 
-- **Modern firmware (>= 512, e.g., 9.512)**: LMS format
+- **Generation 1 (major version 1, e.g., 1.257 or 1.771)**: Non-LMS format
+  - Manifest size: 321 bytes (SHA-256 digest, RSAPSS-2048 signature)
+  - Always non-LMS, no matter how high the minor version goes
+
+- **Generation 9 below 512 (e.g., 9.257)**: Non-LMS format
+  - Manifest size: 177 bytes (SHA-384 digest, ECDSA P-384 signature)
+
+- **Generation 9 at 512 and above (e.g., 9.512)**: LMS format
   - Manifest size: 2697 bytes (includes embedded LMS signature)
-  - Generation 2 firmware (LMS mandatory)
 
-The firmware version is automatically detected from `fwVerMinor` in TPM capabilities. The correct manifest size is determined automatically - no manual format selection is needed.
+The LMS requirement is a generation 9 rule, so both `fwVerMajor` and `fwVerMinor` from TPM capabilities are consulted. The example confirms its choice against the file itself: everything after blob0 is a chain of `[type][length]` records that ends exactly at end of file, and only the correct manifest size lands on the final byte. No manual format selection is needed.
 
 ### Updating the firmware
 
@@ -145,9 +149,10 @@ Policy options (caller-supplied authorization):
 	--policyor  provision+satisfy a PolicyOR (multi-branch)
 	--sha256|--sha384|--sha512  policy hash (default SHA-256)
 
-Firmware format is auto-detected from TPM firmware version:
-      - Firmware < 512: Non-LMS format (177 byte manifest)
-      - Firmware >= 512: LMS format (2697 byte manifest with embedded signature)
+Firmware format is auto-detected from TPM firmware version and the file:
+      - Generation 1 (e.g. 1.771): Non-LMS format (321 byte manifest)
+      - Generation 9 below 512: Non-LMS format (177 byte manifest)
+      - Generation 9 at 512 and above: LMS format (2697 byte manifest)
 
 # Run without arguments to display the current firmware information
 ./st33_fw_update
@@ -156,8 +161,8 @@ TPM2: Caps 0x30000415, Did 0x0003, Vid 0x104a, Rid 0x 1
 TPM2_Startup pass
 Mfg STM (2), Vendor ST33KTPM2X, Fw 9.257 (0x0)
 Firmware version details: Major=9, Minor=257, Vendor=0x0
-Hardware: ST33K (legacy firmware, Generation 1)
-Firmware update: Non-LMS format required
+Hardware: ST33K (generation 9 firmware below 512)
+Firmware update: Non-LMS format required (177 byte manifest)
 
 # Run with firmware file (format auto-detected from TPM version)
 ./st33_fw_update TPM_ST33KTPM2X_00090200_V1.fi
@@ -167,9 +172,9 @@ TPM2: Caps 0x30000415, Did 0x0003, Vid 0x104a, Rid 0x 1
 TPM2_Startup pass
 Mfg STM (2), Vendor ST33KTPM2X, Fw 9.257 (0x0)
 Firmware version details: Major=9, Minor=257, Vendor=0x0
-Hardware: ST33K (legacy firmware, Generation 1)
-Firmware update: Non-LMS format required
-	Format: Non-LMS (from TPM firmware version)
+Hardware: ST33K (generation 9 firmware below 512)
+Firmware update: Non-LMS format required (177 byte manifest)
+	Format: Non-LMS (blob0 177 bytes, verified against the block chain)
 Firmware Update:
 	Total file size: 364290 bytes
 	Manifest (blob0): 177 bytes
@@ -178,7 +183,7 @@ Firmware Update:
 Firmware update completed successfully.
 Please reset or power cycle the TPM.
 
-# Example with LMS firmware (Generation 2 TPM, firmware >= 512)
+# Example with LMS firmware (generation 9 TPM, firmware at 512 and above)
 ./st33_fw_update ST33KTPM2X_FAC_00090200_V2.fi
 ST33 Firmware Update Tool
 	Firmware File: ST33KTPM2X_FAC_00090200_V2.fi
@@ -186,9 +191,9 @@ TPM2: Caps 0x30000415, Did 0x0003, Vid 0x104a, Rid 0x 3
 TPM2_Startup pass
 Mfg STM (2), Vendor ST33KTPM2X, Fw 9.512 (0x0)
 Firmware version details: Major=9, Minor=512, Vendor=0x0
-Hardware: ST33K (modern firmware, Generation 2)
-Firmware update: LMS format required
-	Format: LMS (from TPM firmware version)
+Hardware: ST33K (generation 9 firmware at 512 and above)
+Firmware update: LMS format required (2697 byte manifest)
+	Format: LMS (blob0 2697 bytes, verified against the block chain)
 Firmware Update:
 	Total file size: 360092 bytes
 	Manifest (blob0): 2697 bytes
@@ -204,8 +209,8 @@ TPM2: Caps 0x30000415, Did 0x0003, Vid 0x104a, Rid 0x 1
 TPM2_Startup pass
 Mfg STM (2), Vendor ST33KTPM2X, Fw 9.257 (0x0)
 Firmware version details: Major=9, Minor=257, Vendor=0x0
-Hardware: ST33K (legacy firmware, Generation 1)
-Firmware update: Non-LMS format required
+Hardware: ST33K (generation 9 firmware below 512)
+Firmware update: Non-LMS format required (177 byte manifest)
 Firmware Update Abandon:
 Success: Please reset or power cycle TPM
 ```
--- a/examples/firmware/st33_fw_update.c
+++ b/examples/firmware/st33_fw_update.c
@@ -35,6 +35,7 @@
     !defined(WOLFTPM2_NO_WOLFCRYPT)
 
 #include <examples/firmware/firmware_policy.h>
+#include <examples/firmware/st33_blob0.h>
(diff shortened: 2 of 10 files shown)


examples/firmware/st33_fw_update.c
```diff
@@ -51,8 +51,67 @@
...
+    return 0;
+}
+
+/* Manifest size the running firmware expects for its next update */
+static size_t st33_expected_blob0(WOLFTPM2_CAPS* caps)
+{
+    if (caps->fwVerMajor < 9) {
+        return ST33_BLOB0_SIZE_NON_LMS_RSA;
+    }
+    if (caps->fwVerMinor < 512) {
```
aidangarske: **TPM2_ST33_PrintInfo still uses the broken fwVerMinor-only LMS gate this PR fixes everywhere else**

The PR's stated defect is that `fwVerMinor >= 512` was tested without the major version, so a generation 1 part at 1.771 was misclassified as LMS. The fix was applied in `src/tpm2_wrap.c` (`tpm2_st33_expected_manifest_sz` now checks `fwVerMajor < ST33_FW_GENERATION_LMS_CAPABLE` first), in `examples/firmware/st33_fw_update.c` (`st33_expected_blob0`), and in `tests/unit_tests.c:308` (updated to `caps.fwVerMajor >= 9 && caps.fwVerMinor >= 512`). `TPM2_ST33_PrintInfo` in the same example file was missed and still branches on the bare `caps->fwVerMinor < 512`. It is called at line 407, immediately before the manifest is selected, and again at line 523 after the update. On the exact part this PR exists to support — firmware 1.771, `fwVerMajor` 1, `fwVerMinor` 771 — it prints `Hardware: ST33K (modern firmware, Generation 2)` and `Firmware update: LMS format required`, and then the tool correctly proceeds to pick the 321 byte non-LMS manifest. The tool contradicts itself on screen and tells the operator to go find an LMS `.fi` file that will be rejected. The `Generation 1`/`Generation 2` wording is also now inconsistent with the terminology the PR introduces, where the generation is `fwVerMajor` (1 or 9): the README sample output at line 160 shows `Fw 9.257` labelled `Generation 1`, while the README's own new section two headings above defines 9.257 as "Generation 9 below 512".

**Fix:** Apply the same major+minor gate used by `st33_expected_blob0` and `tpm2_st33_expected_manifest_sz`, and refresh the sample `./st33_fw_update` output in examples/firmware/README.md (around line 160) so the `Fw 9.257` transcript no longer says `Generation 1`.

**Suggestion:**
```suggestion
if (caps->fwVerMajor < 9) {
        printf("Hardware: ST33K (generation 1 firmware)\n");
        printf("Firmware update: Non-LMS format required (321 byte manifest)\n");
    }
    else if (caps->fwVerMinor < 512) {
        printf("Hardware: ST33K (generation 9 firmware below 512)\n");
        printf("Firmware update: Non-LMS format required (177 byte manifest)\n");
    }
    else {
        printf("Hardware: ST33K (generation 9 firmware at 512 and above)\n");
        printf("Firmware update: LMS format required (2697 byte manifest)\n");
    }
```

src/tpm2.c
```diff
@@ -521,6 +521,19 @@ static TPM_RC TPM2_DispatchCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
 {
     TPM_RC rc;
 
+    /* A command that did not fit the packet buffer would otherwise go out
+     * silently truncated, since the marshalling helpers drop what does not
+     * fit and TPM2_Packet_Finalize then stamps the short length. The buffer
+     * is MAX_COMMAND_SIZE / MAX_RESPONSE_SIZE, which --enable-smallstack
+     * lowers to 1024 / 1350, so this is reachable on a normal build. */
+    if (packet->overflow) {
```
aidangarske, line 530: **New TPM2_DispatchCommand overflow guard has no test coverage**

The only behavioral change in `src/tpm2.c` is the early `return (TPM_RC)BUFFER_E` when `packet->overflow` is set. The new `test_TPM2_Packet_overflow` (tests/unit_tests.c:4135) is named after it but exercises only `TPM2_Packet_InitBuf`, `TPM2_Packet_AppendBytes`, `TPM2_Packet_AppendU32` and `TPM2_Packet_Finalize` — all of which are unchanged by this PR. It asserts that the flag gets raised and that `Finalize` stamps a short length, but nothing asserts that the send path now refuses to transmit. If someone deletes the six new lines in `TPM2_DispatchCommand`, the entire unit test suite still passes. Since `TPM2_DispatchCommand` and `TPM2_TransmitCommand` are `static`, the reachable seam is a public command builder: `TPM2_ST33_FieldUpgradeCommand(TPM_CC_FieldUpgradeDataVendor_ST33, data, size)` (src/tpm2.c:6700) does `TPM2_Packet_Init` + `TPM2_Packet_AppendBytes` + `TPM2_SendCommand`, so calling it with `size > XFER_MAX_SIZE - TPM2_HEADER_SIZE` should now return `BUFFER_E` without touching the transport.

**Fix:** Add a case that drives an oversized command through a public entry point (e.g. `TPM2_ST33_FieldUpgradeCommand` with a buffer larger than `XFER_MAX_SIZE`) and asserts `BUFFER_E` is returned and no bytes reached the transport callback. Renaming the existing test to something like `test_TPM2_Packet_overflow_flag` would also stop it reading as coverage for the dispatch guard.

examples/firmware/st33_fw_update.c
```diff
@@ -372,50 +434,39 @@ int TPM2_ST33_Firmware_Update(void* userCtx, int argc, char *argv[])
...
-        printf("\tFormat: %s (from TPM firmware version)\n",
-            (caps.fwVerMinor >= 512) ? "LMS" : "Non-LMS");
-
-        /* Load the complete .fi file */
-        rc = loadFile(fi_file, &fwinfo.fi_buf, &fwinfo.fi_bufSz);
-        if (rc != 0) {
-            printf("Failed to load firmware file: %s\n", fi_file);
-            goto exit;
+
+    candCnt = 0;
```
aidangarske: **Example proceeds with a manifest size the running firmware will reject, after printing that it was "verified"**

`st33_expected_blob0(&caps)` is pushed as the preferred candidate, but if the file does not parse at that size the loop falls through to the other two sizes and `st33_detect_blob0` returns whichever one happens to land on EOF. That is exactly the "user has the wrong .fi for this part" case — a generation 9 image on a generation 1 TPM, say — and the tool responds by printing `Format: Non-LMS (blob0 177 bytes, verified against the block chain)` and continuing. `wolfTPM2_FirmwareUpgrade_ex` then hits the new `manifest_sz != expected_sz` check in `tpm2_st33_firmware_upgrade_hash` (src/tpm2_wrap.c:12218) and returns `BAD_FUNC_ARG`; the explanatory message there is inside `#ifdef DEBUG_WOLFTPM`, so on a release build the operator sees only `ST33 firmware update failed 0x... ` with no hint that the file is for the wrong generation. The example already has both numbers in hand at line 456 and could say so directly.

**Fix:** When `!fwinfo.in_upgrade_mode && blob0_size != cand[0]`, print an explicit mismatch line naming both sizes and the running `caps.fwVerMajor`.`caps.fwVerMinor`, and fail with `BAD_FUNC_ARG` in the example rather than letting the library reject it with a debug-only diagnostic.

examples/firmware/README.md
```diff
@@ -145,9 +149,10 @@ Policy options (caller-supplied authorization):
 	--policyor  provision+satisfy a PolicyOR (multi-branch)
 	--sha256|--sha384|--sha512  policy hash (default SHA-256)
 
-Firmware format is auto-detected from TPM firmware version:
-      - Firmware < 512: Non-LMS format (177 byte manifest)
-      - Firmware >= 512: LMS format (2697 byte manifest with embedded signature)
+Firmware format is auto-detected from TPM firmware version and the file:
```
aidangarske, line 152: **README --help transcript documents output usage() does not print**

The README block is presented as literal `./st33_fw_update --help` output and the PR rewrote its trailing lines to the new three-way format table (`Generation 1 (e.g. 1.771): Non-LMS format (321 byte manifest)` etc). `usage()` in examples/firmware/st33_fw_update.c:128-129 was not touched and still prints the old two lines: `Firmware format is auto-detected from the TPM firmware version.` / `Just provide the correct .fi file for your TPM and it will be handled automatically.` A reader copying the documented invocation gets different text than the README shows, and the README is now the only place the 321/177/2697 breakdown appears at the CLI level.

**Fix:** Update `usage()` in examples/firmware/st33_fw_update.c:128-129 to print the same four lines the README now shows, so the transcript stays accurate.

examples/firmware/st33_fw_update.c
```diff
@@ -51,8 +51,67 @@
...
+#define ST33_BLOB0_SIZE_CNT \
+    (sizeof(st33_blob0_sizes) / sizeof(st33_blob0_sizes[0]))
+
+/* Confirm a candidate blob0 size by walking the block chain that follows it.
+ * Every byte after blob0 is a [type:1][len:2 big-endian][payload] record and
+ * the chain ends exactly at end of file, so only the correct size lands on
+ * the final byte. Candidates are tried in the supplied order, so the size the
+ * TPM firmware version implies wins when more than one could fit. Returns the
+ * blob0 size, or 0 when the file does not parse with any known size. */
+static size_t st33_detect_blob0(const byte* buf, size_t bufSz,
```
aidangarske: **New st33_detect_blob0 block-chain parser has no test coverage**

`st33_detect_blob0` is ~25 lines of new parsing logic — a candidate loop, a record walk with `off += 3 + len`, and three distinct early-exit conditions (`buf[off] == 0`, `len == 0`, `off + 3 > bufSz`) — and it is what decides where the manifest ends and firmware data begins. It replaces the previous heuristic entirely, including the upgrade-mode path where the TPM cannot be consulted at all, so a false positive here splits the image at the wrong byte. It is `static` inside an example and has no tests. The interesting cases are all cheap to cover against synthetic buffers: exact-fit chain at each of 321/177/2697, a chain overshooting EOF by 1-2 bytes, a zero-length record mid-chain, `bufSz == cand[i]`, and an image that parses at two candidate sizes (which must resolve to the caps-preferred one).

**Fix:** Move the detector into a testable location (or expose a thin non-static wrapper) and add unit tests for the boundary cases above, especially the ambiguous-file case where two candidate sizes both parse.

examples/firmware/st33_fw_update.c
```diff
@@ -51,8 +51,67 @@
...
+        }
+        if (off == bufSz) {
+            return cand[i];
+        }
+    }
+    return 0;
+}
+
+/* Manifest size the running firmware expects for its next update */
+static size_t st33_expected_blob0(WOLFTPM2_CAPS* caps)
```
aidangarske: **Example duplicates the generation/LMS thresholds as bare literals and takes a non-const caps pointer**

`src/tpm2_wrap.c` names both thresholds (`ST33_FW_GENERATION_LMS_CAPABLE 9`, `ST33_FW_VERSION_LMS_REQUIRED 512`) and its twin helper takes `const WOLFTPM2_CAPS* caps`. The example's copy hardcodes `9` and `512` and takes a non-const `WOLFTPM2_CAPS*` even though it only reads. Two copies of the same rule with different spellings is how the original `fwVerMinor >= 512` bug survived in three places; the const mismatch also diverges from the wolfSSL C style rule that unmodified pointer params take `const`.

**Fix:** Add `const`, and give the two thresholds named defines in the example mirroring `ST33_FW_GENERATION_LMS_CAPABLE` / `ST33_FW_VERSION_LMS_REQUIRED` so the rule reads the same in both files.

**Suggestion:**
```suggestion
static size_t st33_expected_blob0(const WOLFTPM2_CAPS* caps)
```

src/tpm2.c
```diff
@@ -521,6 +521,19 @@ static TPM_RC TPM2_DispatchCommand(TPM2_CTX* ctx, TPM2_Packet* packet)
...
+    #ifdef DEBUG_WOLFTPM
+        printf("Command exceeds the %d byte packet buffer "
```
aidangarske: **Overflow debug message labels the buffer MAX_COMMAND_SIZE when it is XFER_MAX_SIZE**

`packet->size` here comes from `TPM2_Packet_Init` -> `TPM2_Packet_InitBuf(packet, ctx->cmdBuf, sizeof(ctx->cmdBuf))`, and `cmdBuf` is declared `byte cmdBuf[XFER_MAX_SIZE]` (wolftpm/tpm2.h:2227), where `XFER_MAX_SIZE` is `max(MAX_COMMAND_SIZE, MAX_RESPONSE_SIZE)`. Under `--enable-smallstack` — the configuration the adjacent comment cites — configure.ac:624 sets `MAX_COMMAND_SIZE=1024` and `MAX_RESPONSE_SIZE=1350`, so this prints `Command exceeds the 1350 byte packet buffer (MAX_COMMAND_SIZE)`. The number and the macro name disagree in exactly the build the message is meant to help diagnose.

**Fix:** Drop the `(MAX_COMMAND_SIZE)` label or change it to `XFER_MAX_SIZE`; the same correction applies to the comment above the check and to the comment on tests/unit_tests.c:4128-4134.

**Suggestion:**
```suggestion
printf("Command exceeds the %d byte packet buffer\n", packet->size);
```

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

**Findings:** 7 total — 7 posted, 0 skipped
**7 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **TPM2_ST33_PrintInfo still uses the broken fwVerMinor-only LMS gate this PR fixes everywhere else** — `examples/firmware/st33_fw_update.c:243-250`
- [Medium] **New TPM2_DispatchCommand overflow guard has no test coverage** — `src/tpm2.c:529-535`
- [Medium] **Example proceeds with a manifest size the running firmware will reject, after printing that it was "verified"** — `examples/firmware/st33_fw_update.c:446-468`
- [Medium] **README --help transcript documents output usage() does not print** — `examples/firmware/README.md:152-155`
- [Medium] **New st33_detect_blob0 block-chain parser has no test coverage** — `examples/firmware/st33_fw_update.c:77-101`
- [Low] **Example duplicates the generation/LMS thresholds as bare literals and takes a non-const caps pointer** — `examples/firmware/st33_fw_update.c:105-113`
- [Low] **Overflow debug message labels the buffer MAX_COMMAND_SIZE when it is XFER_MAX_SIZE** — `src/tpm2.c:531-532`

aidangarske (approved):

PR #586, opened 2026-08-27
Select ST33 field upgrade commands from the TPM command set

dgarske:
## Summary

ST33 supports two different field-upgrade command code pairs, and wolfTPM always sent the ST33KTPM vendor pair (`0x2000030C`/`0x2000030D`) - which some ST33 parts do not implement, so the upgrade failed at the very first command with `TPM_RC_COMMAND_CODE`. wolfTPM now asks the TPM which pair it implements and uses that, falling back to the standard codes (`0x0000012F`/`0x00000141`) for generation-2 images, and otherwise leaving the previous behavior unchanged.

ZD 22193

### DIFF
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -11974,6 +11974,8 @@ static int tpm2_st33_family(word16 fwVerMajor)
         case 9:  /* ST33KTPM2X */
         case 10: /* ST33KTPM2A */
             return ST33_FW_FAMILY_KTPM;
+        /* Uncharacterized majors stay unknown so no size or family rule is
+         * asserted against them, even ones the tool can name */
         default:
             return ST33_FW_FAMILY_UNKNOWN;
     }
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -886,7 +886,10 @@ static int wolfTPM2_ParseCapabilities(WOLFTPM2_CAPS* caps,
             case TPM_PT_VENDOR_STRING_3:
             case TPM_PT_VENDOR_STRING_4:
                 val = TPM2_Packet_SwapU32(val); /* swap for little endian */
-                len = (word32)XSTRLEN(caps->vendorStr); /* add to existing string */
+                /* Offset by property, not string length: a chunk starting
+                 * with a zero byte would let the next one overwrite it */
+                len = (word32)(props->tpmProperty[i].property -
+                    TPM_PT_VENDOR_STRING_1) * (word32)sizeof(UINT32);
                 if (len + sizeof(UINT32) < sizeof(caps->vendorStr)) {
                     XMEMCPY(&caps->vendorStr[len], &val, sizeof(UINT32));
                 }
--- a/src/tpm2_wrap.c
+++ b/src/tpm2_wrap.c
@@ -837,7 +837,7 @@ int wolfTPM2_SelfTest(WOLFTPM2_DEV* dev)
  *      comply with all of the FIPS 140-2 requirements at Level 1 or higher.
  *   TPM_PT_FIRMWARE_VERSION_2: ST Internal Additional Version
  */
-static int wolfTPM2_ParseCapabilities(WOLFTPM2_CAPS* caps,
+int wolfTPM2_ParseCapabilities(WOLFTPM2_CAPS* caps,
     TPML_TAGGED_TPM_PROPERTY* props)
 {
     int rc = 0;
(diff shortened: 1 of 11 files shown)


src/tpm2_wrap.c
```diff
@@ -11966,16 +11991,181 @@ int wolfTPM2_FirmwareUpgradeCancel(WOLFTPM2_DEV* dev)
...
+    int rcVendor, rcStd;
+
+    if (ccStart == NULL || ccData == NULL) {
+        return BAD_FUNC_ARG;
+    }
+    if (fromTpm != NULL) {
+        *fromTpm = 0;
+    }
+
+    rcVendor = wolfTPM2_ST33_CmdImplemented(TPM_CC_FieldUpgradeStartVendor_ST33,
```
aidangarske, line 12145: **Capability probes abort the firmware-upgrade resume path**

`caps == NULL` is explicitly documented and used by `st33_fw_update.c:583-584` to mean the TPM is already in firmware-upgrade mode, but the function still issues two `TPM2_GetCapability` commands before falling back to the image-version rule. ST documents that only `FieldUpgradeData` is accepted in this mode; another command returns `TPM_RC_UPGRADE` and exits the mode ([ST33KTPM-IDevID datasheet](https://www.st.com/resource/en/datasheet/st33ktpm-idevid.pdf), section 10.5.3). Consequently, attempting to resume can leave upgrade mode before the first firmware blob is sent. Before this PR, the resume branch proceeded directly to `TPM2_ST33_FieldUpgradeCommand`; these probes introduce the regression.

**Fix:** When `caps` is NULL, skip all TPM capability I/O and immediately apply the version fallback. Add a mocked-transport test proving that this path sends no `GetCapability` command before firmware data.

**Suggestion:**
```suggestion
if (caps == NULL) {
    return wolfTPM2_ST33_FwUpgradeCommands(0, 0, manifestMajor, ccStart,
        ccData);
}
rcVendor = wolfTPM2_ST33_CmdImplemented(
    TPM_CC_FieldUpgradeStartVendor_ST33, &vendorImpl);
```

src/tpm2_wrap.c
```diff
@@ -11966,16 +11991,181 @@ int wolfTPM2_FirmwareUpgradeCancel(WOLFTPM2_DEV* dev)
...
+        &vendorImpl);
+    rcStd = wolfTPM2_ST33_CmdImplemented(TPM_CC_FieldUpgradeStart, &stdImpl);
+
+    if (rcVendor == TPM_RC_SUCCESS && rcStd == TPM_RC_SUCCESS &&
```
aidangarske, line 12149: **A definitive unsupported command set falls through to guessing**

When both capability queries succeed and both `vendorImpl` and `stdImpl` are zero, the TPM has authoritatively reported that neither upgrade pair is implemented. Because `vendorImpl != stdImpl` is false, the function nevertheless falls through to the version heuristic and returns a command pair known to be unsupported. Fallback is appropriate for query failures or the both-supported ambiguity, but not for a successful neither-supported result.

**Fix:** Return an unsupported-command error when both successful probes report false. Cover none, both, each individual pair, and query-error cases with deterministic capability responses.

**Suggestion:**
```suggestion
if (rcVendor == TPM_RC_SUCCESS && rcStd == TPM_RC_SUCCESS &&
        !vendorImpl && !stdImpl) {
    return TPM_RC_COMMAND_CODE;
}
if (rcVendor == TPM_RC_SUCCESS && rcStd == TPM_RC_SUCCESS &&
        vendorImpl != stdImpl) {
```

src/tpm2_wrap.c
```diff
@@ -886,7 +886,10 @@ static int wolfTPM2_ParseCapabilities(WOLFTPM2_CAPS* caps,
             case TPM_PT_VENDOR_STRING_3:
             case TPM_PT_VENDOR_STRING_4:
                 val = TPM2_Packet_SwapU32(val); /* swap for little endian */
-                len = (word32)XSTRLEN(caps->vendorStr); /* add to existing string */
+                /* Offset by property, not string length: a chunk starting
```
aidangarske, line 889: **Add a regression test for binary vendor-string chunk placement**

The PR fixes accumulation specifically for vendor-string chunks containing leading NUL bytes, but no test supplies that input and checks the resulting byte offsets. `test_wolfTPM2_GetCapabilities` only verifies that the live query succeeds and optionally prints the string, so it cannot detect a return to `XSTRLEN`-based placement or another embedded-NUL regression.

**Fix:** Feed a synthetic capability response whose first vendor chunk starts with zero and whose later chunks are nonzero, then verify offsets 0, 4, 8, and 12 plus the final terminator.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

**Findings:** 3 total — 3 posted, 0 skipped

### Posted findings
- [High] **Capability probes abort the firmware-upgrade resume path** — `src/tpm2_wrap.c:12137-12139`
- [Medium] **A definitive unsupported command set falls through to guessing** — `src/tpm2_wrap.c:12141-12168`
- [Medium] **Add a regression test for binary vendor-string chunk placement** — `src/tpm2_wrap.c:889-895`

aidangarske (approved):

PR #588, opened 2026-09-01
Harden fwTPM protocol handling and SPDM authentication

aidangarske:
```
F-12708, F-12709, F-12710, F-12735, F-12736, F-12738
```


src/fwtpm/fwtpm_main.c
```diff
@@ -393,6 +393,13 @@ int main(int argc, char* argv[])
                 rc = wolfSPDM_RespSetIdentityKey(ctx.spdmRespCtx,
                     idPriv, sizeof(idPriv), idPub, sizeof(idPub));
             }
+            if (rc == 0 && useTcg) {
```
dgarske, line 396: Should this be wrapped in a DEBUG_SPDM option? Have we built and tested SPDM with fTPM on an embedded device like the STM32H5 and looked at code size? Perhaps you could take a pass at making sure the SPDM feature is code size reduced (or have options to code size reduce)? I am okay with that being a followup PR.

src/fwtpm/fwtpm_tis.c
```diff
@@ -68,6 +68,15 @@ static UINT32 TisBuildSts(BYTE stsFlags, UINT16 burstCount)
     return ((UINT32)burstCount << 8) | (UINT32)stsFlags;
 }
 
+static void TisPublishMagic(FWTPM_TIS_REGS* regs)
+{
+#if defined(__GNUC__) || defined(__clang__)
+    __atomic_store_n(&regs->magic, FWTPM_TIS_MAGIC, __ATOMIC_RELEASE);
```
dgarske: Let's setup some atomic macros leveraging what we have in wolfCrypt and for non-wolfCrypt builds duplicate. See wc_port.h WOLFSSL_NO_ATOMICS.

src/fwtpm/fwtpm_tis.c
```diff
@@ -451,6 +459,9 @@ int FWTPM_TIS_Init(FWTPM_CTX* ctx)
     regs->did_vid = FWTPM_TIS_DID_VID_VAL;
     regs->rid = FWTPM_VERSION_PATCH;
 
+    /* Release-publish the validity sentinel after all client-visible state. */
+    TisPublishMagic(regs);
```
dgarske: Then let's not use a function for this, just use the atomics helpers.

src/fwtpm/fwtpm_tis_shm.c
```diff
@@ -103,20 +113,39 @@ static int TisShmInit(void* ctx, FWTPM_TIS_REGS** regs)
...
+    if (fstat(fd, &shmStat) != 0 ||
+            shmStat.st_size != (off_t)sizeof(FWTPM_TIS_REGS) ||
+            (shmStat.st_mode & (S_IRWXU | S_IRWXG | S_IRWXO)) !=
+                (S_IRUSR | S_IWUSR)) {
+        fprintf(stderr, "fwTPM TIS: invalid shm endpoint metadata\n");
+        close(fd);
+        (void)unlink(FWTPM_TIS_SHM_PATH);
+        return -1;
+    }
+    if (FWTPM_TIS_MakeSemNames((UINT64)shmStat.st_uid,
```
dgarske: Let's not create a new fwtpm_tis_sem.c for this. Just keep as local static functions. If we need FWTPM_TIS_MakeSemNames for other files then make WOLFTPM_LOCAL and keep in this file please.

wolftpm/fwtpm/fwtpm_tis.h
```diff
@@ -47,9 +48,13 @@
...
 #endif
 
+/* Holds a leading slash, the 251 characters Linux permits after it, and NUL.
+ * Shorter platform limits are enforced by sem_open(). */
+#define FWTPM_TIS_SEM_NAME_SIZE 253
+
 /* Magic and version for shared memory validation */
 #define FWTPM_TIS_MAGIC         0x57544953UL  /* "WTIS" */
-#define FWTPM_TIS_VERSION       1
+#define FWTPM_TIS_VERSION       2
```
dgarske: Did we break compatibility that requires updating version?

wolftpm/tpm2_wrap.h
```diff
@@ -238,11 +245,123 @@ WOLFTPM_API int wolfTPM2_Test(TPM2HalIoCb ioCb, void* userCtx, WOLFTPM2_CAPS* ca
...
+    \param userCtx pointer to a user context (can be NULL)
+    \param rspPubKey trusted raw P-384 responder key in X||Y format
+    \param rspPubKeySz size of rspPubKey; must be 96 bytes
+
+    \sa wolfTPM2_Init
+    \sa wolfTPM2_SpdmSetResponderPubKey
+    \sa wolfTPM2_SpdmConnect
+    \sa wolfTPM2_SpdmIsConnected
+*/
+WOLFTPM_API int wolfTPM2_InitWithSpdmKey(WOLFTPM2_DEV* dev,
```
dgarske, line 290: Are all these new API's really needed? Can you explain why?

dgarske (changes requested):

src/fwtpm/fwtpm_main.c
aidangarske: no it shouldn't since its needed for the ephemeral response key to pin on the client. agreed I will do follow up and test

src/fwtpm/fwtpm_tis.c
aidangarske: Done, FWTPM_TIS_ATOMIC_STORE/LOAD in fwtpm_tis.h follow the wc_port.h WOLFSSL_NO_ATOMICS ladder

src/fwtpm/fwtpm_tis.c
aidangarske: TisPublishMagic and the client's LoadMagic are gone and the macros are used inline.

src/fwtpm/fwtpm_tis_shm.c
aidangarske: okay, fwtpm_tis_sem.c and its test are removed and the per-UID naming is dropped entirely since the shm file authentication is what secures the endpoint and semaphores only carry wakeups.

wolftpm/fwtpm/fwtpm_tis.h
aidangarske: Reverted to 1, the names were the only wire change and those are back to the fixed names

wolftpm/tpm2_wrap.h
aidangarske: not necessarily, the API's where added to give the app a way to supply the pin with the real key and bring it up against the session. 

heres what AI says they each do 

- wolfTPM2_SpdmSetResponderPubKey — installs the trusted responder key, which is the fix: it's the pin that stops Init from trusting whatever key an
  attacker's device presents.
- wolfTPM2_InitWithSpdmKey — brings up an SPDM-only (identity-key) TPM against that pinned key, which is now the only way since plain Init fails closed and its cleartext Startup can't complete on a locked TPM.
- wolfTPM2_InitWithSpdmKey_ex — the standard extended form that adds a mode arg so a dual Nuvoton+Nations build with no DID/VID can force the vendor; the base calls it with AUTO.
- wolfTPM2_InitWithSpdmPsk — the same locked-TPM bring-up for PSK mode, needed because a PSK-locked TPM has the identical problem but uses a shared key instead of a pinned public key.
- wolfTPM2_SpdmModeFromDidVid — test-only helper (WOLFTPM_TEST_API) that maps DID/VID to a vendor so the auto-selection logic can be unit-tested; it's not public in normal builds.

dgarske (approved):

PR #594, opened 2026-09-08
Add TPM2_PolicyTransportSPDM support for client and firmware TPM

aidangarske:
- Adds TPM2_PolicyTransportSPDM (CC 0x1A1) and TPM_CAP_SPDM_SESSION_INFO to the
  client command layer and the firmware TPM, per TPM 2.0 Library v1.84.
- Binds a policy session to require the command arrive over an active SPDM
  session, optionally scoped to a requester and/or TPM identity key name folded
  into the policy digest.
- Enforcement returns TPM_RC_CHANNEL outside SPDM and TPM_RC_CHANNEL_KEY on a
  bound-key mismatch.
- Adds wolfTPM2_PolicyTransportSPDM, wolfTPM2_PolicyTransportSPDMMake (offline
  authPolicy), and wolfTPM2_GetCapability_SPDMSessionInfo, all under WOLFTPM_SPDM.
- Only asymmetric SPDM sessions expose key names; vendor PSK sessions satisfy the
  secure-channel assertion but report empty names, since no asymmetric key
  authenticated them.
- Rejects cross-exchange SPDM finish messages and identity-key rotation on an
  active session so a session cannot be relabeled as identity-key authenticated.
- Fails closed on a truncated capability response and on hash errors while
  computing the policy key-name digest.
- Adds an nv_bind example plus fwTPM unit and end-to-end coverage, including a
  tpmKeyName-bound policy that succeeds over its session and is rejected on a
  name mismatch.
- fwtpm_unit, unit tests, spdm_test.sh tcg 15/15 and psk 27/27, guard
  sweep (no-policy, Nuvoton-only, NO_GETENV, fuzz+spdm), and a non-SPDM build.


dgarske:
All 30 CI failures are one root cause. The new `TPM_RC_CHANNEL` / `TPM_RC_CHANNEL_KEY` are added unconditionally after `TPM_RC_PARMS`, but master already defines both inside the existing `#ifdef WOLFTPM_V185` block just below, so every `--enable-v185` build fails:

```
./wolftpm/tpm2.h:437:5: error: redeclaration of enumerator 'TPM_RC_CHANNEL'
./wolftpm/tpm2.h:438:5: error: redeclaration of enumerator 'TPM_RC_CHANNEL_KEY'
```

Dropping the new pair and widening the existing guard keeps one definition in its correct Table 17 position and still covers SPDM-only builds:

```diff
     /* TCG Part 2 Sec.6.6.3 Table 17 -- present since v1.16, not v1.85 */
     TPM_RC_PARMS              = RC_FMT1 + 0x02A,
-    TPM_RC_CHANNEL            = RC_FMT1 + 0x030,
-    TPM_RC_CHANNEL_KEY        = RC_FMT1 + 0x031,
 #ifdef WOLFTPM_PQC
     /* v185 rc4 Part 2 Sec.6.6.3 Table 17 additions */
     TPM_RC_EXT_MU             = RC_FMT1 + 0x02B,
     TPM_RC_ONE_SHOT_SIGNATURE = RC_FMT1 + 0x02C,
     TPM_RC_SIGN_CONTEXT_KEY   = RC_FMT1 + 0x02D,
 #endif
-#ifdef WOLFTPM_V185
+#if defined(WOLFTPM_V185) || defined(WOLFTPM_SPDM)
     /* Part 2 v1.85 Sec.6.6.3 Table 17 channel protocol codes */
     TPM_RC_CHANNEL            = RC_FMT1 + 0x030,
     TPM_RC_CHANNEL_KEY        = RC_FMT1 + 0x031,
 #endif
```

Verified locally with this applied: `--enable-v185` and `--enable-fwtpm --enable-v185` build clean, `unit.test` and `fwtpm_unit.test` pass including the new PolicyTransportSPDM cases, and `spdm_test.sh` gives tcg 15/15 and psk 27/27.

dgarske (approved):

PR #596, opened 2026-09-09
Harden fwTPM key derivation and command validation

aidangarske:
```
F-12728, F-11915, F-12711, F-12737, F-13450, F-13451, F-13462, F-13537
```


src/fwtpm/fwtpm_crypto.c
```diff
@@ -192,6 +192,10 @@ int FwComputePublicName(TPMT_PUBLIC* pub, TPM2B_NAME* name)
     int digestSz;
 
     FWTPM_ALLOC_BUF(pubBuf, FWTPM_MAX_PUB_BUF);
+    if (rc != 0) {
+        FWTPM_FREE_BUF(pubBuf);
```
dgarske: This doesn't look right. Only failure here could be a malloc failure and it wouldn't need to be free'd.

dgarske (changes requested):

src/fwtpm/fwtpm_crypto.c
aidangarske: okay thanks I updated the FWTPM_FREE_BUF so its not pointless

dgarske (approved):

PR #602, opened 2026-09-09
Add official wolfTPM Rust wrapper crate

aidangarske:
### Current public types

| Type | Purpose |
| --- | --- |
| `Device` | Open TPM context (RAII; closes on drop). One live `Device` at a time. |
| `Key<'d>` | A loaded TPM key; unloads its handle on drop. Borrows the `Device`. |
| `KeyBlob<'d>` | A created-but-not-loaded key (wrapped pub/priv blob); serialize and reload. |
| `Template` | A TPM public-area key template (SRK, AIK, EK, signing, decrypt, HMAC, symmetric, ECDH). |
| `Session<'d>` | A parameter-encryption (salted) session. One at a time; RAII drops it. |
| `Secret` | Owned secret buffer that zeroizes on drop; `Deref` to `&[u8]`. |
| `TpmError` | A failed operation carrying the raw TPM/wolfCrypt return code. Implements `Error`/`Display`/`Debug`. |
| `Result<T>` | Alias for `core::result::Result<T, TpmError>`. |
| `Hierarchy` | Hierarchy selector: `Owner`, `Endorsement`, `Platform`. |
| `KeyAlg` | Key algorithm: `Rsa`, `EccP256`. |
| `HashAlg` | Hash algorithm (with `digest_size()`). |
| `NvSlot<'d>` | A defined NV index handle. |
| `Attestation` | Result of `certify`/`quote`: attest bytes, signature, signature algorithm. |
| `Caps` | TPM capabilities / vendor info. |
| `EcdhResult` | Ephemeral ECDH public point plus the derived shared `Secret`. |
| `Credential` | `MakeCredential` output (encrypted secret + blob) for activation. |

### Public API's

**`Device` — connection, RNG, self-test/caps**
- `open()` *(cfg: any transport)* — open the default TPM (autodetect transport)
- `open_swtpm()` *(cfg: swtpm)* — open against a swtpm/fwTPM socket
- `get_random(buf)` *(cfg: rng)* — fill a buffer from the TPM RNG
- `self_test()` *(cfg: caps)* — run the TPM self-test
- `capabilities()` *(cfg: caps)* — query `Caps`

**`Device` — key creation and import**
- `create_primary(hierarchy, alg, auth)` — create a primary/storage-root key under a hierarchy
- `create_ek(alg)` — create the Endorsement Key
- `create_key(parent, template, auth)` — create a child key blob (not loaded)
- `create_and_load(parent, template, auth)` — create and load a child key in one step
- `import_rsa_key(...)` *(cfg: import)* — import an external RSA private key
- `import_ecc_key(...)` *(cfg: import)* — import an external ECC private key

**`Device` — seal / unseal**
- `seal(parent, data, auth)` *(cfg: seal)* — seal data to the TPM
- `unseal(...)` *(cfg: seal)* — unseal (returns `Secret`)
- `seal_pcr(parent, data, pcr_indices)` *(cfg: seal)* — PCR-policy-bound seal
- `unseal_pcr(...)` *(cfg: seal)* — unseal a PCR-bound blob (returns `Secret`)

**`Device` — NV storage**
- `nv_create(index, size, auth)` *(cfg: nv)* — define an NV index
- `nv_write(slot, data, offset)` *(cfg: nv)*
- `nv_read(slot, buf, offset)` *(cfg: nv)*
- `nv_delete(index)` *(cfg: nv)*
- `read_cert(nv_handle)` *(cfg: nvcert)* — read an EK certificate from NV

**`Device` — PCRs**
- `pcr_read(index, hash)` *(cfg: pcr)*
- `pcr_extend(index, hash, digest)` *(cfg: pcr)*

**`Device` — attestation**
- `certify(...)` — an AIK certifies another object (returns `Attestation`)
- `quote(...)` — signed PCR quote (returns `Attestation`)
- `make_credential(...)` — `TPM2_MakeCredential` (returns `Credential`)
- `activate_credential(...)` *(cfg: ek_policy)* — `TPM2_ActivateCredential` (returns `Secret`)

**`Device` — persistent handles**
- `persist_key(...)` *(cfg: persist)* — persist a key to an NV persistent handle
- `read_persistent(persistent_handle, auth)` *(cfg: persist)*
- `evict_key(key, hierarchy)` *(cfg: persist)*

**`Device` — HMAC and sessions**
- `hmac(...)` *(cfg: hmac)* — one-shot raw-key HMAC
- `start_encrypted_session(salt)` — start a parameter-encryption session

**`Key` — identity, sign/verify**
- `handle()` — the TPM handle
- `name()` — the object Name (hash of its public area)
- `export_public(pem)` *(cfg: pubexport)* — export the public key as DER or PEM
- `sign_hash(digest)` — sign a digest
- `verify_hash(digest, sig)` — verify a signature

**`Key` — RSA (OAEP), AES, ECDH, HMAC**
- `rsa_encrypt(msg)` *(cfg: rsa)* — RSA-OAEP encrypt
- `rsa_decrypt(ciphertext)` *(cfg: rsa)* — RSA-OAEP decrypt (returns `Secret`)
- `rsa_encrypt_with_hash(msg, hash)` *(cfg: rsa)* — OAEP with an explicit label hash (Microsoft enrollment)
- `rsa_decrypt_with_hash(ciphertext, hash)` *(cfg: rsa)*
- `aes_encrypt(data, iv)` *(cfg: symmetric)* — AES-CFB encrypt
- `aes_decrypt(data, iv)` *(cfg: symmetric)* — AES-CFB decrypt (returns `Secret`)
- `ecdh_gen()` *(cfg: ecdh)* — ephemeral ECDH (returns `EcdhResult`)
- `ecdh_z(point)` *(cfg: ecdh)* — Z-agreement with a peer point (returns `Secret`)
- `hmac(data, hash)` *(cfg: keyedhash)* — HMAC with a TPM-resident keyed-hash key

**`KeyBlob` — serialize / load**
- `to_bytes()` — serialize the wrapped blob
- `from_bytes(dev, bytes)` — deserialize a blob
- `load(parent, auth)` — load into the TPM (yields a `Key`)

**`Template` — key templates**
- `srk(alg)` — storage-root/primary (restricted decrypt parent)
- `attestation(alg)` — restricted signing AIK
- `ek(alg)` — endorsement key
- `signing(alg)` — general signing key (ECDSA / RSASSA-SHA256)
- `rsa_decrypt()` — RSA decryption key (OAEP)
- `hmac(hash)` *(cfg: keyedhash)* — TPM-resident keyed-hash HMAC key
- `symmetric(bits)` *(cfg: symmetric)* — AES-CFB key (128/256)
- `ecdh()` *(cfg: ecdh)* — ECDH key-agreement key (P-256)

**`Secret`**: `as_bytes()`, `len()`, `is_empty()`, `Deref<Target=[u8]>`, zeroizes on drop.
**`TpmError`**: `code()` — the raw return code. **`HashAlg`**: `digest_size()`.

### Tests ran on fwtpm

| File | Tests |
| --- | --- |
| `caps.rs` | `self_test_and_capabilities` |
| `certify.rs` | `certify_with_ecc_aik`, `certify_with_rsa_aik` |
| `credential.rs` | `make_and_activate_credential` |
| `ecdh.rs` | `ecdh_gen_then_z_agree` |
| `ek.rs` | `create_ek_and_export_public` |
| `hmac.rs` | `hmac_deterministic_and_keyed`, `loaded_hmac_key_compute` |
| `import.rs` | `import_rsa_private_key`, `import_ecc_private_key` |
| `keys.rs` | `create_and_load_child`, `key_blob_roundtrip_then_load`, `key_blob_roundtrip_preserves_short_auth`, `auth_protected_parent_loads_child` |
| `nv.rs` | `nv_write_read_delete` |
| `pcr.rs` | `pcr_extend_changes_digest` |
| `persist.rs` | `persist_read_evict` |
| `quote.rs` | `quote_ecc_aik`, `quote_rsa_aik`, `quote_rejects_invalid_pcr_selection` |
| `rsa.rs` | `rsa_oaep_roundtrip`, `rsa_oaep_sha1_roundtrip` |
| `seal.rs` | `seal_unseal_roundtrip`, `unseal_wrong_auth_fails` |
| `seal_pcr.rs` | `pcr_bound_seal_roundtrip`, `pcr_change_breaks_unseal`, `seal_pcr_rejects_invalid_selection` |
| `session.rs` | `secret_ops_under_encrypted_session`, `second_session_rejected`, `attestation_refused_during_session` |
| `sign.rs` | `sign_then_verify`, `verify_rejects_tampered_signature` |
| `smoke.rs` | `random_and_primary_keys` |
| `symmetric.rs` | `aes_cfb_roundtrip` |

- Shared helper `tests/common/mod.rs::open()` connects each test to the fwTPM.
- Examples: `examples/create_primary.rs`, `examples/full_flow.rs`.
- CI (`.github/workflows/rust-test.yml`): `rust` job builds wolfSSL (matrix: `master` and latest `-stable`) + wolfTPM with the in-tree fwTPM and runs the suite; `rust-devtpm-compile` job is a compile-only Linux kernel-device (`--enable-devtpm`) build.

## Post merge steps

- [x] Confirm `Cargo.toml` metadata: name `wolftpm`, version, `license = GPL-3.0-or-later`, description, repository, keywords, categories, readme
- [x] Verify the `wolftpm` name is owned by the wolfSSL org on crates.io (reserve it if not)
- [x] Make the docs.rs build succeed without the C libs: gate `build.rs` on `DOCS_RS` (generate bindings from vendored headers / skip the link step), since docs.rs has no prebuilt `libwolftpm`/`libwolfssl`
- [ ] Set `package.include`/`exclude` so the published tarball ships only intended files (src, `build.rs`, `headers.h`, README, examples — not `target/`)
- [ ] `cargo publish --dry-run` from `wrapper/rust/wolftpm` and review the packaged file list
- [ ] `cargo publish` with a crates.io token that has wolfSSL-org publish rights
- [ ] Verify the docs.rs build renders and the crate page is correct
- [ ] Add crates.io + docs.rs badges and install/usage notes to `wrapper/rust/README.md` and the top-level wolfTPM README
- [ ] Add a `ChangeLog.md` entry and tag the release
- [ ] Announce (release notes / blog)


holtrop-wolfssl:
Currently after building the Rust wrapper the git status is no longer clean. I would add `wrapper/rust/wolftpm/Cargo.lock` to version control, and add `/wrapper/rust/wolftpm/target/` to `.gitignore`.

wrapper/rust/wolftpm/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+pub use certify::Attestation;
+#[cfg(all(wrapper, caps))]
+pub use caps::Caps;
+#[cfg(all(wrapper, ecdh))]
+pub use ecdh::EcdhResult;
+#[cfg(all(wrapper, ek_policy))]
+pub use credential::Credential;
+
+use core::fmt;
+use std::os::raw::c_int;
```
holtrop-wolfssl: For increased portability, you could pull in C FFI types from `core` instead of `std`.

While on that topic... would there be any use cases for this Rust wrapper in an embedded type environment? If so, supporting `no_std` would be nice. Many portable/embedded Rust users want to see `no_std` support. You can still use `alloc` even with `no_std` and increase portability over requiring `std`.

wrapper/rust/wolftpm/tests/session.rs
```diff
@@ -0,0 +1,56 @@
...
+        .unwrap();
+
+    // With the session at slot 1, seal (command param) and unseal (response
+    // param) travel encrypted; both must still round-trip correctly.
+    let session = dev.start_encrypted_session(&srk).unwrap();
+    let sealed = dev.seal(&srk, b"secret-under-session", None).unwrap();
+    let out = dev.unseal(sealed, &srk, None).unwrap();
+    assert_eq!(out.as_bytes(), &b"secret-under-session"[..]);
+
+    // GetRandom's response is also encrypted while the session is live.
```
holtrop-wolfssl: It does not look like the response is encrypted. Is that correct?

wrapper/rust/wolftpm/src/session.rs
```diff
@@ -0,0 +1,113 @@
...
+use crate::{sys, Result, TpmError, E_SESSION_IN_USE};
+use core::ffi::c_int;
+use core::marker::PhantomData;
+use core::sync::atomic::{AtomicBool, Ordering};
+
+/// Auth slot the encryption session occupies. It sits at slot 1, immediately
+/// after a command's own object authorization at slot 0. wolfTPM's command
+/// builder (`TPM2_GetCmdAuthCount`) then includes it as an encrypt/decrypt
+/// session for every parameter-encryption-capable command, with no gap that
+/// would drop it. The 2-auth attestation commands (certify, quote, activate)
```
holtrop-wolfssl: Claude is telling me that quote is not a 2-auth command like this comment says. But I don't know enough about wolfTPM to know if that is correct. Is it? This also relates to line 49 below and certify.rs:106.

wrapper/rust/wolftpm/src/nv.rs
```diff
@@ -0,0 +1,134 @@
...
+    let mut h: sys::WOLFTPM2_HANDLE = unsafe { core::mem::zeroed() };
+    h.hndl = sys::TPM_RH_T_TPM_RH_OWNER as sys::TPM_HANDLE;
+    h
+}
+
+impl Device {
+    /// Define an NV index of `size` bytes under the owner hierarchy,
+    /// auth-protected (auth/owner read+write).
+    pub fn nv_create(&self, index: u32, size: u32, auth: Option<&[u8]>) -> Result<NvSlot<'_>> {
+        let attrs = sys::TPMA_NV_AUTHREAD | sys::TPMA_NV_AUTHWRITE;
```
holtrop-wolfssl, line 41: Should OWNERREAD/OWNERWRITE be used here? Or "owner" removed from the line 36 comment?

wrapper/rust/wolftpm/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+//! detected from the linked C library at build time via `cfg` flags emitted by
+//! `build.rs`, so the surface reflects how libwolftpm was actually configured.
+//!
+//! # Safety
+//!
+//! `Device` boxes its `WOLFTPM2_DEV` in an `UnsafeCell` so it is heap-pinned
+//! and self-referential-safe; `Device::ptr()` is therefore a stable pointer
+//! for the whole `Device` lifetime, and every wolfTPM2 C call in this crate
+//! takes it. `Key`/`KeyBlob` borrow `&Device` and hold their own
+//! `UnsafeCell`-wrapped, pinned C struct plus the dev pointer, so their
```
holtrop-wolfssl: `Key` and `KeyBlob` do use `UnsafeCell` but do not appear to pin the C struct as this comment indicates. Is it safe if the Rust compiler moves the contents of the C structs in memory? If so the implementation might be ok but this comment should change. If not, perhaps they should truly be pinned in memory.

wrapper/rust/wolftpm/build.rs
```diff
@@ -0,0 +1,372 @@
...
+        let re =
+            Regex::new(&format!(r"(?m)^\s*#\s*define\s+{}\b", regex::escape(macro_name))).unwrap();
+        if re.is_match(&text) {
+            println!("cargo:rustc-cfg={}", cfg_name);
+        }
+    };
+    flag("WOLFTPM_SWTPM", "swtpm");
+    flag("WOLFTPM_LINUX_DEV", "devtpm");
+    flag("WOLFTPM_MMIO", "mmio");
+    flag("WOLFTPM_FWTPM", "fwtpm");
```
holtrop-wolfssl: Is this right? My options.h has `WOLFTPM_FWTPM_BUILD` instead.

wrapper/rust/wolftpm/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+/// and the attestation commands are refused while one holds the auth slot.
+pub(crate) const E_SESSION_IN_USE: c_int = -901;
+
+/// Zeroize an arbitrary FFI struct by raw bytes.
+///
+/// `Drop` impls that hold secret material scrub the backing bytes with volatile
+/// writes so the compiler cannot elide them (the Rust equivalent of wolfSSL's
+/// `ForceZero`).
+#[inline]
+#[allow(dead_code)] /* used by the sealed-secret modules landing next */
```
holtrop-wolfssl: Is this out of date? It looks like it was added part-way through development.

wrapper/rust/wolftpm/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+//! takes it. `Key`/`KeyBlob` borrow `&Device` and hold their own
+//! `UnsafeCell`-wrapped, pinned C struct plus the dev pointer, so their
+//! `kptr()`/`handle_ptr()` accessors are likewise stable for as long as the
+//! borrow lives. C structs passed to FFI are zero-initialized with
+//! `core::mem::zeroed()` first — valid for these C plain-old-data types — and
+//! then filled by the callee; buffer copies into fixed C arrays are bounds-
+//! checked beforehand so they cannot overflow. Union fields are read only
+//! after the code that set the matching selector (an `is_ecc`/scheme flag) ran
+//! immediately before. `zeroize_raw` and volatile-write scrubbing always
+//! operate on a live, correctly-sized, exclusively-owned local.
```
holtrop-wolfssl: This is incorrect. There are a few calls to `zeroize_raw` that pass struct field addresses, not local variable addresses.

wrapper/rust/wolftpm/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+pub(crate) const E_SESSION_IN_USE: c_int = -901;
+
+/// Zeroize an arbitrary FFI struct by raw bytes.
+///
+/// `Drop` impls that hold secret material scrub the backing bytes with volatile
+/// writes so the compiler cannot elide them (the Rust equivalent of wolfSSL's
+/// `ForceZero`).
+#[inline]
+#[allow(dead_code)] /* used by the sealed-secret modules landing next */
+pub(crate) unsafe fn zeroize_raw<T>(v: &mut T) {
```
holtrop-wolfssl: We should consider using the `zeroize` crate dependency to do this rather than implementing the volatile write and memory barrier ourselves.

If `zeroize_raw()` is still needed after pulling in the `zeroize` crate dependency (my guess is it will be but the number of calls could be reduced), it can use `zeroize` crate functionality like I did for the wolfcrypt wrapper here: https://github.com/wolfSSL/wolfssl/blob/master/wrapper/rust/wolfssl-wolfcrypt/src/lib.rs#L36

Using the `zeroize` crate could also reduce the number of `unsafe` blocks needed in the wrapper for zeroizing functionality.

wrapper/rust/wolftpm/build.rs
```diff
@@ -0,0 +1,372 @@
...
+        // both only produce warnings (libc memcpy/etc. trip the runtime-symbol
+        // lint, and doxygen [in,out]/[0] markers become broken rustdoc links).
+        .blocklist_function("memcpy")
+        .blocklist_function("memmove")
+        .blocklist_function("memset")
+        .blocklist_function("memcmp")
+        .blocklist_function("bcmp")
+        .blocklist_function("strlen")
+        .generate_comments(false)
+        .parse_callbacks(Box::new(bindgen::CargoCallbacks::new()));
```
holtrop-wolfssl: I'd add `.use_core()` call to this like wolfssl-wolfcrypt crate does.

wrapper/rust/wolftpm/build.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    // Preprocess for the actual target, not the build host: wolfTPM headers
+    // select struct fields and transports on __linux__/_WIN32/arch macros, so a
+    // host/target mismatch would generate the wrong FFI layout when cross-compiling.
+    if let Ok(target) = env::var("TARGET") {
+        builder = builder.clang_arg(format!("--target={}", target));
```
holtrop-wolfssl: Consider porting `rust_target_to_clang_target()` from wolfssl-wolfcrypt's build.rs. As is here this might fail with a Rust TARGET like "riscv32imac-unknown-none-elf".

holtrop-wolfssl: Also the check for target to be different from host (cross-compiling) and `bare_metal_sysroot()` might be nice to have here as well.

wrapper/rust/wolftpm/README.md
```diff
@@ -0,0 +1,257 @@
+# wolftpm
```
holtrop-wolfssl, line 1: Consider adding a `CHANGELOG.md` file next to these; it's common practice for a Rust crate.

wrapper/rust/README.md
```diff
@@ -0,0 +1,17 @@
...
+Official Rust bindings for wolfTPM. The crate lives in [`wolftpm/`](wolftpm/);
+see its [README](wolftpm/README.md) for the full API, build, and test details.
+
+```sh
+# 1. build the C library first (with a software TPM for testing)
+cd ../..                       # wolfTPM repo root
+./autogen.sh && ./configure --enable-swtpm --enable-fwtpm && make
+
+# 2. build the Rust crate against it
+cd wrapper/rust/wolftpm
```
holtrop-wolfssl: Optional, but you could add a `Makefile` in `wrapper/rust` so this becomes one line: `make -C wrapper/rust` the same as in the wolfssl repository.

It can also shorten the "default" test command below such as `make -C wrapper/rust test`.

holtrop-wolfssl (changes requested):

wrapper/rust/wolftpm/src/lib.rs
aidangarske: Ya there defiantly would be use case in embedded env I think. I made it so the crate now builds as `no_std`, uses `alloc` for owned types, imports C FFI types from `core`, and asks bindgen to emit `core`-based bindings.

wrapper/rust/wolftpm/tests/session.rs
aidangarske: yes I think so `TPM2_GetRandom` is sent without sessions so its response is not parameter encrypted. I removed the incorrect assertion; unseal continues to exercise encrypted response handling.

wrapper/rust/wolftpm/src/session.rs
aidangarske: thanks quote needs one object authorization so the encryption session can remain in slot 1. I removed the guard and added regression coverage; certify and credential activation remain guarded because they need two authorizations.

wrapper/rust/wolftpm/src/nv.rs
aidangarske: Fixed the comment to say auth read/write. The index operations use the per-index authorization; the owner hierarchy is used to define and delete the index.

wrapper/rust/wolftpm/src/lib.rs
aidangarske: Thanks, fixed the safety text. `Key` and `KeyBlob` may move, and their interior pointers are borrowed only for individual FFI calls wolfTPM does not retain them only the boxed `Device` requires a stable heap address.

wrapper/rust/wolftpm/build.rs
aidangarske: no, fixed fwTPM detection now checks `WOLFTPM_FWTPM_BUILD` from `options.h`.

wrapper/rust/wolftpm/src/lib.rs
aidangarske: Fixed. I removed the stale `dead_code` allowance now that `zeroize_raw` is in use.

wrapper/rust/wolftpm/src/lib.rs
aidangarske: Fixed the safety text to cover live initialized storage under an exclusive mutable borrow, including struct fields.

wrapper/rust/wolftpm/src/lib.rs
aidangarske: Makes sense owned buffers and raw FFI-structure byte slices now use the `zeroize` crate.

wrapper/rust/wolftpm/build.rs
aidangarske: Bindgen now calls `.use_core()`.

wrapper/rust/wolftpm/build.rs
aidangarske: The build script now maps Rust target triples to clang triples, sets `--target` only when cross-compiling, and discovers bare-metal sysroots.

wrapper/rust/wolftpm/README.md
aidangarske: okay will add one

wrapper/rust/README.md
aidangarske: Okay fixed, The generated `Cargo.lock` is now tracked, Cargo commands use `--locked`, and `/wrapper/rust/wolftpm/target/` is ignored.

wrapper/rust/wolftpm/README.md
aidangarske: added changelog for v1.0.0

wrapper/rust/wolftpm/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+categories = ["cryptography", "hardware-support", "api-bindings"]
+readme = "README.md"
+
+[features]
+default = []
+# Gate the socket integration tests that need a running swtpm/fwtpm server.
+swtpm-tests = []
+
+[dependencies]
+zeroize = { version = "=1.8.2", default-features = false, features = ["alloc"] }
```
holtrop-wolfssl: The dependency should not be locked to this specific version. Perhaps version = "1.8" instead, to allow crate users to be able to use a zeroize crate version that works with their other dependencies as well.

wrapper/rust/wolftpm/src/sys.rs
```diff
@@ -0,0 +1,17 @@
...
+
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(non_upper_case_globals)]
+#![allow(improper_ctypes)]
+#![allow(dead_code)]
+#![allow(clippy::missing_safety_doc)]
+#![allow(clippy::ptr_offset_with_cast)]
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::useless_transmute)]
```
holtrop-wolfssl, line 16: When I run `make -C wrapper/rust` now, I get several `warning: unnecessary transmute` warnings. This allow line might be incorrect. Should it be `allow(unnecessary_transmutes)` instead? (I'm not sure if this varies with the version of rustc I have installed)

holtrop-wolfssl (changes requested):

wrapper/rust/wolftpm/Cargo.toml
aidangarske: put to 1.8 now

wrapper/rust/wolftpm/src/sys.rs
aidangarske: I kept the separate Clippy allow and added rustc's `unnecessary_transmutes` allow, with `unknown_lints` so the crate stays warning free on its Rust 1.81 MSRV where that lint is not available.

holtrop-wolfssl (approved):

gasbytes (approved):

PR #607, opened 2026-09-12
Handle TPMs that lack SHA-1, disable EncryptDecrypt, or run a command longer than the TIS wait budget

dgarske:
Six independent robustness fixes plus a benchmark mode, found running the stock examples against current SPI TPMs on a Raspberry Pi 5.

- **Oversized signatures no longer hang the TPM.** `wolfTPM2_VerifyHashTicket()` and `wolfTPM2_VerifyDigestSignature()` now check the signature against `TPM_PT_INPUT_BUFFER` and return `BUFFER_E` before sending, rather than leaving the TPM to discover it; post-quantum signatures exceed that cap routinely and not every TPM rejects the oversized parameter cleanly, with at least one ceasing to respond until a hardware reset. The capability read is skipped at or below the 1024-byte `TPM_MIN_INPUT_BUFFER` floor, so RSA and ECC verifies cost no extra round trip.

- **TIS waits are bounded by real time instead of an iteration count.** `TPM2_TIS_WaitForStatus()` and `TPM2_TIS_GetBurstCount()` now use `TPM_TIMEOUT_MS` (default 300000) through a new `XTPM_GET_TIMEMS()`, so the budget no longer varies with host speed and scheduler; RSA-2048 key generation is an unbounded prime search whose duration varies by an order of magnitude for the same command, so a fixed iteration count timed out intermittently on runs that succeeded moments earlier. Ports without a monotonic clock keep the existing counter unchanged, and `TPM_TIMEOUT_TRIES` of zero keeps its "no cap" meaning on both paths.

- **Examples skip SHA-1 on TPMs that do not implement it.** `wrap_test` and `native_test` carried hard-coded SHA-1 test vectors and PCR bank selections that abort the run with `TPM_RC_HASH` on an increasing share of current parts; all three sites now query `TPM_CAP_ALGS` first and skip with the reason printed.

- **Examples tolerate EncryptDecrypt being disabled rather than absent.** They already skipped the test on `TPM_RC_COMMAND_CODE`, but a TPM that implements the command and ships it switched off answers `TPM_RC_DISABLED`; both codes now go through one helper, and the call sites that were missing the response-code masking gain it.

- **ASN error codes no longer collide with TPM response codes.** The ASN parser's internal error values shared the numeric space with `TPM_RC_*`, so a parse failure could be read as a TPM response. They are renamed and re-based, with deprecated aliases kept behind `WOLFTPM_NO_DEPRECATED_ASN_RC` so existing callers continue to build.

- **New `wolfTPM2_AllocatePCRBanks()` for changing which PCR banks a TPM allocates,** with `examples/pcr/allocate` to report and re-provision them. Parts that keep a single bank active need this to move between SHA-256 and SHA-384. Two details are worth flagging for anyone building the selection themselves:
  - TPM 2.0 Part 3 section 22.5 says a bank omitted from the `TPML_PCR_SELECTION` is deallocated, but Infineon parts answer `TPM_RC_PCR` unless every bank is named, with an all-zero `pcrSelect` bitmap on the ones being dropped. The wrapper mirrors the shape `TPM_CAP_PCRS` reports, which is the form those parts accept.
  - `TPM_CAP_ALGS` is not authoritative for what can be allocated; a part may advertise a hash there and have no PCR bank for it. The bank list from `TPM_CAP_PCRS` is used instead, so an unusable bank is rejected before the command is sent rather than silently ignored by the TPM.

  The change is staged and takes effect at the next `TPM2_Startup(CLEAR)` following a `_TPM_Init`, which no command can trigger, so the caller must power cycle. The firmware TPM gains the matching fixes: it no longer accepts a selection that would leave it with no banks at all, it honours the `pcrSelect` bitmap so a bank can actually be deallocated, and it no longer reports every bank as unallocated after a restart against an existing NV file.

- **`examples/bench/bench` gains a fixed-iteration mode.** `-iter=[n]` runs each algorithm exactly n times and reports the average with the per-iteration minimum and maximum, instead of running for a wall-clock budget. The existing duration mode is unchanged and remains the default. Operations whose cost varies run to run, such as RSA key generation and post-quantum signing, produce a single misleading figure under a short duration; the spread makes that visible.

## Hardware / test status

Validated on a Raspberry Pi 5 driving SPI TPMs over spidev with no kernel TPM driver in the path, against parts that exercise every path: no SHA-1, `TPM2_EncryptDecrypt` disabled, a 1024-byte `TPM_PT_INPUT_BUFFER`, and RSA-2048 key generation whose duration varies by an order of magnitude run to run. `native_test` now runs to completion where it previously aborted at the first SHA-1 PCR read, `examples/wrap/encrypt_decrypt` skips cleanly instead of failing, and `examples/bench/bench` completes its post-quantum rows instead of hanging partway. A full SHA-256 to SHA-384 bank switch and back was performed on hardware across power cycles.

Also run against the firmware TPM and the simulator with no change in results: `make check`, `run_examples.sh`, the SPDM suite in both TCG and PSK modes, and the tpm2-tools compatibility suite. The no-clock fallback is compile-verified with `-DWOLFTPM_NO_MONOTONIC_MS`, and both settings of `TPM_TIMEOUT_TRIES` are compile-verified.

## Scope

`TPM_TIMEOUT_MS` is a single global budget rather than a per-command duration derived from the TPM's own `TPM_PT_*` timeout and duration properties. That is the more correct answer and is left as follow-on work. The default is a tail estimate for the slowest command rather than a measured bound, since RSA key generation has no upper limit; a per-command duration would remove the guesswork.

`wolfTPM2_AllocatePCRBanks()` allocates each bank across all implemented PCRs rather than exposing per-PCR granularity, and deliberately refuses a zero-length selection, which is a legal command that would leave the TPM with no PCRs at all.


src/tpm2_wrap.c
```diff
@@ -5788,6 +5760,109 @@ int wolfTPM2_SignHash(WOLFTPM2_DEV* dev, WOLFTPM2_KEY* key,
...
+ * left to the TPM. Two limits apply: TPM_PT_INPUT_BUFFER bounds the parameter
+ * and TPM_PT_MAX_COMMAND_SIZE bounds the whole command, so the capability read
+ * is skipped only when the signature plus overhead fits inside
+ * TPM_MIN_INPUT_BUFFER, the floor every conformant TPM meets. That still
+ * covers every ECC signature and RSA up to 4096. Above it the limit must be
+ * established, so this fails closed. Returns TPM_RC_SUCCESS if it fits,
+ * BUFFER_E if it provably does not, and the query error otherwise so callers
+ * can tell the two apart. */
+
+/* Bytes that share the command with the signature: 10-byte header, up to two
```
aidangarske, line 5774: **Signature guard can still submit an oversized command**

The new guard calculates every command as `sigSz + 176`, but the guarded APIs have different serialized overhead. In particular, `wolfTPM2_VerifyDigestSignature()` can add a 255-byte context, digest, signature metadata, handles, and authorization data, exceeding the estimate. A signature can therefore pass both checks even though the actual command exceeds `TPM_PT_MAX_COMMAND_SIZE`. The function also returns success when the second capability query fails or returns a malformed property, despite the comment stating that both limits must be established. This re-exposes the oversized-command hang the change is intended to prevent.

**Fix:** Pass the exact operation-specific serialized parameter and command sizes into the guard, including context, digest, signature metadata, and authorization data. Fail closed when `TPM_PT_MAX_COMMAND_SIZE` cannot be read or validated. Add tests for maximum context and a failed or malformed second capability query.

src/tpm2_wrap.c
```diff
@@ -7413,6 +7505,159 @@ int wolfTPM2_ExtendPCR(WOLFTPM2_DEV* dev, int pcrIndex, int hashAlg,
...
+            }
+            in.pcrAllocation.pcrSelections[selIdx].hash =
+                banks->pcrSelections[j].hash;
+            in.pcrAllocation.pcrSelections[selIdx].sizeofSelect =
+                banks->pcrSelections[j].sizeofSelect;
+            in.pcrAllocation.count++;
+        }
+    }
+
+    /* Platform auth setup below overwrites session[0] */
```
aidangarske, line 7634: **PCR allocation discards the caller session's updated nonce**

When a caller supplies an HMAC or policy session, `TPM2_PCR_Allocate()` updates `dev->session[0].nonceTPM` from the response. The wrapper then restores the saved slot without copying that nonce back to the caller's `WOLFTPM2_SESSION`. The caller consequently retains the pre-command nonce, so its next use of the continuing session will generate an invalid authorization HMAC. Existing `wolfTPM2_UnsetAuthSession()` explicitly performs this nonce copy for exactly this reason, but the new wrapper bypasses it. The tests exercise only the `session == NULL` path.

**Fix:** Copy the response nonce back into the supplied session before restoring slot 0, following the `wolfTPM2_UnsetAuthSession()` contract. Add a test that authorizes allocation with a non-NULL continuing session and successfully reuses that session for another command.

src/tpm2_tis.c
```diff
@@ -417,22 +417,119 @@ int TPM2_TIS_Status(TPM2_CTX* ctx, byte* status)
...
+    to->tries = TPM_TIMEOUT_TRIES;
+#ifdef WOLFTPM_HAVE_MONOTONIC_MS
+    /* A zero tick is a real timestamp (boot, or the word32 wrapping through
+     * 0), so it must not be treated as an unreadable clock. */
+    to->start = XTPM_GET_TIMEMS();
+#endif
+}
+
+/* Returns 1 once the budget is spent, 0 while there is still time. */
+static int TPM2_TIS_TimeoutExpired(TPM2_TIS_TIMEOUT* to)
```
aidangarske, line 451: **A clock that stalls after advancing disables every timeout**

The iteration backstop expires only when `elapsed == 0`. If the monotonic source advances once and then stalls, `elapsed` remains nonzero but never reaches `TPM_TIMEOUT_MS`; after `tries` reaches zero the function continues returning false forever. Both TIS wait loops can therefore hang indefinitely, whereas the previous implementation retained a finite iteration bound. This affects port-supplied clocks and RTOS tick sources if their tick stops or becomes unavailable after startup.

**Fix:** Track the last observed tick and count consecutive polls without clock progress. Expire after the configured stagnant-poll budget while retaining the real-time deadline when the clock advances. Add deterministic tests for normal expiry, wraparound, a clock stuck from startup, and a clock that advances before stalling.

src/tpm2.c
```diff
@@ -7919,6 +7919,88 @@ void TPM2_PrintPublicArea(const TPM2B_PUBLIC* pub)
...
+
+    if (isAllocated == NULL) {
+        return BAD_FUNC_ARG;
+    }
+    *isAllocated = 0;
+    if (pcrIndex < 0) {
+        return BAD_FUNC_ARG;
+    }
+
+    XMEMSET(&in, 0, sizeof(in));
```
aidangarske, line 7976: **Truncated PCR capability results are treated as complete**

The helper requests only `HASH_COUNT` PCR selections and ignores `GetCapability_Out.moreData`. `HASH_COUNT` is derived from enabled wolfCrypt hashes and is only two in no-wolfCrypt builds, so a TPM exposing more banks can return a partial page. A requested bank after that page is then reported as unallocated with `TPM_RC_SUCCESS`. `wolfTPM2_AllocatePCRBanks_ex()` repeats the same assumption at `src/tpm2_wrap.c:7544-7558`, causing an implemented bank to be rejected as `TPM_RC_HASH` and preventing the wrapper from mirroring the complete selection.

**Fix:** Represent and consume the complete TPM PCR-bank list independently of wolfCrypt's enabled hash count. At minimum, return an error rather than a false result whenever `moreData` indicates that the list was truncated. Cover a reduced-`HASH_COUNT` build against a TPM reporting more banks.

examples/pcr/allocate.c
```diff
@@ -0,0 +1,284 @@
...
+#include <hal/tpm_io.h>
+#include <examples/tpm_test.h>
+
+#include <stdio.h>
+
+/******************************************************************************/
+/* --- BEGIN TPM2.0 PCR Allocate example tool  -- */
+/******************************************************************************/
+
+/* Banks the report inspects; unimplemented ones print "no" */
```
aidangarske: **The restore option can silently drop existing PCR banks**

The original allocation recorded for `-restore` is limited to SHA-1, SHA-256, SHA-384, and SHA-512, and a bank is considered allocated only if PCR 0 is selected. An allocated SM3 or SHA-3 bank is omitted, as is any legal partial bank without PCR 0. Because `wolfTPM2_AllocatePCRBanks()` replaces the complete allocation, the restore command subsequently deallocates every omitted bank and expands retained partial banks to all PCRs. This contradicts the documented promise that `-restore` leaves the TPM as it was found.

**Fix:** Capture the complete `TPML_PCR_SELECTION`, including every reported algorithm and its original bitmap, and restore that exact selection through `TPM2_PCR_Allocate`. If exact restoration is outside the high-level API's scope, remove the restore guarantee and do not use this option as cleanup in scripts.

examples/run_examples.sh
```diff
@@ -890,6 +890,26 @@ if [ $WOLFCRYPT_ENABLE -eq 1 ] && [ $NO_FILESYSTEM -eq 0 ]; then
     # Keeping keyblob.bin for tests later
 fi
 
+# PCR bank allocation. The query-only form must work everywhere. The
```
aidangarske: **The default example suite performs destructive PCR reprovisioning**

The default example run now executes `allocate -sha256 -restore` against whichever TPM is attached. PCR allocation is persistent platform configuration that invalidates PCR policies and sealed objects after reset. The two commands are not transactional: interruption, termination, or restore failure after the first command leaves a new allocation staged, and the fwTPM persists the first change immediately. The script already has `ENABLE_DESTRUCTIVE_TESTS` for state-changing operations, but this new block is not guarded.

**Fix:** Keep the query-only invocation in the default smoke suite, but place the allocation-changing invocation behind `ENABLE_DESTRUCTIVE_TESTS` or restrict it to a disposable fwTPM NV image.

examples/bench/bench.c
```diff
@@ -43,16 +43,44 @@
...
+    else if (each < gIterMin) {
+        gIterMin = each;
+    }
+    if (gBenchIter > 0) {
+        return (*count < gBenchIter);
+    }
+    return ((now - start) < maxDurSec);
 }
 
 /* countSz is number of bytes that 1 count represents. Normally bench_size,
```
aidangarske, line 86: **Fixed-iteration spread is omitted from symmetric rows**

The new option promises an average plus per-iteration minimum and maximum for each algorithm, and `bench_stats_check()` records those values for every loop. Only `bench_stats_asym_finish()` prints them. RNG, AES, and hash rows use `bench_stats_sym_finish()`, which continues to print aggregate throughput without the iteration count, average, minimum, or maximum.

**Fix:** Print the fixed-iteration count, per-operation average, minimum, and maximum in `bench_stats_sym_finish()`, or narrow the option documentation to the algorithm classes that actually report spread.

examples/bench/bench.c
```diff
@@ -417,6 +455,15 @@ int TPM2_Wrapper_BenchArgs(void* userCtx, int argc, char *argv[])
             const char* maxStr = argv[argc-1] + XSTRLEN("-maxdur=");
             maxKeyGenDurSec = maxDuration = XATOI(maxStr) / 1000.0;
         }
+        else if (XSTRNCMP(argv[argc-1], "-iter=", XSTRLEN("-iter=")) == 0) {
+            const char* iterStr = argv[argc-1] + XSTRLEN("-iter=");
+            gBenchIter = XATOI(iterStr);
```
aidangarske, line 477: **Iteration mode persists across benchmark invocations**

`gBenchIter` is static and is changed only when an `-iter=` argument is parsed. After one call with fixed-iteration mode, a later call to the public `TPM2_Wrapper_Bench()` or `TPM2_Wrapper_BenchArgs()` without that option still runs with the previous iteration count instead of the documented duration-mode default.

**Fix:** Reset `gBenchIter` at the start of each benchmark invocation, or make the parsed mode local and pass it to the statistics helpers. Add a test that calls the entry point first with `-iter` and then without it.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

**Findings:** 8 total — 8 posted, 0 skipped
**8 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **Signature guard can still submit an oversized command** — `src/tpm2_wrap.c:5774-5863`
- [High] **PCR allocation discards the caller session's updated nonce** — `src/tpm2_wrap.c:7612-7637`
- [High] **A clock that stalls after advancing disables every timeout** — `src/tpm2_tis.c:448-481`
- [High] **Truncated PCR capability results are treated as complete** — `src/tpm2.c:7976-8001`
- [High] **The restore option can silently drop existing PCR banks** — `examples/pcr/allocate.c:42-109`
- [High] **The default example suite performs destructive PCR reprovisioning** — `examples/run_examples.sh:893-911`
- [Medium] **Fixed-iteration spread is omitted from symmetric rows** — `examples/bench/bench.c:88-138`
- [Medium] **Iteration mode persists across benchmark invocations** — `examples/bench/bench.c:419-460`

aidangarske (approved):

